PR #4, opened 2019-12-14
Use `Raw string literals` instead of `Interpreted string literals` in…

hueypark:
… test

It is more efficient to express `double quote`.


hackerwins:
@hueypark Thanks for the contribution. :)

PR #25, opened 2020-03-19
Clean up codes for readability

loganstone:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature

Just cleanup codes

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

### DIFF
--- a/pkg/document/time/actor_id.go
+++ b/pkg/document/time/actor_id.go
@@ -19,16 +19,30 @@ package time
 import (
 	"bytes"
 	"encoding/hex"
+	"math"
 )
+
+const actorIDSize = 12
 
 var (
 	InitialActorID = &ActorID{}
 	MaxActorID     = &ActorID{
-		255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
+		math.MaxUint8,
+		math.MaxUint8,
+		math.MaxUint8,
+		math.MaxUint8,
+		math.MaxUint8,
+		math.MaxUint8,
+		math.MaxUint8,
+		math.MaxUint8,
+		math.MaxUint8,
+		math.MaxUint8,
+		math.MaxUint8,
+		math.MaxUint8,
 	}
 )
 
-type ActorID [12]byte
+type ActorID [actorIDSize]byte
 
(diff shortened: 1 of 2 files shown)


pkg/document/time/ticket.go
```diff
@@ -18,6 +18,13 @@ package time
 
 import (
 	"fmt"
+	"math"
+)
+
+const (
+	less = iota - 1
+	equal
+	greater
```
hackerwins: Is this a popular pattern in Go? I haven't seen much of this pattern in the Golang base package.

loganstone: Yes. example, https://golang.org/src/sync/mutex.go?s=765:813#L36
If you want to check more, recommended to try `rg iota ** / *. go` or `grep iota ** / *. go` in `go/src` directory.

pkg/document/time/actor_id.go
```diff
@@ -19,16 +19,30 @@ package time
 import (
 	"bytes"
 	"encoding/hex"
+	"math"
 )
 
+const maxSize = 12
```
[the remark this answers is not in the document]
loganstone: This is non export variable. so I think universal name more appropriate.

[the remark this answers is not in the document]
hackerwins: I think the name maxSize is appropriate when the size is not fixed.

pkg/document/time/ticket.go
hackerwins: I've seen many Go base packages use 1, 0, -1 instead of less, equal, greater. Is this pattern more readable?

pkg/document/time/actor_id.go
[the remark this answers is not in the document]
loganstone: You right. I will change it.

pkg/document/time/ticket.go
loganstone: My answer is "yes". However, I agree that it is good to follow base package. I'll fix it

PR #71, opened 2020-08-02
Add increase operation

### LINKED ISSUE #12: Add increase operation for number types to use as a counter
Increase operation is an operation that can be used when implementing functions such as the Facebook Like button. We can also use Increase operation as Decrease operation if we handle negative values. 

The following are examples of APIs provided to users.

```go
err = doc.Update(func(root *proxy.ObjectProxy) error {
	root.SetInteger("age", 36)
	root.GetInteger("age").Increase(1)
	return nil
}, "The day after your birthday")
```

The operation is relatively easy to implement because there is no conflict between operations created between different clients. And I think implementing Increase operation will help you understand Yorkie's overall data flow.

It would be helpful to refer to other operations to implement the operation.
https://github.com/yorkie-team/yorkie/tree/master/pkg/document/operation

dc7303:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
/kind feature

Needed to manipulate values ​​of number types.  
I add `number_proxy.go` and `increase.go` for increase operation on `Integer` and `Long` types.

Fixes #12 

Please provide feedback this PR.  
I worried that adding `Increase` method to `Primitive`.  
That's because there are many other types in 'Primitive' type besides number type.

```release-note

```

**Additional documentation**:

```docs

```


hackerwins:
@dc7303 Thank you for your contribution. I will review this soon.

pkg/document/operation/increase.go
```diff
@@ -0,0 +1,74 @@
...
+package operation
+
+import (
+	"github.com/yorkie-team/yorkie/pkg/document/json"
+	"github.com/yorkie-team/yorkie/pkg/document/time"
+)
+
+type Increase struct {
+	parentCreatedAt *time.Ticket
+	prevCreatedAt   *time.Ticket
```
hackerwins, line 26: I don't think we need this attribute here. `prevCreatedAt` is used to check the previous insertion position when inserting an element into the RGA.

pkg/document/document_test.go
```diff
@@ -381,6 +381,26 @@ func TestDocument(t *testing.T) {
...
+			return nil
+		})
+		assert.NoError(t, err)
+		assert.Equal(t, `{"age":17}`, doc.Marshal())
+
+		err = doc.Update(func(root *proxy.ObjectProxy) error {
+			root.SetLong("price", 9000000000000000000)
+			root.GetLong("price").Increase(5).Increase(-2)
+			return nil
+		})
```
hackerwins, line 399: It would be nice to add tests for edge cases here. e.g.) a test that puts Long in the Integer, etc...

pkg/document/json/primitive.go
```diff
@@ -220,3 +220,20 @@ func (p *Primitive) Remove(removedAt *time.Ticket) bool {
...
+	vType := v.valueType
+	if pType != Integer && pType != Long && vType == Integer && vType != Long {
+		panic("unsupported type")
+	}
+	switch pType {
+	case Integer:
+		p.value = p.value.(int) + v.value.(int)
+	case Long:
+		p.value = p.value.(int64) + v.value.(int64)
+	}
```
hackerwins, line 236: I would also like to add some handling for float64.

dc7303:
@hackerwins Thanks for your advice.  
I pushed through the correction work. Please tell us if you have any additional requests :)

pkg/document/proxy/number_proxy.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// Increase adds an increase operation.
+// Only numeric types are allowed as operand values, excluding uint64 and uintptr.
+func (p *NumberProxy) Increase(v interface{}) *NumberProxy {
+	if !isAllowedOperand(v) {
+		panic("unsupported type")
+	}
+	var primitive *json.Primitive
+	tickect := p.context.IssueTimeTicket()
```
hackerwins, line 52: It looks like a typo.

dc7303:
@hackerwins Thanks to your advice, I can update. Please tell us if you have any additional requests.

And I think I should add an operation to `yorkie-js-sdk`. Can I keep working it?  
If I can do it, I will proceed from now on.

pkg/document/json/primitive.go
```diff
@@ -220,3 +220,22 @@ func (p *Primitive) Remove(removedAt *time.Ticket) bool {
 func (p *Primitive) ValueType() ValueType {
 	return p.valueType
 }
+
+// Increase increase integer, long or double.
+func (p *Primitive) Increase(v *Primitive) *Primitive {
+	pType := p.valueType
+	vType := v.valueType
+	if pType != Integer && pType != Long && pType != Double && vType != Integer && vType != Long && vType != Double {
```
hackerwins, line 228: If pType or vType is not a numeric type, we will have to judge it as an error condition. It would be less confusing to extract a function that returns whether it is numeric or not(`IsNumericType`). And It would be nice to add test code for edge cases as well.

hackerwins:
```go
if (pType != Integer && pType != Long && pType != Double) || (vType != Integer && vType != Long && vType != Double)
```

PR #84, opened 2020-08-29
Remove panic in types.EncodeOperation

### LINKED ISSUE #50: Remove panic from server code
Currently, panic is called when an exception such as an error occurs in the server code. The server goes down, but it is easier to grasp the error situation.

However, in the production environment, it should be changed to return an error instead of panic so that the server does not stop.

gmlwo530:
Remove panic in types.EncodeOperation. And change return value of types.EncodeOperation to ([][]bytes, error).
Add errors.New() with message used in panic.

**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature

Fixes #50 

```release-note

```

**Additional documentation**:

```docs

```

### DIFF
--- a/yorkie/backend/mongo/client.go
+++ b/yorkie/backend/mongo/client.go
@@ -311,6 +311,11 @@ func (c *Client) CreateChangeInfos(
 		var modelChanges []mongo.WriteModel
 
 		for _, c := range changes {
+			encodedOperations, err := types.EncodeOperations(c.Operations())
+			if err != nil {
+				return err
+			}
+
 			modelChanges = append(modelChanges, mongo.NewUpdateOneModel().SetFilter(bson.M{
 				"doc_id":     docID,
 				"server_seq": c.ServerSeq(),
@@ -319,7 +324,7 @@ func (c *Client) CreateChangeInfos(
 				"client_seq": c.ID().ClientSeq(),
 				"lamport":    c.ID().Lamport(),
 				"message":    c.Message(),
-				"operations": types.EncodeOperation(c.Operations()),
+				"operations": encodedOperations,
 			}}).SetUpsert(true))
 		}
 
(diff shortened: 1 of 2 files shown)


yorkie/backend/mongo/client.go
```diff
@@ -311,6 +311,12 @@ func (c *Client) CreateChangeInfos(
 		var modelChanges []mongo.WriteModel
 
 		for _, c := range changes {
+			encodedOperations, err := types.EncodeOperation(c.Operations())
+			if err != nil {
+				log.Logger.Error(err)
```
hackerwins, line 316: Since log.Logger.Error prints stack trace, it would be better to call only when an error occurs in the method, not our code. Can we move it to below position?
https://github.com/yorkie-team/yorkie/pull/84/files#diff-5a46bd121e6927f28ccfccbafe3ea9ffL46

hackerwins (commented):
Thanks for your contribution.

yorkie/types/change_info.go
```diff
@@ -37,18 +39,18 @@ type ChangeInfo struct {
 	Operations [][]byte           `bson:"operations"`
 }
 
-func EncodeOperation(operations []operation.Operation) [][]byte {
+func EncodeOperation(operations []operation.Operation) ([][]byte, error) {
```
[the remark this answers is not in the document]
gmlwo530: OK, I will change it

yorkie/backend/mongo/client.go
gmlwo530: I changed position of `log.Logger.Error(err)`. Is it right?

hackerwins:
@gmlwo530 Thanks for your contribution.

PR #104, opened 2020-11-17
Add garbage collection for text type

### LINKED ISSUE #58: Garbage collection for Text and RichText
In issue #3, we have implemented Garbage Collection for child elements of Container such as Array and Object.

https://github.com/yorkie-team/yorkie/blob/master/pkg/document/json/root.go#L94

Text and Rich Text use RgaTreeSplit and may have removed TextNodes inside Although they are not Container. We need the Garbage Collection of this removed TextNodes.

https://github.com/yorkie-team/yorkie/blob/master/pkg/document/json/rga_tree_split.go#L225

dc7303:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
/kind feature

There may be RGASplitNode removed when the text type is modified. And this can accumulate and result in wasted memory.
So I implemented a garbage collector for text types.

I will share this PR and continue to analyze and improve it.
Please give us honest feedback.

Fixes #58 

```release-note

```

**Additional documentation**:

```docs

```


hackerwins:
@dc7303 Thanks for contributing. I'll do a code review soon.

hackerwins:
@mojosoeun recently implemented the GC feature in the JS SDK, so if possible, it would be nice to review this PR together.

pkg/document/json/root.go
```diff
@@ -101,6 +108,14 @@ func (r *Root) GarbageCollect(ticket *time.Ticket) int {
...
 	}
 
+	for _, text := range r.editedTextElementMapByCreatedAt {
+		removedNodeCnt := text.cleanupRemovedNodes(ticket)
+		if removedNodeCnt > 0 {
+			delete(r.editedTextElementMapByCreatedAt, text.CreatedAt().Key())
+		}
+		count += removedNodeCnt
+	}
+
```
hackerwins, line 118: I think It would be better to process garbage collection for `Text` that has removed node rather than edited for two reasons below.

- We can reduce the size of the cache(`editedTextElementMapByCreatedAt `).
- It is more readable that GC is performed on the removed node than on the modified node.

api/converter/converter_test.go
```diff
@@ -21,6 +21,8 @@ import (
 	"testing"
 	"time"
 
+	time2 "github.com/yorkie-team/yorkie/pkg/document/time"
+
 	"github.com/stretchr/testify/assert"
 
 	"github.com/yorkie-team/yorkie/api/converter"
```
hackerwins, line 28: I hope we group the package into 3 parts.

- Standard library
- 3rd party library
- internal package

I referenced the [Uber Go Style Guide](https://github.com/uber-go/guide/blob/master/style.md#import-group-ordering).

pkg/document/json/rga_tree_split.go
```diff
@@ -222,10 +222,22 @@ func (t *RGATreeSplitNode) Value() RGATreeSplitValue {
...
+	key := fmt.Sprintf("%s:%d", createdAtKey, offset)
+	return removedNodesMapKey(key)
+}
+
 type RGATreeSplit struct {
 	initialHead *RGATreeSplitNode
 	treeByIndex *splay.Tree
 	treeByID    *llrb.Tree
+
+	removedNodesMap map[removedNodesMapKey]*RGATreeSplitNode
```
hackerwins, line 240: Since `map` already has a plural meaning, it would be nice to use the `{singular}Map` pattern to name this map(`removedNodesMap` -> `removedNodeMap`).

hackerwins (changes requested):
Thank you for sending PR. 👍 

I left a short review before looking at [the purge algorithm](https://github.com/yorkie-team/yorkie/pull/104/files#diff-db323cd470ef566aba2330997b694d23b4d0914fd4cb7caf38260b0841fc7c5dR540-R557).

pkg/document/json/root.go
dc7303: I changed editedTextElementMapByCreatedAt in d16bfc2 to removedNodeTextElementMapByCreatedAt.
I don't like the name because very long, but I changed. I don't come up with a better idea. 
If you have any better ideas, please share.

On the one hand, I think it would have been better if the return value of `Key()` of `CreatedAt` was defined as a type and used. In this case, I would not have had to add a suffix like `ByCreatedAt`.
Let me know what you think.

pkg/document/json/rga_tree_split.go
```diff
@@ -492,3 +524,45 @@ func (s *RGATreeSplit) AnnotatedString() string {
...
+	return count
+}
+
+// purge removes the node passed as a parameter from RGATreeSplit.
+func (s *RGATreeSplit) purge(node *RGATreeSplitNode) {
+	if node.prev != nil {
+		node.prev.next = node.next
+		if node.next != nil {
+			node.next.prev = node.prev
+		}
```
mojosoeun, line 556: @dc7303 I have a question. is the purpose of the purge function to remove a specific node from theLinkedList?

dc7303: Thanks for the review! 👍 

When the text is modified, it does not remove the node for the deleted text, just tombstone marking. (Tombstone is `removedAt`)
This means that there are persistent nodes for deleted text, which can be a waste of memory.
We need to remove the tombstone marked nodes after synchronizing with other clients to fix this problem.

When executing `garbage collection for large size of text garbage test`, try to modify `RGATreeSplit.cleanupRemovedNodes` as shown below. Then you can see that the size of the object before and after the modification is different.

```go
// rga_tree_split.go
func (s *RGATreeSplit) cleanupRemovedNodes(ticket *time.Ticket) int {
	count := 0
	for _, node := range s.removedNodeMap {
		if node.removedAt != nil && ticket.Compare(node.removedAt) >= 0 {
			s.treeByIndex.Delete(node.indexNode)
			// s.purge(node)
			s.treeByID.Remove(node.id)
			delete(s.removedNodeMap, removedNodeMapKey(node.createdAt().Key()))
			count++
		}
	}

	return count
}
```

In addition, you can check the state of the tree through the `Text.AnnotatedString()` method within `doc.Update()`.

```go
	t.Run("garbage collection for text test", func(t *testing.T) {
		
		...

		// Example
		err = doc.Update(func(root *proxy.ObjectProxy) error {
			text := root.GetText("k1")
			fmt.Println(text.AnnotatedString())
			return nil
		}, "edit text k1")
	}

// Before s.purge(node) comment processing
//     [0:0:00:0 ][2:1:00:0 Hi][1:2:00:5  ][2:2:00:0 j][2:3:00:0 ane]
// After s.purge(node) comment processing
//     [2:1:00:0 Hi]{1:2:00:0 Hello}[1:2:00:5  ][2:2:00:0 j][2:3:00:0 ane]{1:3:00:0 m}{1:3:00:1 ario}{1:2:00:6 world}
```
---

I found out that the `garbage collection for large size of text garbage test` test was wrong thanks to your comments. Thanks to you, I modified this test. Thank you so much! 😄

pkg/document/json/rga_tree_split.go
```diff
@@ -426,16 +456,18 @@ func (s *RGATreeSplit) deleteNodes(
...
 			s.treeByIndex.Splay(node.indexNode)
-
 			latestCreatedAt := createdAtMapByActor[actorIDHex]
 			createdAt := node.id.createdAt
 			if latestCreatedAt == nil || createdAt.After(latestCreatedAt) {
 				createdAtMapByActor[actorIDHex] = createdAt
 			}
+
+			key := createRemovedNodeMapKey(createdAt.Key(), node.id.offset)
+			removedNodeMap[key] = node
```
hackerwins, line 466: It was confirmed that the key can be created with the `created_at` and `offset` properties of the node's id. How about generating the key with node id?

```suggestion
			removedNodeMap[node.id.key()] = node
```

api/converter/converter_test.go
```diff
@@ -26,6 +26,7 @@ import (
...
 	"github.com/yorkie-team/yorkie/pkg/document"
 	"github.com/yorkie-team/yorkie/pkg/document/proxy"
+	time2 "github.com/yorkie-team/yorkie/pkg/document/time"
```
hackerwins, line 29: How about using imports as `time` and `gotime`?

```suggestion
	gotime "time"
	...
	"github.com/yorkie-team/yorkie/pkg/document/time"
```

pkg/document/document_test.go
```diff
@@ -19,8 +19,11 @@ package document_test
...
 	"errors"
 	"fmt"
+	"runtime"
 	"testing"
 
+	"github.com/yorkie-team/yorkie/pkg/document/json"
+
 	"github.com/stretchr/testify/assert"
 
 	"github.com/yorkie-team/yorkie/api/converter"
```
hackerwins, line 29: Please group packages in the following order.

- Standard library
- 3rd party library
- Internal packages

https://github.com/uber-go/guide/blob/master/style.md#import-group-ordering

hackerwins (changes requested):
Thank you for your work. 👍 

When I check the algorithm, I think it is okay to delete nodes marked with a tombstone. As the editing progresses, the root node begins to split. When we need to find for `RGATreeSplitNode` for the given location(`offset`) outside the RGATreeSplit, it is not possible to know exactly how nodes are splits, so we use the approximate location by `findFloorNode`. Even if the node the given location is deleted, it will ok because we use the approximate location.

It would be nice to check detailed tests while editing the actual examples in the JS SDK.

Please merge master and fix the errors caused by the recently introduced linters.

pkg/document/document_test.go
hackerwins: I sent a PR related to Import Group Ordering.
https://github.com/yorkie-team/yorkie/pull/117/files

dc7303: Thanks for the feedback.
I seem to make this mistake often. I think I can solve this problem thanks to the PR you sent us. :)

api/converter/converter_test.go
dc7303: Thanks for the feedback!
This proposal has been modified from b52d78d. Please check.

pkg/document/document_test.go
dc7303: This proposal has been modified from b52d78d. Please check.

pkg/document/json/rga_tree_split.go
dc7303: I think this is a really cool idea! I did this in the 1eefbb3 commit.
Removed the removedNodeMapKey type that I added, and it looks cooler. Thank you.

dc7303:
> Thank you for your work. 👍
> 
> When I check the algorithm, I think it is okay to delete nodes marked with a tombstone. As the editing progresses, the root node begins to split. When we need to find for `RGATreeSplitNode` for the given location(`offset`) outside the RGATreeSplit, it is not possible to know exactly how nodes are splits, so we use the approximate location by `findFloorNode`. Even if the node the given location is deleted, it will ok because we use the approximate location.
> 
> It would be nice to check detailed tests while editing the actual examples in the JS SDK.
> 
> Please merge master and fix the errors caused by the recently introduced linters.


Thanks for the nice explanation. I think it would be nice if we could present this well in the RGATreeSplit or GC design documents we will work on later. 😄

pkg/document/json/root_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"testing"
+
+	"github.com/yorkie-team/yorkie/pkg/document/time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/yorkie-team/yorkie/pkg/document/json"
+	"github.com/yorkie-team/yorkie/testhelper"
```
hackerwins, line 27: Please sort the import groups.

```suggestion
	"testing"

	"github.com/stretchr/testify/assert"

	"github.com/yorkie-team/yorkie/pkg/document/json"
	"github.com/yorkie-team/yorkie/pkg/document/time"
	"github.com/yorkie-team/yorkie/testhelper"
```

hackerwins (approved):
LGTM. 👍 

Now, I think we need to implement it in the JS SDK and do various tests using examples.

I left a brief comment.

PR #106, opened 2020-11-22
Move `golint` and `govet` from `make lint` to `golangci-lint`

hackerwins:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature

Move linters to `golangci-lint` so that it is not executed repeatedly.
And add `make lint` to CI to check PR.

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,15 @@
+linters:
+  enable:
+    - bodyclose
+    - goconst
+    - gocyclo
+    - gofmt
+    - goimports
+# TODO: When adding a rule to force comments on exported methods, be sure to uncomment below.
+#    - golint
+    - goprintffuncname
+    - gosec
+    - interfacer
+    - lll
+    - misspell
+    - nakedret
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,7 @@ jobs:
       - checkout
 
       # specify any bash command here prefixed with `run: `
+      - run: make lint
       - run: go get -v -t -d ./...
       - run: go test -race -coverprofile=coverage.txt -covermode=atomic -v ./...
       - codecov/upload:
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ GO_TOOLS = \
   github.com/gogo/protobuf/protoc-gen-gogo \
   github.com/gogo/protobuf/protoc-gen-gofast \
   golang.org/x/tools/cmd/goimports \
-  github.com/golangci/golangci-lint \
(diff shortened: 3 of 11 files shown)


.golangci.yml
```diff
@@ -0,0 +1,26 @@
...
+#    - golint
+    - goprintffuncname
+    - gosec
+    - gosimple
+    - govet
+    - ineffassign
+    - interfacer
+    - lll
+    - misspell
+    - nakedret
```
dc7303, line 21:
```suggestion
  enable:
    - bodyclose
    - goconst
    - gocyclo
    - gofmt
    - goimports
# TODO: When adding a rule to force comments on exported methods, be sure to uncomment below.
#    - golint
    - goprintffuncname
    - gosec
    - interfacer
    - lll
    - misspell
    - nakedret
```

How about clearing the `Enabled By Default Linters` item?
https://golangci-lint.run/usage/linters/#enabled-by-default-linters

dc7303 (changes requested):
That's cool! :)
I gave you some opinions. Let us know what you think.

.golangci.yml
hackerwins: I chose this style([disable-all: true](https://github.com/yorkie-team/yorkie/pull/106/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R2)) because it doesn't seem explicit that there is a default linter. Is it a little too much?
https://github.com/yorkie-team/yorkie/pull/106/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R2

dc7303: Ah! Did not recognize `disable-all: true`. If so, it looks like no problem! :)

dc7303: @hackerwins 
Can you explain why you selected only some of the linters that are enabled by default?

hackerwins: I added the strict linters suggested in [a post](https://medium.com/wesionary-team/introduction-to-linters-in-go-and-know-about-golangci-lint-6486fb2864b3). I didn't notice that the default linters is missing. 

I think it would be better to allow the default linters as new suggestions may be added to the default.

dc7303: That's a good opinion. I think we can check the code more strictly using the default linters. After using it, if there are any inconveniences to us in the future, I think we can select certain linters at that time.

dc7303 (approved):
LGTM 👍 I looked at the code that changed, and I feel like the project is getting more and more solid :)

Issue #48, opened 2020-04-20
Introduce Peer Awareness


hackerwins:
When multiple users edit a document, the user wants to know who is connected and what is being edited by the peer. For example, in Google Docs, the ID and profile of the user editing together are displayed in the upper-right.

- Yjs provides Awareness API: https://github.com/yjs/y-protocols#awareness-protocol
- MPL provides PeerStats: https://github.com/automerge/mpl/blob/e6b4c058e9f9c824ba86eb2603dc4ccf35778d84/src/mpl/network/peer-stats.js

PR #108, opened 2020-12-01
Add customizable metadata for peer awareness

habibrosyad:
**What type of PR is this?**
/kind api-change
/kind feature

Add customizable metadata for peer awareness. Currently the API only exposes client ID and there's no way to add extra information about the client during collaborative activities.

Fixes https://github.com/yorkie-team/yorkie-js-sdk/issues/106
Related with https://github.com/yorkie-team/yorkie/issues/48

This works around `WatchDocuments` rpc call. In the `WatchDocumentsRequest` I add a new meta field, where you could put whatever peer's metadata there. The metadata then passed along with the `WatchDocumentsResponse`.

```release-note

```

**Additional documentation**:

```docs

```


dc7303 (commented):
Thank you very much for doing a great job! 👍 
I reviewed this PR along with the yorkie-team/yorkie-js-sdk#110 PR. And I tested it and confirmed that it works fine.

I would like to see this wonderful PR and give you some opinions. And I need your feedback :)

We provide `Client` in `Yorkie`. And `Client` can subscribe to events like `yorkie-js-sdk` using `Watch`. So, in accordance with the progress of the yorkie-team/yorkie-js-sdk#110 PR, I hope that the modification of `Client` will proceed in this PR as well. If this happens, I think we can clearly define the test case for this feature.
And when this PR is complete, I would like to review on yorkie-team/yorkie-js-sdk#110.

What do you think? Please tell us your honest opinion.

habibrosyad:
@dc7303 I think I missed the Go client in this repo. Sure, I will update this PR with that as well.

habibrosyad:
@dc7303 @hackerwins I noticed that yorkie-js-sdk is not exposing the [EventType](https://github.com/yorkie-team/yorkie/blob/master/api/yorkie.proto#L356-L360) as described in the proto for client's subscription but using a custom defined [ClientEventType](https://github.com/yorkie-team/yorkie-js-sdk/blob/master/src/core/client.ts#L57-L63). 

On the other hand, the [WatchResponse](https://github.com/yorkie-team/yorkie/blob/master/client/client.go#L264-L268) in the Go client uses EventType. Therefore, somehow the event exposes by these two clients is different. Is it intended?

hackerwins:
@habibrosyad 

It seems that all of the features are not implemented in the Go client yet. As more features are implemented in the JS SDK, I think we should also implement ClientEventType in Go Client.

hackerwins:
We will continue this at #138.

habibrosyad:
@hackerwins thanks for continuing the work. Got busy these days, so haven't had a chance to complete this PR.

hackerwins:
@habibrosyad No worries. This PR allows us to add useful features to our users. Thank you.

PR #112, opened 2020-12-06
Change Priority Queue Release method to replace

### LINKED ISSUE #105: Delete a node from a heap by in-place way
[PriorityQueue.Release](https://github.com/yorkie-team/yorkie/blob/master/pkg/pq/priority_queue.go#L53) was intended to delete the given node from the heap. It was difficult to find and delete the target node from the heap, so we added all nodes to the new heap excluding that node.

And @mojosoeun introduced [a way to delete in-place](https://github.com/yorkie-team/yorkie-js-sdk/pull/101#discussion_r525808950) in the JS SDK. It turns out that deleting a node from a heap in-place way is easier to solve than we thought.

I would like to Go to introduce this method as well.

For more about deleting a node from a heap:
  http://www.mathcs.emory.edu/~cheung/Courses/171/Syllabus/9-BinTree/heap-delete.html

ppeeou:
**What type of PR is this?**
>
> /kind cleanup

Fixes #105

### DIFF
--- a/pkg/pq/priority_queue.go
+++ b/pkg/pq/priority_queue.go
@@ -59,16 +59,38 @@ func (pq *PriorityQueue) Len() int {
 // Release deletes the given value from this PriorityQueue.
 // TODO: It has to be reimplemented in-place.
 func (pq *PriorityQueue) Release(value Value) {
-	queue := &internalQueue{}
-	heap.Init(queue)
+	idx := pq.findIdx(func(item *Item) bool {
+		return item.value == value
+	})
 
-	for _, item := range *pq.queue {
-		if item.value != value {
-			heap.Push(queue, item)
-		}
+	if idx == -1 {
+		return
 	}
 
-	pq.queue = queue
+	// Item to replace
+	lastIdx := (*pq.queue).Len() - 1
+
+	lastItem := (*pq.queue)[lastIdx]
+	lastItem.index = idx
+
+	(*pq.queue)[idx] = lastItem
+	heap.Fix(pq.queue, idx)
+
+	(*pq.queue)[lastIdx] = nil
+	(*pq.queue) = (*pq.queue)[:lastIdx]
+
+	if (*pq.queue).Len() == 0 {
+		pq.queue = &internalQueue{}
+	}
+}
+
+func (pq *PriorityQueue) findIdx(fn func(*Item) bool) int {
+	for i, item := range *pq.queue {
+		if fn(item) {
+			return i
+		}
+	}
+	return -1
 }
 
 // Values returns the values of this PriorityQueue.
--- a/pkg/pq/priority_queue.go
+++ b/pkg/pq/priority_queue.go
@@ -57,7 +57,6 @@ func (pq *PriorityQueue) Len() int {
 }
 
 // Release deletes the given value from this PriorityQueue.
-// TODO: It has to be reimplemented in-place.
 func (pq *PriorityQueue) Release(value Value) {
 	idx := -1
 	for i, item := range *pq.queue {
--- a/pkg/pq/priority_queue.go
+++ b/pkg/pq/priority_queue.go
@@ -67,21 +67,14 @@ func (pq *PriorityQueue) Release(value Value) {
 		return
 	}
 
-	// Item to replace
-	lastIdx := (*pq.queue).Len() - 1
+	lastItem := pq.queue.Pop().(*Item)
 
-	lastItem := (*pq.queue)[lastIdx]
-	lastItem.index = idx
+	if (*pq.queue).Len() == 0 || idx == (*pq.queue).Len() {
+		return
+	}
 
 	(*pq.queue)[idx] = lastItem
 	heap.Fix(pq.queue, idx)
-
-	(*pq.queue)[lastIdx] = nil
-	(*pq.queue) = (*pq.queue)[:lastIdx]
-
-	if (*pq.queue).Len() == 0 {
-		pq.queue = &internalQueue{}
-	}
 }
 
 func (pq *PriorityQueue) findIdx(fn func(*Item) bool) int {
--- a/pkg/pq/priority_queue.go
+++ b/pkg/pq/priority_queue.go
@@ -59,31 +59,19 @@ func (pq *PriorityQueue) Len() int {
 // Release deletes the given value from this PriorityQueue.
 // TODO: It has to be reimplemented in-place.
 func (pq *PriorityQueue) Release(value Value) {
-	idx := pq.findIdx(func(item *Item) bool {
-		return item.value == value
-	})
+	idx := -1
+	for i, item := range *pq.queue {
+		if item.value == value {
+			idx = i
+			break
+		}
+	}
 
 	if idx == -1 {
 		return
 	}
 
-	lastItem := pq.queue.Pop().(*Item)
-
-	if (*pq.queue).Len() == 0 || idx == (*pq.queue).Len() {
-		return
-	}
-
-	(*pq.queue)[idx] = lastItem
-	heap.Fix(pq.queue, idx)
-}
-
-func (pq *PriorityQueue) findIdx(fn func(*Item) bool) int {
-	for i, item := range *pq.queue {
-		if fn(item) {
-			return i
-		}
-	}
-	return -1
+	heap.Remove(pq.queue, idx)
 }
 
 // Values returns the values of this PriorityQueue.
--- a/pkg/pq/priority_queue_test.go
+++ b/pkg/pq/priority_queue_test.go
@@ -17,6 +17,7 @@
 package pq_test
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -93,9 +94,11 @@ func TestPQ(t *testing.T) {
 		assert.EqualValues(t, tmp, []int{3, 4, 5, 6, 7, 8, 9, 10, 11, 12})
 	})
 
+}
+
+func TestPQRelease(t *testing.T) {
 	t.Run("priority queue release", func(t *testing.T) {
 		queue := setUpTestNums()
-
 		for i := 1; i <= 12; i++ {
 			queue.Release(newTestValue(i))
 			assert.False(t, exist(i, queue.Values()))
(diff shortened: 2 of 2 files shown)


pkg/pq/priority_queue.go
```diff
@@ -51,16 +51,34 @@ func (pq *PriorityQueue) Len() int {
...
+		pq.queue = &internalQueue{}
+	}
+}
+
+func (pq *PriorityQueue) findIdx(value Value) int {
 	for _, item := range *pq.queue {
-		if item.value != value {
-			heap.Push(queue, item)
+		if item.value == value {
+			return item.index
```
ppeeou, line 78: I am not sure how to use the index of the Item.

This is because the item in the current code is not entered as an argument value, and the value is entered.

hackerwins:
Thank you. We'll review it soon.

dc7303 (changes requested):
Thank you for your great work! 👍 
There are a few things I would like to discuss. Tell us what you think. 😄 

### Test
What about separating test cases?
I think you can divide test cases by referring to [heap_test.js](https://github.com/yorkie-team/yorkie-js-sdk/blob/master/test/util/heap_test.ts#L33-L89) of `yorkie-js-sdk`.
Separating test specifications allows other developers to analyze them more efficiently. And it's efficient in managing tests.

<details>
<summary>Example</summary>
I tested with max heap.
If you want to test the max heap test case that I have experimented with, you can modify the `Less` method in `priority_queue.go` as follows.

```go
func (value testValue) Less(other pq.Value) bool {
	return value.value > other.(testValue).value
}
```

Test Example ::

```go
	t.Run("root node is deleted test", func(t *testing.T) {
		pq := pq.NewPriorityQueue()

		root := NewTestValue(9)
		for i := 0; i < 10; i++ {
			pq.Push(NewTestValue(i))
		}

		pq.Release(root)

		expected := "[{8} {7} {5} {6} {2} {1} {4} {0} {3}]"
		assert.Equal(t, expected, fmt.Sprint(pq.Values()))
	})

	t.Run("if parent node is deleted", func(t *testing.T) {
		pq := pq.NewPriorityQueue()

		parent := NewTestValue(5)
		for i := 0; i < 10; i++ {
			pq.Push(NewTestValue(i))
		}

		pq.Release(parent)

		expected := "[{9} {8} {4} {6} {7} {1} {2} {0} {3}]"
		assert.Equal(t, expected, fmt.Sprint(pq.Values()))
	})

	t.Run("if leaf node is deleted", func(t *testing.T) {
		pq := pq.NewPriorityQueue()

		parent := NewTestValue(3)
		for i := 0; i < 10; i++ {
			pq.Push(NewTestValue(i))
		}

		pq.Release(parent)

		expected := "[{9} {8} {5} {6} {7} {1} {4} {0} {2}]"
		assert.Equal(t, expected, fmt.Sprint(pq.Values()))
	})

	t.Run("if a heap has one node", func(t *testing.T) {
		pq := pq.NewPriorityQueue()

		node := NewTestValue(0)

		pq.Push(node)
		pq.Release(node)

		assert.Equal(t, 0, pq.Len())
	})
```
</details>

Please tell us your opinion.

--- 

I'd also like to ask @hackerwins. We're using max heap, is there any reason the tests were written with min heap?

pkg/pq/priority_queue.go
```diff
@@ -51,16 +51,38 @@ func (pq *PriorityQueue) Len() int {
...
+
+	(*pq.queue)[idx] = lastItem
+	heap.Fix(pq.queue, idx)
+
+	(*pq.queue)[lastIdx] = nil
+	(*pq.queue) = (*pq.queue)[:lastIdx]
+
+	if (*pq.queue).Len() == 0 {
+		pq.queue = &internalQueue{}
+	}
```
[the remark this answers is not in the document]
ppeeou: thank you for your review!
Now that I think about it..! It wouldn't need to use the following cases.
>  it creates a new memory queue when the queue size reaches zero


But isn't it necessary in this case to prevent memory leaks??
`(*pq.queue)[lastIdx] = nil`

https://github.com/golang/go/wiki/SliceTricks#delete-without-preserving-order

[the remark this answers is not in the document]
dc7303: `(*pq.queue)[lastIdx] = nil`
Oh! I didn't know about this. It seems correct to assign `nil` to the value of the last index. 
Thanks for correcting my mistakes! :)

mojosoeun, line 76: In the `Release` function, how about using the `peek` function to implement this? 🤔
For example, 
```go
idx := pq.findIdx(func(item *Item) bool {
  return item.value == value
})

lastItem := pq.queue.Peek().(*Item)
if idx == -1 || (*pq.queue).Len() == 0 {
  return
}

(*pq.queue)[idx] = lastItem
heap.Fix(pq.queue, idx)
```

pkg/pq/priority_queue_test.go
```diff
@@ -75,6 +78,40 @@ func TestPQ(t *testing.T) {
...
+		for i := 12; i >= 1; i-- {
+			pq.Release(NewTestValue(i))
+			assert.False(t, exist(i, pq.Values()))
+		}
+		assert.Equal(t, 0, pq.Len())
+
+		for i := 1; i <= 12; i++ {
+			pq.Push(NewTestValue(i))
+		}
+
```
mojosoeun, line 107: How about separating some of the test cases in this test suite? If we separate it, in my view, I think the test code will be easier to read :) 

For example, 
t.Run("priority queue peek test")
t.Run("priority queue pop test")
t.Run("priority queue release test")

mojosoeun (commented):
Thank you for your great work 👏. I left a few comments. 

Note: I'm not sure I reviewed it well because I'm not familiar with Golang 😂

pkg/pq/priority_queue_test.go
ppeeou: cool It would be better

pkg/pq/priority_queue.go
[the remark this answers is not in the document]
ppeeou: After removing unnecessary parts, I will request a review again :)

hackerwins:
@dc7303 
From the point of view of the `pq`(priority queue) package, it is not known whether to use the Max Heap or Min Heap from the outside. Probably, at the time of implementation, it seems that the test code was written only considering the behavior of the `pq` package.

pkg/pq/priority_queue.go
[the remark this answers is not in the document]
dc7303: @ppeeou 
I think @mojosoeun gave us a better suggestion, so this suggestion can be closed! :)

dc7303:
@hackerwins Thanks for the feedback!

pkg/pq/priority_queue.go
```diff
@@ -51,16 +51,31 @@ func (pq *PriorityQueue) Len() int {
...
-		if item.value != value {
-			heap.Push(queue, item)
-		}
+	if idx == -1 {
+		return
+	}
+
+	lastItem := pq.queue.Pop().(*Item)
+
+	if (*pq.queue).Len() == 0 || idx == (*pq.queue).Len() {
```
hackerwins, line 64: Since we can execute `Len()` without dereferencing, why not express it simply?
```suggestion
	if pq.queue.Len() == 0 || idx == pq.queue.Len() {
```

pkg/pq/priority_queue.go
```diff
@@ -51,16 +51,31 @@ func (pq *PriorityQueue) Len() int {
...
+	heap.Fix(pq.queue, idx)
+}
+
+func (pq *PriorityQueue) findIdx(fn func(*Item) bool) int {
+	for i, item := range *pq.queue {
+		if fn(item) {
+			return i
+		}
+	}
+	return -1
```
hackerwins, line 78: I think we abstracted it too quickly since there is only one place to use `findIdx` yet. How about inline it now and extract it later when duplicates occur?

pkg/pq/priority_queue.go
```diff
@@ -51,16 +51,31 @@ func (pq *PriorityQueue) Len() int {
...
+
+	lastItem := pq.queue.Pop().(*Item)
+
+	if (*pq.queue).Len() == 0 || idx == (*pq.queue).Len() {
+		return
 	}
 
-	pq.queue = queue
+	(*pq.queue)[idx] = lastItem
+	heap.Fix(pq.queue, idx)
```
hackerwins, line 69: When I first implemented this, I didn't look at the `heap` package properly. Through this PR, I learned that the `heap` already has an element deletion function. Thank you. 👍 

How about using `heap.Remove` directly after finding the index of the element?
```suggestion
	heap.Remove(pq.queue, idx)
```

hackerwins (changes requested):
Sorry for the late review. 🙏 
Through this PR, I was able to look inside the `heap` package. Thank you.

I have a few comments.

hackerwins (approved):
LGTM. Thank you for your work. 
I left a brief comment.

pkg/pq/priority_queue.go
```diff
@@ -59,16 +59,19 @@ func (pq *PriorityQueue) Len() int {
 // Release deletes the given value from this PriorityQueue.
 // TODO: It has to be reimplemented in-place.
```
hackerwins, line 60: It would be nice to delete this comment as well.

```suggestion
```

mojosoeun (approved):

ppeeou:
> Thank you for your great work! 👍
> There are a few things I would like to discuss. Tell us what you think. 😄
> 
> ### Test
> What about separating test cases?
> I think you can divide test cases by referring to [heap_test.js](https://github.com/yorkie-team/yorkie-js-sdk/blob/master/test/util/heap_test.ts#L33-L89) of `yorkie-js-sdk`.
> Separating test specifications allows other developers to analyze them more efficiently. And it's efficient in managing tests.
> 
> Example
> Please tell us your opinion.
> 
> I'd also like to ask @hackerwins. We're using max heap, is there any reason the tests were written with min heap?


Thank you for your review 
I also added a part that is being tested in yorkie-js-sdk

dc7303 (approved):
LGTM 👍  Thank you for your work!

PR #115, opened 2020-12-18
Introduce tools.go

hackerwins:

`tools.go` ensures tool dependencies are kept in sync. This is the recommended way of doing this according to the below.

https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

After running `go mod tidy`, `go.mod` should remain unchanged.

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


tools/tools.go
```diff
@@ -0,0 +1,27 @@
...
+import _ "github.com/gogo/protobuf/gogoproto"
+
+//go:generate go install github.com/gogo/protobuf/protoc-gen-gogo
+import _ "github.com/gogo/protobuf/protoc-gen-gogo"
+
+//go:generate go install github.com/gogo/protobuf/protoc-gen-gofast
+import _ "github.com/gogo/protobuf/protoc-gen-gofast"
+
+//go:generate go install github.com/golangci/golangci-lint/cmd/golangci-lint
+import _ "github.com/golangci/golangci-lint/cmd/golangci-lint"
```
dc7303, line 27: Even if the module is not imported, it is confirmed that the module is installed by `go install`. Can you explain the reason for the import modules?

dc7303 (commented):
Thanks for the wonderful work. 👍  I learned about `go generate` thanks to this work.

I asked a question about the part I was curious about. Please check.

tools/tools.go
hackerwins: Thanks for the comment. Sorry I didn't fully understand the question. But the behavior I understood is below.

Typing `make tools` will run `go generate -tags tools tools/tools.go`. `go generate` runs comments in patterns that begin with `go:generate`. And modules imported into `tools.go` are not used directly in the source code, but are tracked by go mod. For more details please refer below link:

https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

dc7303: Thank you for your feedback! 👍

dc7303 (approved):
LGTM 👍

PR #123, opened 2020-12-24
Connect adjacent insNext and insPrev when running purge

hackerwins:

Connect adjacent insNext and insPrev when running purge.

In #104, GC was also introduced for text nodes marked with tombstones. However, in `release()`, the link between adjacent insPrev and insNext is missing, so this commit fixes it.

block-wise-rga

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/pkg/document/json/rga_tree_split.go
+++ b/pkg/document/json/rga_tree_split.go
@@ -580,22 +580,19 @@ func (s *RGATreeSplit) cleanupRemovedNodes(ticket *time.Ticket) int {
 	return count
 }
 
-// purge removes the node passed as a parameter from RGATreeSplit.
+// purge physically purge the given node from RGATreeSplit.
 func (s *RGATreeSplit) purge(node *RGATreeSplitNode) {
-	if node.prev != nil {
-		node.prev.next = node.next
-		if node.next != nil {
-			node.next.prev = node.prev
-		}
-	} else {
-		s.initialHead, node.next.prev = node.next, nil
+	node.prev.next = node.next
+	if node.next != nil {
+		node.next.prev = node.prev
 	}
-	node.next, node.prev = nil, nil
+	node.prev, node.next = nil, nil
 
 	if node.insPrev != nil {
-		node.insPrev.insNext, node.insPrev = nil, nil
+		node.insPrev.insNext = node.insNext
 	}
 	if node.insNext != nil {
-		node.insNext.insPrev, node.insNext = nil, nil
+		node.insNext.insPrev = node.insPrev
 	}
+	node.insPrev, node.insNext = nil, nil
 }
--- a/pkg/document/json/root_test.go
+++ b/pkg/document/json/root_test.go
@@ -26,33 +26,53 @@ import (
 	"github.com/yorkie-team/yorkie/testhelper"
 )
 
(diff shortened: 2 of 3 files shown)


pkg/document/json/rga_tree_split.go
```diff
@@ -580,22 +580,19 @@ func (s *RGATreeSplit) cleanupRemovedNodes(ticket *time.Ticket) int {
...
-	node.next, node.prev = nil, nil
+	node.prev, node.next = nil, nil
 
 	if node.insPrev != nil {
-		node.insPrev.insNext, node.insPrev = nil, nil
+		node.insPrev.insNext = node.insNext
 	}
 	if node.insNext != nil {
-		node.insNext.insPrev, node.insNext = nil, nil
+		node.insNext.insPrev = node.insPrev
```
dc7303, line 595: I implemented it in a way that breaks the relationship altogether because the split link represents the relationship that occurs when the node is split. I'm curious about a possible problem when not connecting `insPrev` and `insNext`.

pkg/document/json/root_test.go
```diff
@@ -26,33 +26,53 @@ import (
...
-func registerTextElement(fromPos, toPos *json.RGATreeSplitNodePos, root *json.Root, text json.TextElement) {
+func registerRemovedNodeTextElement(fromPos, toPos *json.RGATreeSplitNodePos, root *json.Root, text json.TextElement) {
 	if fromPos.Compare(toPos) != 0 {
 		root.RegisterRemovedNodeTextElement(text)
 	}
 }
 
 func TestRoot(t *testing.T) {
-	t.Run("garbage collect for text test", func(t *testing.T) {
+	t.Run("garbage collection for array test", func(t *testing.T) {
```
dc7303, line 36: This test seems to be out of scope for this PR. Any reason added in this PR?

dc7303 (commented):
Thank you for your great work! 👍  I had a question, so I left a comment. Please check.

pkg/document/json/root_test.go
hackerwins: Disconnecting `node.prev` and `node.next` is missing when purging a node from RGATreeList. As I added related tests in this PR, I changed the names of other adjacent tests as well.

pkg/document/json/rga_tree_split.go
hackerwins: `insPrev` and `insNext` are used to remember other nodes connected to the insert. For example, when `abc` is divided into `a`, `b`, `c`, it looks like this: `[abc]` divided to `[a]<->[b]<->[c]` Even if we completely delete `b`, I think it should keep the relationship of `[a]` <-> `[c]`.

`insPrev` is used by `findFloorNodePreferToLeft`. I think that it will not work properly when looking for the left node with priority from the given location.

https://github.com/yorkie-team/yorkie/blob/master/pkg/document/json/rga_tree_split.go#L342

dc7303: Oh! Thank you for the clear explanation :D

dc7303 (approved):
LGTM 👍  Thanks to you, I learned a lot!

PR #124, opened 2020-12-24
Fix error log for ErrNoDocuments

dc7303:
err is always nil because it has already been checked before this logic.
And it is judged that it is correct to print the error currently being
checked to the on.

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins (approved):
LGTM. Thanks for the correction.

PR #129, opened 2020-12-25
Wrap additional information to sentinel errors

hackerwins:
**What this PR does / why we need it**:

Previously, if there was no client or document, only sentinel errors were returned. This commit wraps sentinel errors and returns additional information, such as the document's key.

I implemented this code by referring to the below.
https://blog.golang.org/go1.13-errors

And the code that returns `status.error` from `error` was duplicated, so I extracted it as a function, `toStatusError`.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Added relevant tests or not required
- [x] Didn't break anything

PR #130, opened 2020-12-26
Extract db interface

hackerwins:

Extract `db` interface.
This commit separates high-level package, `db` from low-level `mongo` at the level of abstraction.

Fixes #

- We need to merge #129 first.
- `db` still has a dependency on a low-level package `go.mongodb.org/mongo-driver/bson/primitive`. We need to replace `primitive.ObjectID` with `string` later to remove the dependency. https://github.com/yorkie-team/yorkie/pull/130/files#diff-735613584d9d5d8a97e968e7efa3bcb9999770e56490789587ea33a0827ba2a9R7

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


dc7303:
@hackerwins 
So cool! When #129 is finished, I'll review and comment!

---

> db still has a dependency on a low-level package go.mongodb.org/mongo-driver/bson/primitive. We need to replace primitive.ObjectID with string later to remove the dependency.

I agree with this comment. Actual `primitive.ObjectID` is used by several entities. It seems dangerous for entities to dependency on low-level. So, after this PR, it would be nice to register an issue on this issue.

hackerwins:
@dc7303 I merged #129 into the master branch and rebase it.

dc7303 (approved):
LGTM 👍  It feels good to see the project stabilizing while organizing one by one like this.

PR #131, opened 2020-12-27
Use status for one-dimensional value

hackerwins:

Use status for one-dimensional value.
When I first implemented Yorkie, I named variables without distinction between `status` and `state`. To improve consistency, we use `status` to represent `one-dimensional` values.

I referenced the following discussion.

- https://stackoverflow.com/questions/1162816/naming-conventions-state-versus-status

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -200,7 +200,7 @@ func (c *Client) Attach(ctx context.Context, doc *document.Document) error {
 		return err
 	}
 
-	doc.UpdateState(document.Attached)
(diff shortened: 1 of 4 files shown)


dc7303 (approved):
LGTM 👍 Thanks to the link you attached, I learned the difference between State and Status. Thank you!

PR #132, opened 2020-12-27
Remove panics from the converter

### LINKED ISSUE #50: Remove panic from server code

hackerwins:

Remove panics from the converter.

If an error occurs in the converter, the converter must return an error
and allow the caller to decide how to handle it.

In particular, when a request that is not supported by the old server is
given, the server should return an error response instead of being
terminated.

This commit is based on the recommendations below.
https://github.com/uber-go/guide/blob/master/style.md#dont-panic

Fixes #50

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Didn't break anything

### DIFF
--- a/pkg/document/proxy/text_proxy.go
+++ b/pkg/document/proxy/text_proxy.go
@@ -70,3 +70,30 @@ func (p *TextProxy) Edit(from, to int, content string) *TextProxy {
 	}
 	return p
 }
+
+// Select stores that the given range has been selected.
+func (p *TextProxy) Select(from, to int) *TextProxy {
+	if from > to {
+		panic("from should be less than or equal to to")
+	}
+	fromPos, toPos := p.Text.CreateRange(from, to)
+	log.Logger.Debugf(
+		"SELT: f:%d->%s, t:%d->%s",
+		from, fromPos.AnnotatedString(), to, toPos.AnnotatedString(),
+	)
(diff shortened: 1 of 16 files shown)


pkg/document/proxy/text_proxy.go
```diff
@@ -70,3 +70,30 @@ func (p *TextProxy) Edit(from, to int, content string) *TextProxy {
 	}
 	return p
 }
+
+// Select stores that the given range has been selected.
+func (p *TextProxy) Select(from, to int) *TextProxy {
```
dc7303, line 75: It seems to be outside the scope of the current PR title and content. I think we can revise the title and content so we can see that this feature has been added to this PR.
Let me know what you think.

dc7303 (approved):
LGTM :) That's great work!
I gave you an opinion. please check.

pkg/document/proxy/text_proxy.go
hackerwins: Thank you for the review. I have added the missing features to increase coverage. I'll add it to the body of this commit.

PR #139, opened 2021-01-03
Add peer awareness design document

### LINKED ISSUE #48: Introduce Peer Awareness

hackerwins:

Writing a design document for Peer Awareness.

Fixes #48

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/design/peer-awareness.md
+++ b/design/peer-awareness.md
@@ -1,12 +1,30 @@
+---
+title: peer-awareness
+target-version: 0.1.2
+---
+
 # Peer Awareness
 
-Target version: Yorkie 0.1.2
+## Summary
 
-## Use case
+We will provide Peer Awareness which is a simple algorithm that manages end-user
+status like who is connected and metadata like username or email address and etc.
+For example, users can implement a list of people participating in the editing,
+such as a box in the top right of Google Docs.
 
-When multiple users edit a single document, the users want to know who is connected and what is being edited by the peers. For example, in Google Docs, the ID and profile of the user editing together are displayed in the upper-right.
+### Goals
 
-## How to use
+Implement Peer Awareness and provide API to users to use the feature. The goal of
+the first version is to implement simple functionality and check usability.
+
+### Non-Goals
+
+The first version does not implement complex features such as dynamic metadata
+updates.
+
+## Proposal details
+
+### How to use
 
 Users can pass metadata along with client options when creating a client.
 
@@ -25,7 +43,8 @@ const doc = yorkie.createDocument('examples', 'codemirror');
 await client.attach(doc);
 ```
 
-When a new peer registers or leaves, `peers-changed` event is fired, and the other peer's clientID and metadata can be obtained from the event.
+When a new peer registers or leaves, `peers-changed` event is fired, and the
+other peer's clientID and metadata can be obtained from the event.
 
 ```typescript
 client.subscribe((event) => {
@@ -38,7 +57,7 @@ client.subscribe((event) => {
 });
 ```
 
-## How does it work?
+### How does it work?
 
 ```
  +--Client "A"----+                      +--Agent-----------------+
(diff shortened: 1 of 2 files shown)


design/peer-awareness.md
```diff
@@ -0,0 +1,64 @@
+# Peer Awareness
+
+Target version: Yorkie 0.1.2
```
dc7303, line 3: I am not sure if it is absolutely necessary to specify the version. Can you explain why you entered the version?

design/peer-awareness.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Use case
```
dc7303, line 5: What about `Overview` or `Goals` instead of `Use case`? I think that this section is more of an explanation of what problems `peer awareness` solves, rather than talking about use cases. What do you think?

dc7303 (commented):
This is really cool!
It seems to be well written for easy understanding. I gave some opinions, so please check.

design/peer-awareness.md
hackerwins: I thought writing a release version would allow us to think about the scope of features and focus more on our goals when writing your design documentation.

design/peer-awareness.md
hackerwins: I think the goals could be suitable as well. But I wanted to write what value it would give the user with Peer Awareness.

habibrosyad: Agree with @dc7303. Additionally I thought `Motivation` would suit this section as well, since this section mentions what's driving this feature.

habibrosyad (commented):
In my opinion, this design document is great, simple and easy to understand.

For design document I think we can introduce more solid and consistent structures (especially later when some collaborator/contributor want to convey their ideas for enhancing yorkie with more features). I thought a good reference for this would be the KEP (Kubernetes Enchancement Proposal), e.g. [this one](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1281-network-proxy).

hackerwins:
@dc7303, @habibrosyad Thank you for your review. I adjusted the section to reflect your comments.

Also, @dc7303 commented that the first document will be referenced in future documents, so it's a good idea to handle it carefully. So I added a template too.

design/peer-awareness.md
```diff
@@ -0,0 +1,95 @@
...
+ | | { num: 1 } | <-- WatchDocuments --> | | {                  | |
+ | +------------+ |                      | |   docA: {          | |
+ +----------------+                      | |     A: { num: 1 }, | |
+                                         | |     B: { num: 2 }  | |
+ +--Client "B"----+                      | |   },               | |
+ | +--Metadata--+ |                      | |   ...              | |
+ | | { num: 2 } | <-- WatchDocuments --> | | }                  | |
+ | +------------+ |                      | +--------------------+ |
+ +----------------+                      +------------------------+
+```
```
dc7303, line 74: I think we need to turn a diagram written in text like this into an image file. But I don't think it is now. When the branding of our project is done, it looks like we can draw it by applying our colors, designs, and feelings.
This can be a cumbersome task, but it seems that even if there are more documents, you can have unity. What do you think?

dc7303 (approved):
LGTM 👍 I think the written template and the `peer-awareness.md` document are appropriate as our first design document.

design/peer-awareness.md
hackerwins: Thanks for the comment. In my opinion, there seems to be no need to unify the style of diagrams in internal documents. However, it would be nice to apply it to the documentation we write on https://yorkie.dev.

mojosoeun (approved):
Thank you for your work! The design document and template are awesome and easy to understand! Also, I was able to learn about `peer-awareness`!

PR #140, opened 2021-01-07
Fix non-deterministic test coverage of RGATreeSplit

hackerwins:

Fix non-deterministic test coverage of RGATreeSplit.

Problems that seem to reduce test coverage were caused by non-deterministic factors in the test code. I added a test so that it is always done because the cause cannot be found.

Screen Shot 2021-01-08 at 4 20 49 AM
Screen Shot 2021-01-08 at 4 22 13 AM

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/pkg/document/json/root_test.go
+++ b/pkg/document/json/root_test.go
@@ -93,6 +93,38 @@ func TestRoot(t *testing.T) {
 		assert.Equal(t, 1, nodeLen)
 	})
 
+	t.Run("garbage collection for fragments of text", func(t *testing.T) {
+		type test struct {
+			from    int
+			to      int
+			content string
+			want    string
+			garbage int
+		}
+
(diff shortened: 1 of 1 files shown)


dc7303 (approved):
LGTM 👍 
I also looked for why non-deterministic test coverage occurs, but I haven't solved it yet.

When we receive a PR, we think we need this test code to not affect that PR, but there are also concerns that we may hide unintended behavior.
So, I think it would be a good idea to add relevant comments to the test code and register an issue to prevent us from forgetting this issue.
What is your opinion?

hackerwins:
@dc7303 
Yes. Thanks for the good comment. It would be better to leave it as an issue rather than a comment.

Issue #11, opened 2020-01-23
Support HA servers in active-active manner


hackerwins:
Currently Yorkie only provides a single server. Servers can always be down, we need to set up the HA servers. Unlike other servers, Yorkie will receive many edit operations, so we will configure it in an active-active manner.

If we handle the following tasks, It seems like we can achieve set up HA servers.

- [x] Introduce distributed lock with lease
- [ ] Introduce distributed PubSub

hackerwins:
If we introduce the ETCD adapter, We seem to be able to handle this simply.

hackerwins:
## Distributed lock performance test using ETCD

### Description of test conditions
- Multiple routines access the same key at the same time.
- ETCD is also run on the machine running the test.
- Machine info
  - MacBook Air (Retina, 13-inch, 2020)
  - Processor 1.1 GHz Quad-Core Intel Core i5
  - Memory 16 GB 3733 MHz LPDDR4X

```go
start := time.Now()

size := 500
sum := 0
var wg gosync.WaitGroup
for i := 0; i < size; i++ {
	wg.Add(1)
	go func() {
		defer wg.Done()

		ctx := context.Background()
		cli.Lock(ctx, sync.Key(t.Name()))
		sum += 1
		cli.Unlock(ctx, sync.Key(t.Name()))
	}()
}
wg.Wait()
assert.Equal(t, size, sum)

log.Logger.Infof("lock count: %d, elapsed: %s", size, time.Since(start))
```

### Results

```
// Performance of sync.Mutex (Go)
2021-02-12T23:11:44.384+0900	INFO	lock count: 10, elapsed: 177.267µs
2021-02-12T23:11:57.554+0900	INFO	lock count: 100, elapsed: 331.794µs
2021-02-12T23:06:24.930+0900	INFO	lock count: 1000, elapsed: 2.271639ms

// Performance of concurrency.Mutex (ETCD)
2021-02-12T23:11:27.319+0900	INFO	lock count: 10, elapsed: 199.070746ms
2021-02-12T23:10:52.406+0900	INFO	lock count: 100, elapsed: 1.133782094s
2021-02-12T23:14:37.825+0900	INFO	lock count: 200, elapsed: 3.849113763s
2021-02-12T23:15:32.183+0900	INFO	lock count: 500, elapsed: 14.822853146s
2021-02-12T23:01:29.410+0900	INFO	lock count: 1000, elapsed: 48.902339709s
```

[...]

PR #141, opened 2021-01-09
Extract MemPubSub from PubSub

hackerwins:

Extract MemPubSub from PubSub

Before implementing the distributed PubSub implementation using ETCD, we
need to extract the temporarily implemented MemoryPubSub and leave only
the high-level abstraction logic in PubSub.

Address #11

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/yorkie/backend/pubsub/mempubsub/mempubsub.go
+++ b/yorkie/backend/pubsub/mempubsub/mempubsub.go
@@ -1,0 +1,156 @@
+/*
+ * Copyright 2020 The Yorkie Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package mempubsub
+
+import (
+	"sync"
+
+	"github.com/yorkie-team/yorkie/pkg/document/time"
+	"github.com/yorkie-team/yorkie/pkg/log"
+	"github.com/yorkie-team/yorkie/pkg/types"
+	"github.com/yorkie-team/yorkie/yorkie/backend/pubsub"
+)
+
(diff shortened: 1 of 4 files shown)


yorkie/backend/pubsub/mempubsub/mempubsub.go
```diff
@@ -0,0 +1,156 @@
...
+type MemPubSub struct {
+	mu               *sync.RWMutex
+	subscriptionsMap map[string]*Subscriptions
+}
+
+// New creates an instance of MemoryPubSub.
+func New() *MemPubSub {
+	return &MemPubSub{
+		mu:               &sync.RWMutex{},
+		subscriptionsMap: make(map[string]*Subscriptions),
```
dc7303, line 69: Both `subscriptions` and `Map` have multiple data representations. So `subscriptionMap` or `subscriptions` looks good, what do you think?

yorkie/backend/pubsub/mempubsub/mempubsub.go
```diff
@@ -0,0 +1,156 @@
...
+
+	"github.com/yorkie-team/yorkie/pkg/document/time"
+	"github.com/yorkie-team/yorkie/pkg/log"
+	"github.com/yorkie-team/yorkie/pkg/types"
+	"github.com/yorkie-team/yorkie/yorkie/backend/pubsub"
+)
+
+// Subscriptions is a collection of subscriptions that subscribe to a specific
+// topic.
+type Subscriptions struct {
```
dc7303, line 30: The `Subscriptions` doesn't seem to need to be known from the external package. So I think it would be okay to replace it with `subscriptions`.
Could you please tell me why you declared `Subscriptions` as a public structure?

dc7303 (changes requested):
That's a great job! Abstraction work is always enjoyable and beneficial to us.
It seems that there will be no problem when introducing a key-value store in the future.

I have given you some opinions, so please check.

yorkie/backend/pubsub/mempubsub/mempubsub.go
hackerwins: Thank you. I think it would be better to change it.

yorkie/backend/pubsub/mempubsub/mempubsub.go
hackerwins: Thanks for the comments. But `subscriptionsMap` represents a map of `subscriptions` not a map of `subscription`. 
So `subscriptionMap` or `subscriptions` is not appropriate.

dc7303 (approved):
LGTM 👍  Thank you for confirming my suggestion.

PR #148, opened 2021-02-12
Employ etcd for distributed locking

hackerwins:

Employ etcd for distributed locking

In order to allow multiple agents to process client requests, etcd is
employed. In standalone mode with only one agent, memory-based locker is
used, not etcd.

Address #11

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins:
## Distributed lock performance test using ETCD

### Description of test conditions
- Multiple routines access the same key at the same time.
- ETCD is also run on the machine running the test.
- Machine info
  - MacBook Air (Retina, 13-inch, 2020)
  - Processor 1.1 GHz Quad-Core Intel Core i5
  - Memory 16 GB 3733 MHz LPDDR4X

```go
start := time.Now()

size := 500
sum := 0
var wg gosync.WaitGroup
for i := 0; i < size; i++ {
	wg.Add(1)
	go func() {
		defer wg.Done()

		ctx := context.Background()
		cli.Lock(ctx, sync.Key(t.Name()))
		sum += 1
		cli.Unlock(ctx, sync.Key(t.Name()))
	}()
}
wg.Wait()
assert.Equal(t, size, sum)

log.Logger.Infof("lock count: %d, elapsed: %s", size, time.Since(start))
```

### Results

```
// Performance of sync.Mutex (Go)
2021-02-12T23:11:44.384+0900	INFO	lock count: 10, elapsed: 177.267µs
2021-02-12T23:11:57.554+0900	INFO	lock count: 100, elapsed: 331.794µs
2021-02-12T23:06:24.930+0900	INFO	lock count: 1000, elapsed: 2.271639ms

// Performance of concurrency.Mutex (ETCD)
2021-02-12T23:11:27.319+0900	INFO	lock count: 10, elapsed: 199.070746ms
2021-02-12T23:10:52.406+0900	INFO	lock count: 100, elapsed: 1.133782094s
2021-02-12T23:14:37.825+0900	INFO	lock count: 200, elapsed: 3.849113763s
2021-02-12T23:15:32.183+0900	INFO	lock count: 500, elapsed: 14.822853146s
2021-02-12T23:01:29.410+0900	INFO	lock count: 1000, elapsed: 48.902339709s

// Performance of concurrency.Mutex (ETCD, When accessing different keys) 
2021-02-12T23:21:36.212+0900	INFO	lock count: 500, elapsed: 997.036022ms
```

ppeeou (approved):
LGTM 🔥

dc7303:
Sorry, I am not familiar with ETCD, so there is a delay in reviewing it. 😭 
I will take a closer look and give you an opinion.

mojosoeun:
@hackerwins Thank you for your PR! Could you elaborate on this PR in the weekly meeting? Because im also not used to etcd, it would be great to hear about this task!

yorkie/backend/sync/etcd/client.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// Client is a client that connects to ETCD.
+type Client struct {
+	config *Config
+	client *clientv3.Client
+}
+
+// NewClient creates a new instance of Client.
+func NewClient(conf *Config) *Client {
```
dc7303, line 50: I seem to be able to change this method to a private method.
Currently, it is called in the `TestETCDStress` test of `etcd_stress_test.go`, but it seems that it can be replaced with `Dial(conf *Config)`.

yorkie/rpc/server.go
```diff
@@ -217,18 +230,31 @@ func (s *Server) DetachDocument(
 	}
 
 	// if pack.HasChanges() {
-	lockKey := fmt.Sprintf("pushpull-%s", pack.DocumentKey.BSONKey())
-	if err := s.backend.MutexMap.Lock(lockKey); err != nil {
+	locker, err := s.backend.MutexMap.NewLocker(
+		ctx,
+		sync.NewKey(fmt.Sprintf("pushpull-%s", pack.DocumentKey.BSONKey())),
```
dc7303, line 235: The `fmt.Sprintf("pushpull-%s", pack.DocumentKey.BSONKey())` code seems to be used in several places. How about subtracting with a separate function?

dc7303 (commented):
Thanks for the wonderful work. Thanks to you, I am learning a lot. 🙇‍♂️ 

I am trying to fully understand this feature and have made a few comments along the way. Please check.

yorkie/rpc/server.go
hackerwins: Yes, this is a reasonable opinion. However, the abstraction for Key seems to be early now, so how about proceeding later?

dc7303: Yes, as you said, it would be better to proceed later :D Thank you for your opinion.

hackerwins:
Since we are using [lease](https://en.wikipedia.org/wiki/Lease_(computer_science)) based lock, we need to introduce fencing token.
Lease lock and fencing token are described in [the article by Martin Kleppmann](https://martin.kleppmann.com/2016/02/08/how-to-do-distributed-locking.html)

yorkie/backend/sync/etcd/client.go
```diff
@@ -0,0 +1,141 @@
...
+	})
+	if err != nil {
+		log.Logger.Error(err)
+		return err
+	}
+	log.Logger.Infof("etcd connected, URI: %s", c.config.Endpoints)
+
+	c.client = cli
+	return nil
+}
```
dc7303, line 87: How about checking the actual connection?
Currently, it is not possible to check the connection status to the ETCD endpoint.

### If not intended:

Actually, I tested `TestETCD` of `etcd_test.go` without running `ETCD` docker. And even if it couldn't actually be connected, it was confirmed that the test passed.
Even if you run `make test` in the absence of `ETCD` docker, the test will not proceed somewhere and will be in a standby state. (**Where the problem occurs is another matter.**)

While I analyzing this problem, I found that when I wrote the code below, I could check the connection.
[Reference]: https://github.com/etcd-io/etcd/issues/9877

In the issue I referenced, `[]grpc.DialOption{grpc.WithBlock()}` was mentioned, but in our code it did not work as intended. If we look into the reasons together, it seems that simpler code can solve this problem.
```go
func (c *Client) Dial() error {
	cli, err := clientv3.New(clientv3.Config{
		Endpoints:   c.config.Endpoints,
		DialTimeout: c.config.DialTimeoutSec * time.Second,
		Username:    c.config.Username,
		Password:    c.config.Password,
	})
	if err != nil {
		log.Logger.Error(err)
		return err
	}

	timeoutCtx, cancel := context.WithTimeout(context.Background(), c.config.DialTimeoutSec*time.Second)
	defer cancel()

	_, err = cli.Status(timeoutCtx, c.config.Endpoints[0])
	if err != nil {
		log.Logger.Error(err)
		return err
	}
        ...
	return nil
}
```

README.md
```diff
@@ -87,7 +87,7 @@ $ bin/yorkie
 
 Tests can be run by typing `make test`.
 
-*NOTE: `make test` includes integration tests that require a local MongoDB listen on port 27017. To start MongoDB, type `docker-compose up`.*
+*NOTE: `make test` includes integration tests that require a local MongoDB listen on port 27017. To start MongoDB, type `docker-compose -f docker/docker-compose up`.*
```
dc7303, line 90: Any reason not to mention `docker-compose-cluster-mode.yml`?

Our integration tests include ETCD and Prometheus.
In the actual `TestYorkie()` of `helper.go`, `Metrics` and `ETCD` option values ​​are added, and an instance is created with `yorkie.New()` function.  And at this time, `yorkie.New()` creates an actual related instance.

So I think it's more efficient to recommend the `docker-compose-cluster-mode.yml` file to the developer.

Let me know what you think.

dc7303 (changes requested):
I have a few inquiries and suggestions!
Please answer after checking. :)

yorkie/backend/sync/etcd/client.go
hackerwins: Thank you for telling me. I modified it by referring to the PR below.
https://github.com/kubernetes/kubernetes/pull/81435

dc7303 (approved):
LGTM 👍  I learned a lot while reviewing. Thank you! :)

PR #154, opened 2021-02-27
Add PubSub unit tests and code cleanup

hackerwins:

Add PubSub unit tests and code cleanup.
Before implementing etcd-based PubSub, add test code to check whether it works properly.

Address #11

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


dc7303 (commented):
I don't think there is any problem with the current work. :+1:
However, I would like to offer you one suggestion, so I will comment.

The code of the added `pubsub_test.go` are testing the details of the `memory` package. So I think it should be managed in the `sync/memory` directory.
If we add tests for `pubsub` in `etcd`, we can manage them in `sync/etcd` directory, and we have the advantage of being able to manage the details separately.

Please tell me your opinion :)

hackerwins:
@dc7303 Thanks for the comments. I think it's a meaningful suggestion.

dc7303 (approved):
LGTM 👍

PR #164, opened 2021-03-24
Remove errors that occur when insPrev does not exist

hackerwins:

Remove errors that occur when `insPrev` does not exist.

InsPrev can be removed due to GC, so remove the panic code added as an assertion.

Fixes yorkie-team/yorkie-codepair#71

Text implementation in Agent is used by Agent, and unlike JS SDK, it does not expose `Change` events. Therefore, the error related to `insPrev` could not be reproduced.

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/pkg/document/document.go
+++ b/pkg/document/document.go
@@ -176,6 +176,14 @@ func (d *Document) RootObject() *json.Object {
 	return d.doc.RootObject()
 }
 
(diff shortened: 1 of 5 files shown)


dc7303 (approved):
LGTM 👍

PR #169, opened 2021-03-31
Change PushPull changes size metric type

dc7303:
The accumulated value of the counter type is changed to a gauge because the practical utility is poor.
Using the Gauge, we can check the point when the size of the change suddenly increases, like the attached image.

image

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins (approved):
LGTM. 👍

PR #174, opened 2021-04-13
add a .dockerignore to lighten docker image

computerphilosopher:

This PR adds a '.dockerignore' file. This file's purpose is similar to '.gitignore'. Docker will ignore files in the list when building a container image. It can help to lighten the docker image.

reference: https://docs.docker.com/engine/reference/builder/#dockerignore-file
 
Files that related an item of the following list will be ignored

- git, github
- CI
- document
- integration test

Fixes #

```release-note
NONE
```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+.git
+.gitignore
+.github
+.golangci.yml
+*.md
+design
+LICENSE
+test


hackerwins:
@computerphilosopher Thanks for the contribution.

Using .dockerignore seems to prevent unnecessary files from blending into the image. But we are using [multi-stage build](https://github.com/yorkie-team/yorkie/blob/main/Dockerfile#L29) and only add the binary to the image. I don't think it is necessary to reduce the image size.

Are there any other benefits?

computerphilosopher:
I tested both with and without dockerignore. Image size is not reduced actually as you expected. 

I will close PR until I could prove dockerignore's benefit.

PR #181, opened 2021-04-24
Improve Peer Awareness's metadata to be updatable

### LINKED ISSUE #153: Improve Peer Awareness's metadata to be updatable

Realtime collaborative editors usually need to propagate volatile data, such as cursor position to other peers. This type of data needs to be propagated, but not stored.

For now, we implemented [Peer Awareness](https://github.com/yorkie-team/yorkie/blob/main/design/peer-awareness.md) to propagate the metadata of the client, but updates are not possible. We can cover this scenario by improving the Peer Awareness's metadata so that it can be updated.

**Why is this needed**:

The cost of storing data is expensive, so it is economical to propagate this type of data but not store it.

dc7303:
Added UpdateMetadata API to update the client's metadata.

Fixes #153

If the added API design is ok, I will also work on `yorkie-js-sdk`. I'll add a `How to use?` section in the design document later.

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


yorkie/backend/sync/pubsub.go
```diff
@@ -78,6 +78,11 @@ func (s *Subscription) SubscriberID() string {
 	return s.subscriber.ID.String()
 }
 
+// UpdateMetadata updates metadata of subscriber.
+func (s *Subscription) UpdateMetadata(subscriber types.Client) {
+	s.subscriber.Metadata = subscriber.Metadata
+}
```
ppeeou, line 84: Is there any need to lock the mutex in this part🤔??

hackerwins: It would be nice to write test cases that modify the metadata at the same time.

dc7303: @ppeeou @hackerwins Thank you for your comment!

This is the reason why the method was not locked.
To access `Subscription`, you can access it through `PubSub.subscriptionsMap`. And since `subscriptionsMap` is private, I thought that PubSub's mutex would be sufficient. This is because all PubSub methods use the same mutex. As a result, it was determined that it is safe to call other `PubSub` methods at the same time.

At the same time, when modifying the metadata, it was confirmed that the order in which the metadata is updated and the order in which it is published does not match. The UpdateMetadata API has been improved so that can be locked by the client ID. (c92ff31)

dc7303:
> Is there any reason to use `wg.Wait()` here??🤔

Oh! My mistake. thank you for telling me :)

ppeeou (approved):
LGTM!

dc7303:
@hackerwins @ppeeou 
Is it okay if I do PR on js-sdk as well?
It would be nice to test it together.

hackerwins:
I would like to do this work together after the distributed pubsub implementation.

PR #188, opened 2021-05-21
Add basic behavior of authorization webhook

### LINKED ISSUE #182: Provide Authorization Webhook

Provide authorization WebHook.

Webhook is an HTTP POST that is called when something happens. When specified, Authorization Webhook causes Agent to query an outside REST service when determining user privileges.

We can specify the webhook by the `--authorization-webhook=SOME_URL` flag:
```bash
./bin/yorkie agent --authorization-webhook=http://localhost:3000/auth-hook
```

And we can pass tokens when creating a client.
```typescript
yorkie.createClient('http://api.yorkie.dev', {
  token: SOME_TOKEN,
});
```

When the client sends every request, it passes the token to Agent. The Agent who receives the token calls WebHook before processing the requests.

We need to define the payload of the webhook:

```javascript
// request
{
  token: '',
  documentAttributes: [{
    key: 'col1$doc1',
    verb: 'r' // 'r' or 'rw'
  }, {
    key: 'col1$doc2',
    verb: 'rw'
  }]
}

// response
{
  allowed: false,
  reason: 'user does not have read access to the documents'
}
```

AuthorizationWebhook

flags | description | default
-----|-------------|--------
authorization-webhook | URL of remote service to query. | `null`
authorization-methods | Target methods to check authorization e.g: `AttachDocument,DetachDocument,WatchDocuments,PushPull` | `all`
authorization-cache-ttl | The duration to cache 'authorized' responses from the webhook authorizer. | `0s`

References

- https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/

**Why is this needed**:
- We can determine user privileges for documents.

hackerwins:

Add authorization webhook. In this PR, we only implement the default behavior.

- [x] Delete `RequestHeader` because we can carry the version in gRPC metadata.
- [x] Implement basic authorization webhook.
- [x] Add access attributes such as key and verb to webhook request payload.

Fixes #182

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/pkg/types/auth_webhook.go
+++ b/pkg/types/auth_webhook.go
@@ -16,15 +16,44 @@
 
 package types
 
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+)
+
 // VerbType represents an action taken on the document.
 type VerbType string
 
 const (
+
 	// Read represents the case of only reading the given document.
 	Read VerbType = "r"
 
 	// ReadWrite represents the case of reading and writing the given document.
 	ReadWrite VerbType = "rw"
+)
+
+var (
+	// ErrInvalidWebhookRequest is returned when the given webhook request is not valid.
+	ErrInvalidWebhookRequest = errors.New("invalid authorization webhook request")
+
+	// ErrInvalidWebhookResponse is returned when the given webhook response is not valid.
+	ErrInvalidWebhookResponse = errors.New("invalid authorization webhook response")
+)
+
+// Method represents a method name of RPC.
+type Method string
+
+// Belows are the names of RPCs.
+const (
+	ActivateClient   Method = "ActivateClient"
+	DeactivateClient Method = "DeactivateClient"
+	AttachDocument   Method = "AttachDocument"
+	DetachDocument   Method = "DetachDocument"
+	PushPull         Method = "PushPull"
 )
 
 // AccessAttribute represents an access attribute.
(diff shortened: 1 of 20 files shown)


hackerwins:
The following additional work will be done in the follow-up PR.

- [ ] Implement retry and retryBackoff.
- [ ] Add `authorization-methods` flags.
- [ ] Add `authorization-cache-ttl` flags (LRUExpireCache).
- [ ] Provide a way to refresh the token.

mojosoeun (approved):

pkg/types/auth_webhook.go
```diff
@@ -0,0 +1,124 @@
...
+	Token      string            `json:"token"`
+	Method     Method            `json:"method"`
+	Attributes []AccessAttribute `json:"attributes"`
+}
+
+// NewAuthWebhookRequest creates a new instance of AuthWebhookRequest.
+func NewAuthWebhookRequest(reader io.Reader) (*AuthWebhookRequest, error) {
+	req := &AuthWebhookRequest{}
+
+	bytes, err := ioutil.ReadAll(reader)
```
ppeeou, line 82: How about using `io.Copy`??🧐
I think this part is likely to be called frequently
https://daryl-ng.medium.com/why-you-should-avoid-ioutil-readall-in-go-e6be4de180f8

ppeeou (approved):
LGTM🚀

pkg/types/auth_webhook.go
hackerwins: @ppeeou I think your opinion is meaningful.
So I changed `ioutil.ReadAll` to `Decoder.Decode` by referring to the link below.
https://stackoverflow.com/questions/52539695/alternative-to-ioutil-readall-in-go/52540512

dc7303 (approved):
LGTM 👍

PR #189, opened 2021-05-24
Add document change event broadcast using members list

### LINKED ISSUE #183: Introduce distributed PubSub

Introduce distributed PubSub.

Screen Shot 2021-05-05 at 4 56 53 PM

- [x] Store a member list(agent list) of a cluster individually
- [x] Broadcast DocChanged events using the member list
- [x] Build a PeersMap across the cluster

**Why is this needed**:
- We can create a Highly Available Yorkie cluster.
- We can split the client's PushPull request load across multiple Agents.

dc7303:
It broadcasts document change events to replicated agents.

- [x] Add grpc broadcast service.
- [x] Implement the Publish API.
- [x] Broadcast client management.
- [x] ~Add grpc_recover.~
 - When requesting from the server to the server, the server should not be stopped when the replicated agent is in a state where communication is not possible.
- [ ] Add stress test
- [ ] Add design document for broadcast
- [x] Code cleanup

Added issue
- [ ] If you enable the server in multiple tests for the cluster mode test, they affect each other and fail.
example)
```bash
 Error Trace:    cluster_stress_test.go:32
                Error:          Received unexpected error:
                                listen tcp :21501: bind: address already in use
---
 Error Trace:    cluster_mode_test.go:59
                                cluster_mode_test.go:128
                Error:          "map[c2oi......]" should have 2 item(s), but has 7


```

Address #183 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


dc7303:
https://github.com/yorkie-team/yorkie/blob/a1b8407dfed5e9a1389e62e7bf98a1831f0ce629/test/helper/helper.go#L110-L136

Port conflicts occur when multiple packages use this function because it recompiles when the package being tested is replaced. I am looking for a way to use ports stably to run multiple servers.
[go-test doc](https://golang.org/pkg/cmd/go/internal/test/)
>'Go test' recompiles each package along with any files with names matching
the file pattern "*_test.go"
....

yorkie/rpc/broadcast.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+
+	"github.com/yorkie-team/yorkie/api"
+	"github.com/yorkie-team/yorkie/api/converter"
+	"github.com/yorkie-team/yorkie/internal/log"
+	"github.com/yorkie-team/yorkie/pkg/document/time"
+)
+
+// Publish publishes the given event to the given Topic.
+func (s *Server) Publish(_ context.Context, request *api.PublishRequest) (*api.PublishResponse, error) {
```
hackerwins, line 13: Is there any reason to break the context chain?

yorkie/rpc/broadcast.go
```diff
@@ -0,0 +1,35 @@
...
+	actorID, err := time.ActorIDFromBytes(request.PublisherId)
+	if err != nil {
+		log.Logger.Fatal(err)
```
hackerwins, line 16: `Fatal` kills the server. Do you have any reason to kill me here?

hackerwins (changes requested):
Thanks for your contribution.
I left a few simple questions.

yorkie/rpc/server.go
```diff
@@ -23,6 +23,7 @@ import (
 	gotime "time"
 
 	grpcmiddleware "github.com/grpc-ecosystem/go-grpc-middleware"
+	grpc_recovery "github.com/grpc-ecosystem/go-grpc-middleware/recovery"
```
hackerwins, line 26:
```suggestion
	grpcrecovery "github.com/grpc-ecosystem/go-grpc-middleware/recovery"
```

hackerwins: It would be nice to put this when we release our agent as a product.
But for now, It would be good to explicitly handle the error.

test/stress/cluster_stress_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/yorkie-team/yorkie/pkg/document/proxy"
+	"github.com/yorkie-team/yorkie/pkg/types"
+	"github.com/yorkie-team/yorkie/test/helper"
+	"github.com/yorkie-team/yorkie/yorkie"
+	"github.com/yorkie-team/yorkie/yorkie/backend/sync"
+	"github.com/yorkie-team/yorkie/yorkie/backend/sync/memory"
+)
+
+func TestClusterStress(t *testing.T) {
+	t.Run("watch document test", func(t *testing.T) {
```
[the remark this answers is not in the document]
dc7303: This is a test written to ensure that documents are safely synchronized when multiple agents are activated and requests are randomly sent to each agent.

test/integration/cluster_mode_test.go
```diff
@@ -96,10 +96,10 @@ func TestClusterMode(t *testing.T) {
 		wg.Add(1)
 		rch := clientA.Watch(ctx, docA)
 		go func() {
-			defer wg.Done()
 
 			select {
 			case resp := <-rch:
+				defer wg.Done()
```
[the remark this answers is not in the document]
dc7303: This part is a mistake. I'll fix it.

test/stress/cluster_stress_test.go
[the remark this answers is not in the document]
hackerwins: Why not just be in the integration test?

yorkie/backend/sync/etcd/client.go
```diff
@@ -48,15 +49,22 @@ type Config struct {
...
 	config *Config
 	client *clientv3.Client
 
 	pubSub *memory.PubSub
 
-	memberMapMu *gosync.RWMutex
-	memberMap   map[string]*sync.AgentInfo
+	memberMapMu        *gosync.RWMutex
+	memberMap          map[string]*sync.AgentInfo
+	broadcastClientMap map[string]*broadcastClientInfo
```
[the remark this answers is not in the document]
dc7303: When accessing `broadcastClientMap`, usually used with `memberMap` to lock through `memberMapMu`.
So, in the current implementation direction, I thought that there is no possibility of accessing them simultaneously.
Do you have any concerns?

yorkie/rpc/broadcast.go
dc7303: The Publish API simply publishes the requested event to the local clients, so I couldn't think of anything to use the context for right now.
Can you suggest a situation that I might consider?

yorkie/rpc/broadcast.go
dc7303: This seems to be wrong. I'll fix it.

test/stress/cluster_stress_test.go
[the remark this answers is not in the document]
dc7303: I wanted to see what happens when I forward a lot of requests to multiple servers. For example, I thought it would be good to check unexpected situations such as unexpected errors and performance issues. Do you think it is appropriate to move to integration testing?
Come to think of it, it seems that this type of test is also necessary for integration tests.

yorkie/rpc/server.go
dc7303: That's right. So what do you think about adding an option to grpc config deciding whether to add grpc recovery interceptor or not?

test/stress/cluster_stress_test.go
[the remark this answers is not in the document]
hackerwins: I think this test is complicated. Let's start with the integration test by simplifying the test.

yorkie/backend/sync/etcd/client.go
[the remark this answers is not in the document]
hackerwins: I think this assumption is easy to make a mistake.
For example, the code below is inserting a new value into the map with readlock.

https://github.com/yorkie-team/yorkie/pull/189/files#diff-12847ffa877b49cdead390469adee37ad5b1bf871f09d554bda4403391a86443R60

yorkie/rpc/server.go
hackerwins: How about removing it from this PR and handling it as a separate issue?

yorkie/rpc/broadcast.go
hackerwins: I thought it was related to the routine below, but it's not.

Please check whether the context of the routine below can be linked with the callers.
https://github.com/yorkie-team/yorkie/pull/189/files#diff-12847ffa877b49cdead390469adee37ad5b1bf871f09d554bda4403391a86443R113

api/yorkie.proto
```diff
@@ -353,3 +353,23 @@ enum EventType {
     DOCUMENTS_WATCHED = 1;
     DOCUMENTS_UNWATCHED = 2;
 }
+
+service Broadcast {
+    rpc Publish (PublishRequest) returns (PublishResponse) {}
+}
```
hackerwins, line 359:
```suggestion
service Cluster {
    rpc BroadcastEvent (BroadcastEventRequest) returns (BroadcastEventResponse) {}
}
```

yorkie/backend/sync/etcd/pubsub.go
```diff
@@ -41,11 +48,73 @@ func (c *Client) Publish(
...
+}
+
+// publishToMember publishes events to other agents.
+func (c *Client) publishToMember(
+	memberAddr string,
+	publisherID *time.ActorID,
+	topic string,
+	event sync.DocEvent,
+) {
+	bcm := c.broadcastClientMap[memberAddr]
```
hackerwins, line 105: What does `bcm` stand for?

yorkie/backend/sync/etcd/pubsub.go
```diff
@@ -41,11 +48,73 @@ func (c *Client) Publish(
...
+
+	docEvent, err := converter.ToDocEvent(event)
+	if err != nil {
+		log.Logger.Error(err)
+		return
```
hackerwins, line 110: Is there any reason we broke the error chain here?

api/yorkie.proto
```diff
@@ -353,3 +353,23 @@ enum EventType {
...
+
+message PublishResponse {
+    string topic = 1;
+}
+
+message DocEvent {
+    EventType event_type = 1;
+	string doc_key = 2;
+	Client publisher = 3;
+}
```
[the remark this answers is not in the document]
dc7303: I thought that `message Event` has a different structure from `DocEvent` and has a different role.
I don't quite understand what you mean by integrating. Can you explain a little bit more?

yorkie/backend/sync/etcd/pubsub.go
dc7303: It is a variable representing broadcastClientMap, but it seems to have been completely misnamed.
The name of the broadcastClientMap property has been changed, so the variable name has also been modified.

yorkie/backend/sync/etcd/pubsub.go
dc7303: I didn't think it was necessary to return an error from the API where a particular Document event occurred just because it couldn't broadcast to another agent.
So I left an error log at the time of occurrence. This is because we thought that document sync could be synced through other sync events.

I have been thinking about how to alert when a problem that can occur within the method (such as a convert failure or an abnormal problem in the target agent) occurs continuously, but I was not sure, so I tried to suggest it later.

Can you please let me know if there are more things I should consider?

api/yorkie.proto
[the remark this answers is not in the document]
hackerwins: Both `Event` and `DocEvent` mean a document event. Previously, Event was used between Agent and Client, and I would like to use it for Agent Event Publish as well.

yorkie/backend/sync/etcd/pubsub.go
hackerwins: When we try to convert an event type that is not supported, an error occurs, and there is no need to loop further in the caller. I wish we could handle errors explicitly.

yorkie/backend/sync/etcd/pubsub.go
```diff
@@ -41,11 +48,78 @@ func (c *Client) Publish(
...
+
+// addClusterClient activates the cluster grpc server and creates a client.
+func (c *Client) addClusterClient(addr string) {
+	c.clusterClinetMapMu.Lock()
+	defer c.clusterClinetMapMu.Unlock()
+
+	if _, ok := c.clusterClientMap[addr]; !ok {
+		conn, err := grpc.Dial(addr, grpc.WithInsecure())
+		if err != nil {
+			log.Logger.Error(err)
```
hackerwins, line 91: Wouldn't it be better to return an error or try again?

yorkie/rpc/server.go
```diff
@@ -393,6 +394,7 @@ func (s *Server) WatchDocuments(
 		return err
 	}
 
+	// TODO(dc7303): broadcast
```
hackerwins, line 397: What does this mean?

yorkie/rpc/server.go
```diff
@@ -93,6 +93,7 @@ func NewServer(conf *Config, be *backend.Backend) (*Server, error) {
 		backend:    be,
 	}
 	api.RegisterYorkieServer(rpcServer.grpcServer, rpcServer)
+	api.RegisterClusterServer(rpcServer.grpcServer, rpcServer)
```
hackerwins, line 96: How about creating a new type like other servers?

yorkie/backend/sync/etcd/client.go
```diff
@@ -48,15 +49,23 @@ type Config struct {
...
 	config *Config
 	client *clientv3.Client
 
 	pubSub *memory.PubSub
 
-	memberMapMu *gosync.RWMutex
-	memberMap   map[string]*sync.AgentInfo
+	memberMapMu        *gosync.RWMutex
+	memberMap          map[string]*sync.AgentInfo
+	clusterClinetMapMu *gosync.RWMutex
```
hackerwins, line 67:
```suggestion
	clusterClientMapMu *gosync.RWMutex
```

yorkie/backend/sync/etcd/pubsub.go
```diff
@@ -41,11 +48,78 @@ func (c *Client) Publish(
 	publisherID *time.ActorID,
 	topic string,
 	event sync.DocEvent,
+) {
+	c.PublishToLocal(publisherID, topic, event)
+
+	c.memberMapMu.RLock()
+	defer c.memberMapMu.RUnlock()
```
hackerwins, line 55: Why do we need this lock?

yorkie/rpc/server.go
dc7303: Sorry. It seems to have been written roughly as a task to be carried out in this PR.
I plan to add an API to notify other Agents of initialization information when WatchDocument is called.

hackerwins (approved):
Thank you for your work. In particular, changing the topic to multi-key based in PubSub seems to be neat.

Broadcast seems to have many weaknesses compared to Gossip-based algorithms, but first of all, it seems to be able to operate a small-scale cluster.

(I made a few changes during the review.)

PR #201, opened 2021-06-12
 Improve error handling using pkg/errors

### LINKED ISSUE #199: Questions and suggestions for handling error logs
Question:
Are there any standards currently suggesting when to print (or resolve) the error log?

The reason I am asking this question is that when I check the current error handling code, there is a possibility that some mistakes can occur, and I thought that it might also cause confusion for developers who are starting Yorkie.

Let's take a few examples.
1. Duplicate log output
example:
[locker](https://github.com/yorkie-team/yorkie/blob/main/yorkie/backend/sync/etcd/locker.go#L32-L53) and [When to use lockers](https://github.com/yorkie-team/yorkie/blob/main/yorkie/rpc/server.go#L329-L336)
[watchDocs()](https://github.com/yorkie-team/yorkie/blob/main/yorkie/rpc/server.go#L471-L474) and [When to use watchDocs()](https://github.com/yorkie-team/yorkie/blob/0961a697448886a1383c2beb51cb46c84515fe14/yorkie/rpc/server.go#L387-L394)

2. Errors that are not output to the log.
example:
The [ObjectToBytes](https://github.com/yorkie-team/yorkie/blob/0961a697448886a1383c2beb51cb46c84515fe14/api/converter/to_bytes.go#L31-L43) function does not log output to the `to JSONElement(obj)` method, so if the caller does not output the error, the corresponding error is not logged.
Also, if the caller leaves a log, the error log for `proto.Marshal(pbElem)` may be duplicated.

3. The hassle of checking whether the log is output along the call stack to determine when to output the log.
For example, in the case of [PushPull()](https://github.com/yorkie-team/yorkie/blob/0961a697448886a1383c2beb51cb46c84515fe14/yorkie/packs/pack_service.go#L45-L150), a code that outputs an error and a code that does not return an error are mixed.

I think would be nice to document how the error log is handled to prevent mistakes like this.
If there is no standard, I would like to suggest 'Only handle errors once' section of ['Don't just check errors, handle them gracefully' aticle](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully).

Suggestion:
How about making the error log output when the error chain is broken (the first caller, the time of panic, etc.) by referring to the attached article?

dc7303:
In the case of zap currently in use, the place where the log is output is shown as the error occurrence point in the stack
trace. Introduced pkg/errors to allow log output to be processed at the end of the logic.

Fixes #199

Changed the code that wraps using `fmt.Errorf()` to use `errors.Wrap()`. If you have a better idea than this output format, please share.

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -18,9 +18,10 @@ package client
 
 import (
 	"context"
-	"errors"
+	goerrors "errors"
 
 	"github.com/google/uuid"
+	"github.com/pkg/errors"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 
@@ -43,11 +44,11 @@ const (
 var (
 	// ErrClientNotActivated occurs when an inactive client executes a function
 	// that can only be executed when activated.
-	ErrClientNotActivated = errors.New("client is not activated")
+	ErrClientNotActivated = goerrors.New("client is not activated")
 
 	// ErrDocumentNotAttached occurs when the given document is not attached to
 	// this client.
-	ErrDocumentNotAttached = errors.New("document is not attached")
+	ErrDocumentNotAttached = goerrors.New("document is not attached")
 )
 
 // Client is a normal client that can communicate with the agent.
@@ -105,8 +106,7 @@ func NewClient(opts ...Option) (*Client, error) {
 			serverNameOverride,
 		)
 		if err != nil {
-			log.Logger.Error(err)
-			return nil, err
+			return nil, errors.WithStack(err)
 		}
 		dialOptions = append(dialOptions, grpc.WithTransportCredentials(creds))
 	} else {
@@ -150,8 +150,7 @@ func Dial(rpcAddr string, opts ...Option) (*Client, error) {
 func (c *Client) Dial(rpcAddr string) error {
 	conn, err := grpc.Dial(rpcAddr, c.dialOptions...)
 	if err != nil {
-		log.Logger.Error(err)
-		return err
+		return errors.WithStack(err)
 	}
 
 	c.conn = conn
@@ -163,6 +162,7 @@ func (c *Client) Dial(rpcAddr string) error {
 // Close closes all resources of this client.
 func (c *Client) Close() error {
 	if err := c.Deactivate(context.Background()); err != nil {
+		log.Logger.Errorf("%+v", err)
 		return err
 	}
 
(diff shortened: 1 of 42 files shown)


hackerwins (commented):
Thanks for your contribution.

Let me ask you a few questions.

1. Please explain the role of `errors.Wrapf`, `errors.WithStack`, and `goerrors.New` and where it should be called.
2. Where should the call to `log.Logger.Errorf` be?

dc7303:
Thanks for leaving a comment.

### About error
`goerrors.New` is used when creating predefined errors. At this point, I tried to use New() of `pkg/errors`, but it was not used because a stack trace didn't indicate the correct error source.

`errors.Wrapf` and `errros.WithStack` were used where the error occurred.  The part where these functions are used appears as the point where the error occurred in the stack trace.  (Previously, we called the error log output function at the point where error occurred.)

`errors.Wrapf` replaces `fmt.Errorf` which wraps the error that occurred.  And `errors.WithStack` is used to return an error without wrapping the error that occurred.

### About log
It would be good to call `log.Logger.Error` from the package that the process starts. For example, it could be a Controller in the MVC model. This way we can be responsible for the error output in one place.

dc7303:
There is a caveat when we use `pkg/errors`. Errors returned using `pkg/errors` should not be wrapped again in `pkg/errors` or using `WithStack`.

This is because it can cause duplicate call stacks. More details are attached at the link below.
https://github.com/pkg/errors/issues/242

client/client.go
```diff
@@ -43,11 +44,11 @@ const (
 var (
 	// ErrClientNotActivated occurs when an inactive client executes a function
 	// that can only be executed when activated.
-	ErrClientNotActivated = errors.New("client is not activated")
+	ErrClientNotActivated = goerrors.New("client is not activated")
 
 	// ErrDocumentNotAttached occurs when the given document is not attached to
 	// this client.
-	ErrDocumentNotAttached = errors.New("document is not attached")
+	ErrDocumentNotAttached = goerrors.New("document is not attached")
```
alucic, line 51: 👋  I came here from https://github.com/pkg/errors/issues/242

I tend to use `fmt.Errorf` for declaring error variables so the only `errors` package in the codebase is `github.com/pkg/errors`. This way I don't have to think about which `errors` package I'm using, it is never std errors.

dc7303:
@hackerwins I think it would be better to close this PR first and proceed again.
I'm sorry I'm requesting a review of too many changes at once. I'd like to ask for a review of small units when I start again. What do you think? It would be good to continue the discussion then.

hackerwins:
@dc7303 I agree.
I think it would be better to start with a small scope while defining the problems together.

PR #207, opened 2021-07-05
Add Attachment type to pass the peers' status

### LINKED ISSUE #137: The events exposed when subscribing a client are different from `yorkie-js-sdk`
@habibrosyad informed us that the events exposed in [#108 PR discussion](https://github.com/yorkie-team/yorkie/pull/108#issuecomment-745925842) are different. 👍 
I tested the actual `Watch()` event, you can see that the contents are different.

```bash
# js:
Object{name: 'documents-watching-peer-changed', value: Object{test-col$Can watch documents-1609554095519: [...]}}

# go:
{EventType:documents-watched Keys:[0xc0002a45e0] Err:<nil>}
```

It might be a good idea to solve this problem before implementing the `peer awareness` feature.

hackerwins:

Add `Attachment` type to pass the peers' status and provide `WatchResponse` similar to [that of the JS SDK](https://github.com/yorkie-team/yorkie-js-sdk/blob/main/src/core/client.ts#L63-L64)(`document-changed`, `peers-changed`).

Fixes #137

```go
wrch := c1.Watch(watch1Ctx, d1)
wr := <-wrch:
if wr.Type == client.PeersChanged {
	// peers changed
	wr.PeersMapByDoc[doc.Key().BSONKey()]
} else if wr.Type == DocumentsChanged {
	// document changed
}
```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/api/converter/from_pb.go
+++ b/api/converter/from_pb.go
@@ -159,6 +159,20 @@ func FromDocEvent(docEvent *api.DocEvent) (*sync.DocEvent, error) {
 	}, nil
 }
 
+// FromClients converts the given Protobuf format to model format.
+func FromClients(pbClients *api.Clients) ([]*types.Client, error) {
+	var clients []*types.Client
+	for _, pbClient := range pbClients.Clients {
+		client, err := FromClient(pbClient)
+		if err != nil {
+			return nil, err
+		}
+		clients = append(clients, client)
+	}
+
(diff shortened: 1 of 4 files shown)


dc7303 (approved):
LGTM 👍

PR #212, opened 2021-07-11
Add authorization webhook retry logic

### LINKED ISSUE #194: Implement retry and retryBackoff in authorization webhook

- [ ] Add retry in authorization webhook
- [ ] Add retryBackoff(exponential time in ms to wait between retries, https://github.com/kubernetes/kubernetes/blob/fcf9300cc8e6918fce26e9ad1e9f7360a7cbf552/staging/src/k8s.io/apiserver/pkg/util/webhook/webhook.go#L123)

**Why is this needed**:

Authorizers can also be unstable, so we should retry it in a safe way.

dc7303:
When the webhook is executed, it should be able to safely retry the request based on the webhook server response.

Fixes #194 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/test/integration/auth_webhook_test.go
+++ b/test/integration/auth_webhook_test.go
@@ -148,7 +148,7 @@ func TestAuthWebhook(t *testing.T) {
 
 		conf := helper.TestConfig(server.URL)
 		conf.Backend.AuthorizationWebhookMaxRetries = recoveryCnt
-		conf.Backend.AuthorizationWebhookMaxWaitInterval = 1000
+		conf.Backend.AuthorizationWebhookMaxWaitIntervalMillis = 1000
 		agent, err := yorkie.New(conf)
 		assert.NoError(t, err)
 		assert.NoError(t, agent.Start())
@@ -174,7 +174,7 @@ func TestAuthWebhook(t *testing.T) {
 
 		conf := helper.TestConfig(server.URL)
 		conf.Backend.AuthorizationWebhookMaxRetries = 2
-		conf.Backend.AuthorizationWebhookMaxWaitInterval = 1000
+		conf.Backend.AuthorizationWebhookMaxWaitIntervalMillis = 1000
 		agent, err := yorkie.New(conf)
 		assert.NoError(t, err)
 		assert.NoError(t, agent.Start())
--- a/test/integration/auth_webhook_test.go
+++ b/test/integration/auth_webhook_test.go
@@ -20,7 +20,6 @@ package integration
 
 import (
 	"context"
-	"encoding/json"
 	"net/http"
 	"net/http/httptest"
 	"sync"
@@ -42,8 +41,8 @@ func newAuthServer(t *testing.T) (*httptest.Server, string) {
 	token := xid.New().String()
 
 	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		var req types.AuthWebhookRequest
-		assert.NoError(t, json.NewDecoder(r.Body).Decode(&req))
+		req, err := types.NewAuthWebhookRequest(r.Body)
+		assert.NoError(t, err)
 
 		var res types.AuthWebhookResponse
 		if req.Token == token {
@@ -51,34 +50,30 @@ func newAuthServer(t *testing.T) (*httptest.Server, string) {
 		} else {
 			res.Reason = "invalid token"
 		}
-		bytes, err := json.Marshal(res)
-		_, err = w.Write(bytes)
+
+		_, err = res.Write(w)
 		assert.NoError(t, err)
 	})), token
 }
 
-func newUnavailableAuthServer(t *testing.T, recoveryCnt uint64) (*httptest.Server, string) {
-	token := xid.New().String()
-
+func newUnavailableAuthServer(t *testing.T, recoveryCnt uint64) *httptest.Server {
 	var retries uint64
 	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		var req types.AuthWebhookRequest
-		assert.NoError(t, json.NewDecoder(r.Body).Decode(&req))
+		_, err := types.NewAuthWebhookRequest(r.Body)
+		assert.NoError(t, err)
 
 		var res types.AuthWebhookResponse
 		res.Allowed = true
-		bytes, err := json.Marshal(res)
-
 		if retries < recoveryCnt-1 {
 			w.WriteHeader(http.StatusServiceUnavailable)
 			retries++
 		} else {
 			retries = 0
 		}
-		_, err = w.Write(bytes)
-		assert.NoError(t, err)
 
-	})), token
+		_, err = res.Write(w)
+		assert.NoError(t, err)
+	}))
 }
 
 func TestAuthWebhook(t *testing.T) {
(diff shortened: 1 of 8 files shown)


dc7303:
First, I implemented the algorithm required to retry the webhook, and I ask for your review.

There are a few additional things you might want to consider together.
1. Timeout seems to be required.
-> Retrying the webhook can sometimes cause severe delays in response. How about being able to set the timeout of the API with the Agent option?

2. About the maximum number of retries and the maximum wait interval options.
-> It would be nice if the user could customize it according to the user's webhook state. What do you think?
The current default setting is both set to 0. If you have any better ideas please tell me.

3. How about adding a library for testing so that you can test using monkey patch?
 -> I wanted to validate the details of `VerifyAccess` in `yorkie/auth/webhook.go`, so I tried to write a unit test. However, since I had to depend on the `http` library, we added the test only to the integration test.

test/integration/auth_webhook_test.go
```diff
@@ -57,6 +57,30 @@ func newAuthServer(t *testing.T) (*httptest.Server, string) {
 	})), token
 }
 
+func newUnavailableAuthServer(t *testing.T, recoveryCnt uint64) (*httptest.Server, string) {
+	token := xid.New().String()
+
+	var retries uint64
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var req types.AuthWebhookRequest
+		assert.NoError(t, json.NewDecoder(r.Body).Decode(&req))
```
hackerwins, line 66:
```suggestion
		req, err := types.NewAuthWebhookRequest(r.Body)
		assert.NoError(t, err)
```

test/integration/auth_webhook_test.go
```diff
@@ -57,6 +57,30 @@ func newAuthServer(t *testing.T) (*httptest.Server, string) {
...
+
+		if retries < recoveryCnt-1 {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			retries++
+		} else {
+			retries = 0
+		}
+		_, err = w.Write(bytes)
+		assert.NoError(t, err)
+
```
hackerwins, line 80:
```suggestion
		var res types.AuthWebhookResponse
		res.Allowed = true

		if retries < recoveryCnt-1 {
			w.WriteHeader(http.StatusServiceUnavailable)
			retries++
		} else {
			retries = 0
		}
		
		_, err = res.Write(w)
		assert.NoError(t, err)
```
```

File: yorkie/auth/webhook.go [hunk 3]
```diff
@@ -49,29 +54,70 @@ func VerifyAccess(ctx context.Context, be *backend.Backend, info *types.AccessIn
...
+			return err
+		}
+
+		time.Sleep(waitInterval(retries, config.AuthorizationWebhookMaxWaitInterval))
+
+		statusCode, err := webhookFn()
+		if !shouldRetry(statusCode) {
+			return err
+		}
+		retries++
```
hackerwins, line 108: 1. It would be nice to be able to cancel with the context passed from the caller.
2. I think It would be good to explicitly return ErrWaitTimeout.

```suggestion
	var retries uint64
	for retries < config.AuthorizationWebhookMaxRetries {
		waitBeforeRetry := waitInterval(retries, config.AuthorizationWebhookMaxWaitIntervalSec)
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-time.After(waitBeforeRetry):
		}

		statusCode, err := webhookFn()
		if !shouldRetry(statusCode) {
			return err
		}
		retries++
	}
	
	return ErrWaitTimeout
```

yorkie/auth/webhook.go
```diff
@@ -49,29 +54,70 @@ func VerifyAccess(ctx context.Context, be *backend.Backend, info *types.AccessIn
...
 	}
+}
+
+func waitInterval(retries uint64, maxWaitInterval uint64) time.Duration {
```
hackerwins, line 112: It would be nice to express the unit of interval.

```suggestion
func waitInterval(retries uint64, maxWaitInterval time.Duration) time.Duration {
```

yorkie/auth/webhook.go
```diff
@@ -49,29 +54,70 @@ func VerifyAccess(ctx context.Context, be *backend.Backend, info *types.AccessIn
...
+	interval = math.Min(interval, float64(maxWaitInterval))
+	return time.Duration(int64(interval)) * time.Millisecond
+}
 
-	return nil
+func shouldRetry(statusCode int) bool {
+	return statusCode == http.StatusRequestTimeout ||
+		statusCode == http.StatusInternalServerError ||
+		statusCode == http.StatusBadGateway ||
+		statusCode == http.StatusServiceUnavailable
```
hackerwins, line 122: Is there any reason to check these conditions? If there is no particular reason, why not start with the status below?

```suggestion
	return statusCode == http.StatusInternalServerError ||
		statusCode == http.StatusGatewayTimeout ||
		statusCode == http.StatusTooManyRequests
```

https://github.com/kubernetes/kubernetes/blob/fcf9300cc8e6918fce26e9ad1e9f7360a7cbf552/staging/src/k8s.io/apiserver/pkg/util/webhook/webhook.go#L63-L66

hackerwins (changes requested):
Thanks for your contribution.
I left a few comments.

yorkie/backend/backend.go
```diff
@@ -48,6 +48,14 @@ type Config struct {
...
 	// AuthorizationWebhookMethods is the methods that run the authorization webhook.
 	AuthorizationWebhookMethods []string `json:"AuthorizationWebhookMethods"`
+
+	// AuthorizationWebhookMaxRetries is the max count
+	// that retries the authorization webhook.
+	AuthorizationWebhookMaxRetries uint64
+
+	// AuthorizationWebhookMaxWaitInterval is the max interval
+	// that waits before retrying the authorization webhook.
+	AuthorizationWebhookMaxWaitInterval uint64
```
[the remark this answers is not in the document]
dc7303: Thanks for commenting :)
While doing this, how about adding a command option so that it can be input as well?

test/integration/auth_webhook_test.go
```diff
@@ -57,6 +57,30 @@ func newAuthServer(t *testing.T) (*httptest.Server, string) {
 	})), token
 }
 
+func newUnavailableAuthServer(t *testing.T, recoveryCnt uint64) (*httptest.Server, string) {
+	token := xid.New().String()
```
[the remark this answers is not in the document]
dc7303: This is the part I didn't pay attention to. There is already a test to check whether or not it is allowed, so I think to delete it!

yorkie/auth/webhook.go
dc7303: 1. I agree. I think what you said is related to the opinion you gave in point 1 of my discussion. What do you think about adding the Timeout option of the API to the Agent?

2. I thought it would be easier for users to figure out the problem by returning the last state of the webhook rather than a Timeout error after the last retry. Could you explain a bit more about the benefit of returning a timeout error on failure after the last retry?

yorkie/auth/webhook.go
dc7303: @hackerwins Modifications to `auth_webhook_test.go` have been reflected. And I left a few more comments before proceeding with the rest of the modifications. Please check :)

dc7303:
@hackerwins Thanks so much for your feedback! Modifications to auth_webhook_test.go have been reflected. And I left a few more comments before proceeding with the rest of the modifications. Please check :)

hackerwins (commented):
Thanks for the correction.
I'm more curious about some intentions of the questions. It would be good to first modify codes and then verbally discuss the specification you asked for.

hackerwins (approved):
LGTM. 👍

Issue #168, opened 2021-03-30
Change `XXXTimeoutSec` flags in CLI to duration


hackerwins:
**What would you like to be added**:

Change `XXXTimeoutSec` flags in CLI to duration so that users can input the flags in `5s` or `5m`.

https://github.com/yorkie-team/yorkie/blob/120075576a431de3382740ec1844948ff16177bb/pkg/cli/agent.go#L50-L51

```
# AS-IS
yorkie agent --mongo-connection-timeout-sec 60

# TO-BE
yorkie agent --mongo-connection-timeout 1m
```

We can easily convert it using [ParseDuration](https://golang.org/pkg/time/#ParseDuration).

**Why is this needed**:
- Users can set durations in flags more conveniently.

computerphilosopher:
If we set flags to [DurationVar](https://golang.org/src/flag/flag.go?s=27461:27543#L809), we can use duration string as a flags' value.

But I don't use it in #152 because flags' names are 'second', not 'duration'.

hackerwins:
@computerphilosopher Yes of course. We have to change it.

dc7303:
I was distracted the other day and couldn't take care of it.
I will proceed!

PR #216, opened 2021-07-18
Implement authorization webhook cache

### LINKED ISSUE #192: Implement authorization webhook cache

- [ ] Implement authorization webhook cache([LRUExpireCache](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/cache/lruexpirecache.go)).
- [ ] Add `authorization-webhook-cache-authorized-ttl` flag
- [ ] Add `authorization-webhook-cache-unauthorized-ttl` flag

**Why is this needed**:

We can reduce the load caused by the authorization webhooks that affect response time.

dc7303:
We can reduce the load caused by the authorization webhooks that affect response time.

Fixes #192 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


ppeeou (approved):
LOL LGTM👍

yorkie/backend/backend.go
```diff
@@ -155,13 +162,19 @@ func New(
 		agentInfo.RPCAddr,
 	)
 
+	cache, err := NewLRUExpireCache(5000)
```
hackerwins, line 165: It would be better to extract it as a constant.

yorkie/backend/cache.go
```diff
@@ -0,0 +1,110 @@
...
+
+package backend
+
+import (
+	"container/list"
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/yorkie-team/yorkie/pkg/types"
```
hackerwins, line 28: How about removing the AuthWebhookResponse dependency and moving `cache.go` to `pkg/cache`?

hackerwins (changes requested):
Thanks for the work I left a few comments.

test/integration/auth_webhook_test.go
```diff
@@ -188,4 +190,138 @@ func TestAuthWebhook(t *testing.T) {
...
+		go func() {
+			for {
+				select {
+				case <-wrch:
+					continue
+				case <-time.After(5 * time.Second):
+					return
+				}
+			}
+		}()
```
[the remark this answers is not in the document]
dc7303: I was going to do something.. in the process of erasing it, this is left.. my mistake.. 😭 

Something I was going to do was out of scope for this test, so I deleted it.
Thanks to you, I was able to improve on unnecessary elements in this test code. Thank you! 👍

yorkie/backend/cache.go
dc7303: This decision was made because it is only used in AuthWebhook. But I agree with you because this code is likely to be used for utility. It's corrected!

test/integration/auth_webhook_test.go
```diff
@@ -188,4 +190,138 @@ func TestAuthWebhook(t *testing.T) {
...
+		assert.NoError(t, err)
+
+		// request(PushPull) 3
+		for i := 0; i < 3; i++ {
+			doc2.Update(func(root *proxy.ObjectProxy) error {
+				root.SetNewObject("k1")
+				return nil
+			})
+		}
+		cli2.Sync(ctx2)
```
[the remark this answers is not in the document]
dc7303: Oh! Fixed to request multiple times. Thank you.

yorkie/backend/backend.go
```diff
@@ -34,6 +35,8 @@ import (
 	"github.com/yorkie-team/yorkie/yorkie/metrics"
 )
 
+const lruExpireCacheSize = 5000
```
hackerwins, line 38: I think the readability will improve if we express where the cache is used.
```suggestion
const authWebhookCacheSize = 5000
```

pkg/cache/cache_test.go
```diff
@@ -0,0 +1,55 @@
...
+		lruCache, err = cache.NewLRUExpireCache(0)
+		assert.Error(t, err)
+		assert.Nil(t, lruCache)
+	})
+
+	t.Run("add test", func(t *testing.T) {
+		lruCache, err := cache.NewLRUExpireCache(1)
+		assert.NoError(t, err)
+
+		lruCache.Add("request1", "response1", time.Second*5)
```
hackerwins, line 27: Wouldn't it be more readable to write the unit after the value?
```suggestion
		lruCache.Add("request1", "response1", 5*time.Second)
```

test/integration/auth_webhook_test.go
```diff
@@ -188,4 +190,111 @@ func TestAuthWebhook(t *testing.T) {
...
+
+		for i := 0; i < 3; i++ {
+			doc.Update(func(root *proxy.ObjectProxy) error {
+				root.SetNewObject("k1")
+				return nil
+			})
+			cli.Sync(ctx)
+		}
+
+		assert.Equal(t, reqCnt, 2)
```
hackerwins, line 251: This test seems to contain constants related to the value of authorizedTTL. How about expressing the value of this constant based on authorizedTTL(e.g. `3` -> `authorizedTTL + 1`)?

hackerwins (changes requested):
Thanks for the correction. 👍 

How about changing `AuthorizationWebhookXXX` to `AuthWebhookXXX` in this PR?
In addition, I left a brief comment regarding readability.

dc7303:
@hackerwins Thanks for the additional feedback!
I was thinking of proceeding with issue #168 as the next task. And I was going to try to modify the `AuthorizationWebhookXXX` in the new PR. (expected PR title: Improve configs)
I think the correction part seems to overlap in #168 progress.
ex) `AuthWebhookCacheAuthorizedTTLSec` -> `AuthWebhookCacheAuthorizedTTL`
So, how about correcting the part you mentioned in my next PR after this PR is over?

test/integration/auth_webhook_test.go
dc7303: Thank you so much for your comments.
I didn't understand your intentions. Could you please explain a little more?

The reason for requesting requests 3 times was intended to be multiple requests.
I didn't think the constant 3 had anything to do with the authorizedTTL value.

test/integration/auth_webhook_test.go
```diff
@@ -188,4 +190,111 @@ func TestAuthWebhook(t *testing.T) {
...
+		assert.Equal(t, reqCnt, 1)
+
+		time.Sleep(time.Duration(unauthorizedTTL) * time.Second)
+
+		for i := 0; i < 3; i++ {
+			err = cli.Activate(ctx)
+			assert.Equal(t, codes.Unauthenticated, status.Convert(err).Code())
+		}
+		assert.Equal(t, codes.Unauthenticated, status.Convert(err).Code())
+		assert.Equal(t, reqCnt, 2)
```
hackerwins, line 298: 1. The error check code doesn't seem necessary.
2. We need to switch `exected` and `actual` in assert.Equal
```suggestion
		assert.Equal(t, 2, reqCnt)
```

test/integration/auth_webhook_test.go
hackerwins: 1. 2 seconds to 1 second to avoid increasing test time
2. We need to switch exected and actual in assert.Equal
3. More detailed explanation in the comments

```suggestion
		var authorizedTTLSec uint64 = 1
		conf := helper.TestConfig(server.URL)
		conf.Backend.AuthorizationWebhookCacheAuthorizedTTLSec = authorizedTTLSec

		agent, err := yorkie.New(conf)
		assert.NoError(t, err)
		assert.NoError(t, agent.Start())
		defer func() { assert.NoError(t, agent.Shutdown(true)) }()

		ctx := context.Background()
		cli, err := client.Dial(agent.RPCAddr(), client.Option{Token: "token"})
		assert.NoError(t, err)
		defer func() { assert.NoError(t, cli.Close()) }()

		err = cli.Activate(ctx)
		assert.NoError(t, err)

		doc := document.New(helper.Collection, t.Name())
		err = cli.Attach(ctx, doc)
		assert.NoError(t, err)

		// 01. multiple requests to update the document.
		for i := 0; i < 3; i++ {
			assert.NoError(t, doc.Update(func(root *proxy.ObjectProxy) error {
				root.SetNewObject("k1")
				return nil
			}))
			assert.NoError(t, cli.Sync(ctx))
		}
		assert.Equal(t, 1, reqCnt)

		// 02. multiple requests to update the document after eviction by ttl.
		time.Sleep(time.Duration(authorizedTTLSec) * time.Second)
		for i := 0; i < 3; i++ {
			doc.Update(func(root *proxy.ObjectProxy) error {
				root.SetNewObject("k1")
				return nil
			})
			cli.Sync(ctx)
		}
		assert.Equal(t, 2, reqCnt)
```

hackerwins (changes requested):
Thanks for the correction. 👍 
I left a few more comments.

test/integration/auth_webhook_test.go
dc7303: @hackerwins Thank you so much for the detailed explanation! It's corrected :)

hackerwins (approved):
LGTM. 👍

PR #217, opened 2021-07-20
Improve Client's metadata to be updatable

### LINKED ISSUE #153: Improve Peer Awareness's metadata to be updatable

hackerwins:

Improve Client's metadata to be updatable

Fixes #153

```
// TODO(hackerwins): We temporarily use Unary Call to update metadata,
// because grpc-web can't handle Bi-Directional streaming for now.
// After grpc-web supports bi-directional streaming, we can remove the
// Unary Call.
```

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


client/client.go
```diff
@@ -426,18 +410,22 @@ func (c *Client) Watch(
...
 					if err != nil {
 						return nil, err
 					}
 
 					attachment := c.attachments[k.BSONKey()]
-					if eventType == types.DocumentsWatchedEvent {
-						attachment.peerClients[cli.ID.String()] = cli.Metadata
+					if eventType == types.DocumentsWatchedEvent || eventType == types.MetadataChangedEvent {
+						if info, ok := attachment.peers[cli.ID.String()]; ok {
+							info.Update(cli.MetadataInfo)
```
dc7303, line 423: Currently, `attachment.peers` is of type `map[string]types.MetadataInfo`. `MetadataInfo` is not a pointer.
So even if you `Update()` the `info` variable, the instance managed inside `attachment.peers` is not updated. Is this intentional?

I think the WatchResponse is being returned as an unintended response value.
I think it would be nice to add a test that checks that the metadata of the Client instance and the peers information synced by other clients match.
Could tell me what you think?

dc7303 (changes requested):
That's great work! 👍 
I left a few comments. please confirm.

client/client.go
hackerwins: Thanks to you, we found a bug that would be hard to find later.

test/integration/cluster_mode_test.go
```diff
@@ -67,104 +111,92 @@ func TestClusterMode(t *testing.T) {
...
+
+						if wr.Type == client.PeersChanged {
+							peers := wr.PeersMapByDoc[d1.Key().BSONKey()]
+							responsePairs = append(responsePairs, watchResponsePair{
+								Type:  wr.Type,
+								peers: peers,
+							})
+						} else if wr.Type == client.DocumentsChanged {
+							assert.NoError(t, c1.Sync(ctx, wr.Keys...))
+						}
```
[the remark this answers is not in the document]
hackerwins: It seems that readability will decrease if we use switch here because it is nested with the select case statement.

dc7303 (approved):
LGTM 👍

ppeeou (approved):
LGTM

PR #221, opened 2021-07-27
Add setting index value to array

### LINKED ISSUE #215: Questions for providing value setting API based on Array index
**Description**:

Before implementing the value setting API based on Array index, I have a question to ask
related https://github.com/yorkie-team/yorkie-js-sdk/issues/191

Our strategy is to set value at the value position of `createdAt` in the array index.

However, if the above strategy is used in the current code, the following issue bellow

```go
// base array
d1.Update(func(root *proxy.ObjectProxy) error {
	root.SetNewArray("k1").AddInteger(0, 1, 2)
	return nil
})
c1.Sync(ctx)
c2.Sync(ctx)

d1.Update(func(root *proxy.ObjectProxy) error {
	root.GetArray("k1").AddInteger(0, 4)
	assert.Equal(t, `{"k1":[0,4,1,2]}`, root.Marshal())
	return nil
})

d2.Update(func(root *proxy.ObjectProxy) error {
	root.GetArray("k1").SetValue(1, 5)
	assert.Equal(t, `{"k1":[0,5,2]}`, root.Marshal())
	return nil
})

c1.Sync(ctx)
c2.Sync(ctx)
c1.Sync(ctx)
assert.Equal(t, `{"k1":[0,4,5,2]}`, d1.Marshal()) // expected value
assert.Equal(t, `{"k1":[0,5,4,2]}`, d2.Marshal()) // but 😭
```

`[0, 4, 5, 2]` are expected values. But the value of d2 is  `[ 0, 5, 4, 2 ]`.
The reason is that `InsertIntegerAfter` compares the index's next node `createdAt` and puts it after createdAt that is larger than itself.

https://github.com/yorkie-team/yorkie/blob/main/pkg/document/json/rga_tree_list.go#L306

array `[0(t0),1(t1),2(t2)]`

d1 case
1. In d1, `add operation` is applied first.
`d1.InsertIntegerAfter(0, 4)`
 -> `[0(t0),4(t3),1(t1),2(t2)]`
 2. `setIndex` is applied to create the desired value.
`d1.setIndex( 1, 5 )`
 -> `[0(t0),4(t3),5(t4),2(t2)]`

However
d2 case
1. In d2, `setIndex operation` is applied first.
`d2.setIndex( 1, 5 )`
-> `[0(t0),5(t4),2(t2)]`

2. `add operation` applies, but `createdAt` of add operation puts t3 after 5.
`d2.InsertIntegerAfter(0, 4)`
-> `[0(t0),5(t4),4(t3),2(t2)]`

In the present state, The commutative law does not hold.

**Why**:

ppeeou:

we can insert a value at a specific index into arrray
yorkie-sdk-js also needs work

Fixes #215

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


api/converter/from_pb.go
```diff
@@ -440,6 +442,31 @@ func fromIncrease(pbInc *api.Operation_Increase) (*operation.Increase, error) {
...
 
+func fromSetIndex(pbInc *api.Operation_SetIndex) (*operation.SetIndex, error) {
+	parentCreatedAt, err := fromTimeTicket(pbInc.ParentCreatedAt)
+	if err != nil {
+		return nil, err
+	}
+	positionAt, err := fromTimeTicket(pbInc.PositionAt)
+	if err != nil {
+		return nil, err
+	}
```
hackerwins, line 453: Since it is the creation time of the target to be changed, how about changing the variable name to `targetCreatedAt`?

pkg/document/json/rga_tree_list.go
```diff
@@ -358,3 +492,20 @@ func (a *RGATreeList) insertAfter(
 
 	a.size++
 }
+
+func (a *RGATreeList) insertNodeAfter(
```
hackerwins, line 496: It seems that insertNodeAfter has been newly added to solve the problem. Can you briefly describe the algorithm?

hackerwins (changes requested):
Thanks for the work. 👍 
I asked for a brief explanation of some changes and algorithms in the comments.

pkg/document/json/rga_tree_list.go
ppeeou: Thank you for your comment!

Before `setByIndex` existed, only one element could be added when inserted into an array.
image



So, when `setByIndex` is added and an operation such as movement occurs, the node of the index (with multiple elements) is moved, so the above method was added.
array 

image

ppeeou:
@hackerwins 
Is it correct to proceed in this direction🤔?? I want to check `yorkie-sdk-js` before working

api/converter/from_pb.go
```diff
@@ -440,6 +442,31 @@ func fromIncrease(pbInc *api.Operation_Increase) (*operation.Increase, error) {
 	), nil
 }
 
+func fromSetIndex(pbInc *api.Operation_SetIndex) (*operation.SetIndex, error) {
```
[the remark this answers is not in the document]
ppeeou: Oh my mistake  It's corrected!

api/converter/from_pb.go
```diff
@@ -198,6 +198,8 @@ func FromOperations(pbOps []*api.Operation) ([]operation.Operation, error) {
 			op, err = fromStyle(decoded.Style)
 		case *api.Operation_Increase_:
 			op, err = fromIncrease(decoded.Increase)
+		case *api.Operation_SetIndex_:
```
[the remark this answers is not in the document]
ppeeou: It's fixed https://github.com/yorkie-team/yorkie/pull/221/commits/ffa5a1f1210f31f28e465332a4df692f653ecaaf

test/integration/array_test.go
```diff
@@ -199,4 +200,202 @@ func TestArray(t *testing.T) {
...
+
+		assert.NoError(t, d2.Update(func(root *proxy.ObjectProxy) error {
+			root.GetArray("k1").SetInteger(0, 4)
+			assert.Equal(t, `{"k1":[4,1,2]}`, root.Marshal())
+			return nil
+		}))
+		c1.Sync(ctx)
+		c2.Sync(ctx)
+		c1.Sync(ctx)
+		syncClientsThenAssertEqual(t, []clientAndDocPair{{c1, d1}, {c2, d2}})
```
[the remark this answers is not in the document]
ppeeou: It's fixed https://github.com/yorkie-team/yorkie/pull/221/commits/2edf27c67d5df0e01bf12b9f24758cad569fbf73

api/converter/from_pb.go
ppeeou: It's corrected https://github.com/yorkie-team/yorkie/pull/221/commits/70e81a4bee1953c01014b8f110617cd6d7cd0bf6

hackerwins (changes requested):
This PR changed the internal behavior and might be hard to find later, so take a little more time to look at it.

Makefile
```diff
@@ -32,7 +32,7 @@ fmt:
 	gofmt -s -w $(GO_SRC)
 
 lint:
-	 golangci-lint run ./...
+	golangci-lint run ./...
```
loganstone, line 35: It would be good to separate other commit. It doesn't seem intent of commit.

hackerwins:
The users who requested this feature are not currently seeking it. Therefore, I'll close this PR for now and will reopen it when new requests for the feature arise.

PR #222, opened 2021-08-07
Add RPC Config validation and handle the errors

### LINKED ISSUE #206: Validate the entire setup configuration
https://github.com/yorkie-team/yorkie/blob/6eccd481eb74fc90b0d041a56280c129d43e2675/internal/cli/agent.go#L69-L71

**What would you like to be added:**
Option values ​​that need validation should be fully checked. After the check, if there is a problem, the agent can stop the execution and ask the user to fix it.

**Why is this needed:**
We need to validate the options entered by the user to prevent them from being executed incorrectly.

umi0410:

The following can be check if they are validated.
e.g. wrong ports, negative interval values, wrong format of mongo URI.
If the wrong config terminates the process and let users know why the config is wrong, users can config yorkie a lot easier.

Address #206 

I just wrote some draft codes. If it is not a bad approach, I want to be assigned for this issue.

```release-note
No.
```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/internal/cli/agent.go
+++ b/internal/cli/agent.go
@@ -67,7 +67,10 @@ func newAgentCmd() *cobra.Command {
 			}
 
 			// TODO(hackerwins): We need to check that the entire config is valid.
-			// err = conf.Validate()
+			if err := conf.Validate(); err != nil {
+				return err
+			}
+
 			err := conf.Backend.Validate()
 			if err != nil {
 				return err
--- a/internal/cli/agent.go
+++ b/internal/cli/agent.go
@@ -66,7 +66,6 @@ func newAgentCmd() *cobra.Command {
 				conf = parsed
 			}
 
-			// TODO(hackerwins): We need to check that the entire config is valid.
 			if err := conf.Validate(); err != nil {
 				return err
 			}
--- a/internal/cli/agent.go
+++ b/internal/cli/agent.go
@@ -71,11 +71,6 @@ func newAgentCmd() *cobra.Command {
 				return err
 			}
 
-			err := conf.Backend.Validate()
-			if err != nil {
-				return err
-			}
-
 			r, err := yorkie.New(conf)
 			if err != nil {
 				return err
--- a/yorkie/config.go
+++ b/yorkie/config.go
@@ -110,7 +110,7 @@ func newConfig(port int, metricsPort int, dbName string) *Config {
 // Validate returns an error if the provided Config is invalidated.
 func (c *Config) Validate() error {
 	var err error
-	
+
 	if err = c.Backend.Validate(); err != nil {
 		return err
 	}
--- a/yorkie/config.go
+++ b/yorkie/config.go
@@ -118,7 +118,7 @@ func (c *Config) Validate() error {
 		return err
 	}
 
-	// other validations will be here later.
+	// TODO(umi0410): Other validations will be here later.
 	// ...
 
 	return nil
--- a/yorkie/config.go
+++ b/yorkie/config.go
@@ -106,3 +106,13 @@ func newConfig(port int, metricsPort int, dbName string) *Config {
 		},
 	}
 }
+
+// Validate returns an error if the provided Config is invalidated.
+func (c *Config) Validate() error {
+	var err error
+
+	// some validations will be here.
+	// ...
+
+	return err
+}
--- a/yorkie/config.go
+++ b/yorkie/config.go
@@ -110,9 +110,16 @@ func newConfig(port int, metricsPort int, dbName string) *Config {
 // Validate returns an error if the provided Config is invalidated.
 func (c *Config) Validate() error {
 	var err error
+	
+	if err = c.Backend.Validate(); err != nil {
+		return err
+	}
+	if err = c.RPC.Validate(); err != nil {
+		return err
+	}
 
-	// some validations will be here.
+	// other validations will be here later.
 	// ...
 
-	return err
+	return nil
 }
--- a/yorkie/config.go
+++ b/yorkie/config.go
@@ -109,12 +109,10 @@ func newConfig(port int, metricsPort int, dbName string) *Config {
 
 // Validate returns an error if the provided Config is invalidated.
 func (c *Config) Validate() error {
-	var err error
-
-	if err = c.Backend.Validate(); err != nil {
+	if err := c.Backend.Validate(); err != nil {
 		return err
 	}
-	if err = c.RPC.Validate(); err != nil {
+	if err := c.RPC.Validate(); err != nil {
 		return err
 	}
 
--- a/yorkie/rpc/server.go
+++ b/yorkie/rpc/server.go
@@ -37,12 +37,12 @@ import (
 )
 
 var (
-	// ErrWrongPort occurs when the port in the config is invalid.
-	ErrWrongPort = errors.New("wrong port number for RPC server")
-	// ErrNoCertFile occurs when yorkie cannot find the cert file.
-	ErrNoCertFile = errors.New("no such cert file for RPC server")
-	// ErrNoKeyFile occurs when yorkie cannot find the key file.
-	ErrNoKeyFile = errors.New("no such key file for RPC server")
+	// ErrInvalidRPCPort occurs when the port in the config is invalid.
+	ErrInvalidRPCPort = errors.New("invalid port number for RPC server")
+	// ErrInvalidCertFile occurs when yorkie cannot find the cert file.
+	ErrInvalidCertFile = errors.New("no such cert file for RPC server")
+	// ErrInvalidKeyFile occurs when yorkie cannot find the key file.
+	ErrInvalidKeyFile = errors.New("no such key file for RPC server")
 )
 
 // Config is the configuration for creating a Server instance.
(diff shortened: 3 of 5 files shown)


hackerwins:
@umi0410 I think it's a good approach.

umi0410:
```shell
$ go run cmd/yorkie/*.go agent -c my-wrong-config.json
Error: not supported method for authorization webhook: this is wrong authorization method sample
Usage:
  yorkie agent [options] [flags]
...
```

I confirmed the original backend validation works well and the output of new validation is like the following.

```text
$ go run cmd/yorkie/*.go agent -c my-wrong-config.json
Error: wrong port number for grpc: -100
Usage:
  yorkie agent [options] [flags]
...
```
```text
$ go run cmd/yorkie/*.go agent -c my-config.json
Error: no such cert file for RPC server: fake.crt
Usage:
  yorkie agent [options] [flags]
...
```
```text
$ go run cmd/yorkie/*.go agent -c my-config.json
Error: no such key file for RPC server: fake.pem
Usage:
  yorkie agent [options] [flags]
...
```

internal/cli/agent.go
```diff
@@ -67,9 +67,7 @@ func newAgentCmd() *cobra.Command {
 			}
 
 			// TODO(hackerwins): We need to check that the entire config is valid.
```
hackerwins, line 69:
```suggestion
```
I think we can remove the comments here.

yorkie/config.go
```diff
@@ -106,3 +106,20 @@ func newConfig(port int, metricsPort int, dbName string) *Config {
...
+// Validate returns an error if the provided Config is invalidated.
+func (c *Config) Validate() error {
+	var err error
+
+	if err = c.Backend.Validate(); err != nil {
+		return err
+	}
+	if err = c.RPC.Validate(); err != nil {
+		return err
+	}
```
hackerwins, line 119: How about making this a little shorter?

```suggestion
	if err := c.Backend.Validate(); err != nil {
		return err
	}
	if err := c.RPC.Validate(); err != nil {
		return err
	}
```

yorkie/config.go
```diff
@@ -106,3 +106,20 @@ func newConfig(port int, metricsPort int, dbName string) *Config {
...
+
+	// other validations will be here later.
+	// ...
```
hackerwins, line 122:
```suggestion
	// TODO(umi0410): Other validations will be here later.
	// ...
```

yorkie/rpc/server.go
```diff
@@ -125,3 +136,26 @@ func (s *Server) listenAndServeGRPC() error {
 
 	return nil
 }
+
+// Validate validates port number and checks if files exists
+func (c *Config) Validate() error {
+	var err error
+	if c.Port < 0 || 65535 < c.Port {
+		return fmt.Errorf("%w: %d", ErrWrongPort, c.Port)
+	}
```
hackerwins, line 145:
```suggestion
	if c.Port < 1 || 65535 < c.Port {
		return fmt.Errorf("must be between 1 and 65535, given %d : %w", c.Port, ErrInvalidPortNumber)
	}
```

1. A port of 0 is not a normal value.
2. It would be nice to place the sentinel error at the end: https://blog.golang.org/go1.13-errors#TOC_5

yorkie/rpc/server.go
```diff
@@ -125,3 +136,26 @@ func (s *Server) listenAndServeGRPC() error {
...
+
+	// when specific cert or key file are configured
+	if c.CertFile != "" {
+		if _, err = os.Stat(c.CertFile); err != nil {
+			return fmt.Errorf("%w: %s", ErrNoCertFile, c.CertFile)
+		}
```
hackerwins, line 151: `os.Stat` can return various errors, and whether the file exists can be checked with `os.IsNotExist`. It would be better to use sentinel error in a broader sense here.

```suggestion
		if _, err = os.Stat(c.CertFile); err != nil {
			return fmt.Errorf("%w: %s", ErrInvalidCertFile, c.CertFile)
		}		
```

yorkie/rpc/server_test.go
```diff
@@ -352,3 +352,21 @@ func TestRPCServerBackend(t *testing.T) {
 		assert.Equal(t, codes.FailedPrecondition, status.Convert(err).Code())
 	})
 }
+
+func TestConfig_Validate(t *testing.T) {
+	scenarios := []*struct {
+		config *rpc.Config
+		expect error
```
hackerwins, line 359:
```suggestion
		expected error
```

yorkie/rpc/server.go
```diff
@@ -34,6 +36,15 @@ import (
 	"github.com/yorkie-team/yorkie/yorkie/rpc/interceptors"
 )
 
+var (
+	// ErrWrongPort occurs when the port in the config is invalid.
+	ErrWrongPort = errors.New("wrong port number for RPC server")
```
hackerwins, line 41: I think it would be better to use the expression invalid.
```suggestion
	ErrInvalidRPCPort = errors.New("invalid port number for RPC server")
```

hackerwins (commented):
Thanks for the first contribution. 👍 
I left a few issues. please check.

hackerwins:
This PR is a partial solution to #206, so it would be good to change `Fixes` to `Address`.

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

```
Fixes #206 

to

Address #206
```

yorkie/config.go
umi0410: LGTM

yorkie/rpc/server.go
umi0410: I didn't know that! Thank you.

yorkie/rpc/server.go
```diff
@@ -34,6 +36,15 @@ import (
 	"github.com/yorkie-team/yorkie/yorkie/rpc/interceptors"
 )
 
+var (
+	// ErrInvalidRPCPort occurs when the port in the config is invalid.
+	ErrInvalidRPCPort = errors.New("invalid port number for RPC server")
+	// ErrInvalidCertFile occurs when yorkie cannot find the cert file.
+	ErrInvalidCertFile = errors.New("no such cert file for RPC server")
+	// ErrInvalidKeyFile occurs when yorkie cannot find the key file.
+	ErrInvalidKeyFile = errors.New("no such key file for RPC server")
```
hackerwins, line 45:
```suggestion
	// ErrInvalidCertFile occurs when the certificate file is invalid.
	ErrInvalidCertFile = errors.New("invalid cert file for RPC server")
	// ErrInvalidKeyFile occurs when the key file is invalid.
	ErrInvalidKeyFile = errors.New("invalid key file for RPC server")
```

yorkie/rpc/server.go
```diff
@@ -125,3 +136,27 @@ func (s *Server) listenAndServeGRPC() error {
...
+	if c.Port < 1 || 65535 < c.Port {
+		return fmt.Errorf("must be between 1 and 65535, given %d: %w", c.Port, ErrInvalidRPCPort)
+	}
+
+	// when specific cert or key file are configured
+	if c.CertFile != "" {
+		_, err := os.Stat(c.CertFile)
+		if os.IsNotExist(err); err != nil {
+			return fmt.Errorf("%w: %s", ErrInvalidCertFile, c.CertFile)
+		}
```
hackerwins, line 151:
```suggestion
		if _, err := os.Stat(c.CertFile); err != nil {
			return fmt.Errorf("%s: %w", c.CertFile, ErrInvalidCertFile)
		}
```

yorkie/rpc/server.go
```diff
@@ -125,3 +136,27 @@ func (s *Server) listenAndServeGRPC() error {
...
+	}
+
+	if c.KeyFile != "" {
+		_, err := os.Stat(c.KeyFile)
+		if os.IsNotExist(err); err != nil {
+			return fmt.Errorf("%w: %s", ErrInvalidKeyFile, c.KeyFile)
+		}
+	}
```
hackerwins, line 159:
```suggestion
		if _, err := os.Stat(c.KeyFile); err != nil {
			return fmt.Errorf("%s: %w", c.KeyFile, ErrInvalidKeyFile)
		}
	}
```

yorkie/rpc/server.go
```diff
@@ -125,3 +136,27 @@ func (s *Server) listenAndServeGRPC() error {
 
 	return nil
 }
+
+// Validate validates port number and checks if files exists
```
hackerwins, line 140:
```suggestion
// Validate validates the port number and the files for certification.
```

hackerwins (changes requested):
I left a few more comments.

yorkie/rpc/server.go
umi0410: I thought you wanted to use `os.IsNotExist()` not `os.Stat()`. I'm sorry for misunderstanding. As you wrote, I think this is a better approach because `os.Stat()` can detect more errors and in that case `ErrInvalidCertFile` would be more proper than `ErrNoCertFile` or `ErrNoKeyFile`. Thank you.

hackerwins (approved):
LGTM 👍

PR #224, opened 2021-08-09
Improve agent's options

### LINKED ISSUE #168: Change `XXXTimeoutSec` flags in CLI to duration

dc7303:
- [x] Improve time-related options like `XXXTimeoutSec`. e.g `MongoTimeoutSec: 5` -> `MongoTimeout: 5s`
- [x] Clean up option names that are too long.
- [x] Set default value when read config file

Fixes #168

1st review request: https://github.com/yorkie-team/yorkie/pull/224#issuecomment-895015490

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


dc7303:
1st review request.
First, improved `mongo-connection-timeout-sec`.
 It was not easy to determine the type of `mongo.Config.ConnectionTimeout`. This is because if it is `time.Duration`, it cannot be entered in the json file as a string type like `5s`.
So, when Dial(), the logic converts it to Duration type.

Please review, and if you have a better improvement method than this method, please give your honest opinion!

hackerwins:
@dc7303 Thanks for sharing. 👍 
It looks like this way needs validation. How about starting with the timeout of RPC configuration to check the structure with validation logic?

dc7303:
@hackerwins Thank you so much for your comment 😄 
I'll ask a question because I need to make sure I understand your thoughts properly.

Currently, when an incorrect duration value is entered as a command option or a json file, a related error is displayed to the user.

### case 1
```bash
$ ./bin/yorkie agent -c ./yorkie/config.sample.json --mongo-connection-timeout="5"

Error: invalid argument "5" for "--mongo-connection-timeout" flag: time: missing unit in duration "5"
Usage:
  yorkie agent [options] [flags]

Flags:
      --authorization-webhook-cache-authorized-ttl-sec uint     TTL value to set when caching authorized webhook response. (default 10)
      --authorization-webhook-cache-unauthorized-ttl-sec uint   TTL value to set when caching unauthorized webhook response. (default 10)
....
```

### case 2
```json
// config.sample.json
{
  ...
  "Mongo": {
    "ConnectionTimeout": "s",  // invalid value
    ...
  },
  ...
}
```
```bash
$ ./bin/yorkie agent -c ./yorkie/config.sample.json

Error: time: invalid duration "s"
Usage:
  yorkie agent [options] [flags]

Flags:
      --authorization-webhook-cache-authorized-ttl-sec uint     TTL value to set when caching authorized webhook response. (default 10)
      --authorization-webhook-cache-unauthorized-ttl-sec uint   TTL value to set when caching unauthorized webhook response. (default 10)
....
```


I thought that if the error message in case 2 (error when mongo is dialed) was made clearer, users could improve their mistakes. And I didn't validate the improved duration because we are giving a push to a user.
Is your intention to have all these error validations happen inside `Config.Validate()`?

hackerwins:
@dc7303 Thanks to you, I see that case1 and case2 have different ways of returning errors.

I have two additional questions.

1. Where do case1 and case2 return an error?
2. How can we express the message in detail in case2?

dc7303:
@hackerwins 
**1. Where do case1 and case2 return an error?**
Case1 is raised by `cobra.Command.Execute()`. You can also try outputting errors directly from [internal/cli/commands.go](https://github.com/yorkie-team/yorkie/blob/main/internal/cli/commands.go#L30).

Case2 occurs when the `ConnectionTimeout` parameter value is incorrect when the [`ParseDuration` method is called](https://github.com/yorkie-team/yorkie/pull/224/files#diff-94887ea9e8dc284bbc89afc889cbdfac2212c03ad662d75c9063716ff6107c5dR53-R56) before the DB connection.

**2. How can we express the message in detail in case2?**
It seems that it can be improved by wrapping the error in the location where the aforementioned Case2 occurs.

hackerwins:
@dc7303 Thanks for the detailed reply.

It would be better to use `Config.Validate` pattern for case2 to manage validation logic consistently. And I think the other parts of this PR are good.

test/integration/auth_webhook_test.go
```diff
@@ -264,9 +264,9 @@ func TestAuthWebhook(t *testing.T) {
 			reqCnt++
 		}))
 
-		var unauthorizedTTLSec uint64 = 1
+		unauthorizedTTLSec := 1 * time.Second
```
hackerwins, line 267:
```suggestion
		unauthorizedTTL := 1 * time.Second
```

test/integration/auth_webhook_test.go
```diff
@@ -206,9 +206,9 @@ func TestAuthWebhook(t *testing.T) {
 			}
 		}))
 
-		var authorizedTTLSec uint64 = 1
+		authorizedTTLSec := 1 * time.Second
```
hackerwins, line 209: We can drop unit-specific suffix now because we replace the type with time.Duration.
```suggestion
		authorizedTTL := 1 * time.Second
```

yorkie/auth/webhook.go
```diff
@@ -100,14 +100,20 @@ func VerifyAccess(ctx context.Context, be *backend.Backend, info *types.AccessIn
 		return resp.StatusCode, nil
 	}); err != nil {
 		if errors.Is(err, ErrNotAllowed) {
-			unauthorizedTTL := time.Duration(be.Config.AuthorizationWebhookCacheUnauthorizedTTLSec) * time.Second
+			unauthorizedTTL, ttlParseErr := time.ParseDuration(be.Config.AuthWebhookCacheUnauthTTL)
```
hackerwins, line 103: How about moving parsing logic to Config? Since the value is validated, if an error is returned during parsing, it can be treated as panic.
```suggestion
			unauthorizedTTL := be.Config.ParseAuthWebhookCacheUnauthTTL()
```

yorkie/backend/db/mongo/config_test.go
```diff
@@ -0,0 +1,29 @@
+package mongo_test
+
+import (
+	"testing"
+
+	"github.com/yorkie-team/yorkie/yorkie/backend/db/mongo"
+
+	"github.com/stretchr/testify/assert"
+)
```
hackerwins, line 9:
```suggestion
	"testing"

	"github.com/stretchr/testify/assert"

	"github.com/yorkie-team/yorkie/yorkie/backend/db/mongo"	
)
```

hackerwins (approved):
LGTM. Thanks for making it clean.
I left minor comments.

yorkie/auth/webhook.go
dc7303: Thank you so much for your comments.
I share my question and my opinion on this.

1. Do you mean changing all the options for converting from string type to duration to the pattern you suggested?
For example, `AuthWebhookMaxWaitInterval`, `AuthWebhookCacheAuthTTL`, etc.

2. Although the possibility of `be.Config.AuthWebhookCacheUnauthTTL` being an incorrect value was prevented through validation when the Agent was started, an error was returned so that the user could recognize it if it was wrong. 
However, there is also a way to handle ParseDuration without returning an error. `ParseDuration()` by default returns 0 if an invalid string parameter is passed. So, if an error occurs in ParseDuration, unauthorizedTTL can be treated as 0 and an error log can be output.
***Is it correct that the `ParseAuthWebhookCacheUnauthTTL()` you suggested is intended for the aforementioned behavior?***

dc7303: It's corrected. please check!

hackerwins:
@dc7303 Thank you for working by reflecting all the minor requests. 👍

PR #237, opened 2021-08-25
Change config type json to yaml

### LINKED ISSUE #223: Support yaml as a config file
I'd like to yorkie support `.yaml` or other config formats. 
**Why is this needed**:

https://github.com/yorkie-team/yorkie/blob/7c254bb96d68b1f4db4164901918f2acaf8d5867/yorkie/config.go#L81

`yorkie` supports only `json` config files. However, `yaml` is also a good format to contains application configurations so it would be nice if `yaml` is supported.

parkjh688:
**This is created to change config type json to yaml**:

Fixes #223

**Special notes for your reviewer**: I need to check did I make config yaml file and the read config part correctly.

```release-note
User can use yaml config file instead json config file.
```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Didn't break anything

### DIFF
--- a/yorkie/config.go
+++ b/yorkie/config.go
@@ -14,172 +14,100 @@
  * limitations under the License.
  */
 
-package yorkie
+ package yorkie
 
-import (
-	"encoding/json"
-	"fmt"
-	"os"
-	"path/filepath"
-	"time"
-
-	"github.com/yorkie-team/yorkie/internal/log"
-	"github.com/yorkie-team/yorkie/yorkie/backend"
-	"github.com/yorkie-team/yorkie/yorkie/backend/db/mongo"
-	"github.com/yorkie-team/yorkie/yorkie/backend/sync/etcd"
-	"github.com/yorkie-team/yorkie/yorkie/metrics/prometheus"
-	"github.com/yorkie-team/yorkie/yorkie/rpc"
-)
-
-// Below are the values of the default values of Yorkie config.
-const (
-	DefaultRPCPort     = 11101
-	DefaultMetricsPort = 11102
-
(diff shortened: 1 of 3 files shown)


yorkie/config.go
```diff
@@ -14,172 +14,100 @@
...
-		},
-		Mongo: &mongo.Config{
-			ConnectionURI:     DefaultMongoConnectionURI,
-			ConnectionTimeout: DefaultMongoConnectionTimeout.String(),
-			PingTimeout:       DefaultMongoPingTimeout.String(),
-			YorkieDatabase:    dbName,
-		},
-	}
-}
+ package yorkie
```
hackerwins, line 17: It looks like every line in the file has a single space added. Can you please check the editor settings?
```suggestion
package yorkie
```

hackerwins (changes requested):
Thanks for your contribution.
Please resolve conflicts with the main branch.

dc7303 (changes requested):
In order to merge this PR, it seems better to fix the failed test case first.
I checked the error that occurred. The cause is that the yaml file is not unmarshaled properly.
Please refer to [this document](https://pkg.go.dev/reflect#StructTag). And try adding a yaml tag to yorkie's Config struct.

```go
type Config struct {
	RPC     *rpc.Config        `json:"RPC" yaml:"RPC"`
	Metrics *prometheus.Config `json:"Metrics" yaml:"Metrics"`
	Backend *backend.Config    `json:"Backend" yaml:"Backend"`
	Mongo   *mongo.Config      `json:"Mongo" yaml:"Mongo"`
	ETCD    *etcd.Config       `json:"ETCD" yaml:"ETCD"`
}
```

PR #240, opened 2021-08-27
Draft of supporting thread-safe client

### LINKED ISSUE #190: Improve the client to be safe in multiple Goroutines

Improve the client to be safe in multiple Goroutines.

**Why is this needed**:

Go Client is likely to be used on the server, so it is better to be safe in multiple routines, such as other database drivers.

shmruin:

Try to make `client` thread-safe. We can use one client for multiple request with go routines.

Fixes #190 

This is draft PR because cannot pass my own test of `test/integration/client_concurrency_test.go`.
How can I make test that verifying this is a thread-safe client?
Multiple other tests also fails when `test/integration/client_concurrency_test.go` added.
Also, I doubt if this support real thread-safe way in Yorkie client.

```release-note
NONE
```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Didn't break anything

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -19,6 +19,7 @@ package client
 import (
 	"context"
 	"errors"
+	"sync"
 
 	"github.com/google/uuid"
 	"google.golang.org/grpc"
@@ -73,6 +74,14 @@ type Client struct {
 	metadataInfo types.MetadataInfo
 	status       status
 	attachments  map[string]*Attachment
+
+	// A mutex for client id, key, metadataInfo, status changed
+	// gRPC itself statisfy concurrency, so conn and client are NOT included
+	// attachments is NOT included - it use it's own 'attachmentMutex' below
+	clientMutex sync.Mutex
+
+	// A mutex for Attachment struct
+	attachmentMutex sync.Mutex
 }
 
 // Option configures how we set up the client.
@@ -225,8 +234,10 @@ func (c *Client) Activate(ctx context.Context) error {
 		return err
 	}
 
+	c.clientMutex.Lock()
 	c.status = activated
 	c.id = clientID
+	c.clientMutex.Unlock()
 
 	return nil
 }
@@ -245,7 +256,9 @@ func (c *Client) Deactivate(ctx context.Context) error {
 		return err
 	}
 
+	c.clientMutex.Lock()
 	c.status = deactivated
+	c.clientMutex.Unlock()
 
 	return nil
 }
@@ -284,10 +297,13 @@ func (c *Client) Attach(ctx context.Context, doc *document.Document) error {
 	}
 
 	doc.SetStatus(document.Attached)
+
+	c.attachmentMutex.Lock()
 	c.attachments[doc.Key().BSONKey()] = &Attachment{
 		doc:   doc,
 		peers: make(map[string]types.MetadataInfo),
 	}
+	c.attachmentMutex.Unlock()
 
 	return nil
 }
@@ -332,7 +348,10 @@ func (c *Client) Detach(ctx context.Context, doc *document.Document) error {
 	}
 
 	doc.SetStatus(document.Detached)
+
+	c.attachmentMutex.Lock()
 	delete(c.attachments, doc.Key().BSONKey())
+	c.attachmentMutex.Unlock()
 
 	return nil
 }
@@ -391,10 +410,12 @@ func (c *Client) Watch(
 					return nil, err
 				}
 
+				c.attachmentMutex.Lock()
 				attachment := c.attachments[docID]
 				for _, cli := range clients {
 					attachment.peers[cli.ID.String()] = cli.MetadataInfo
 				}
+				c.attachmentMutex.Unlock()
 			}
 
 			return nil, nil
@@ -417,6 +438,7 @@ func (c *Client) Watch(
 						return nil, err
 					}
 
+					c.attachmentMutex.Lock()
 					attachment := c.attachments[k.BSONKey()]
 					if eventType == types.DocumentsWatchedEvent ||
 						eventType == types.MetadataChangedEvent {
@@ -427,6 +449,7 @@ func (c *Client) Watch(
 					} else {
 						delete(attachment.peers, cli.ID.String())
 					}
+					c.attachmentMutex.Unlock()
 				}
 				return &WatchResponse{
 					Type:          PeersChanged,
@@ -472,8 +495,10 @@ func (c *Client) UpdateMetadata(ctx context.Context, k, v string) error {
 		return ErrClientNotActivated
 	}
 
+	c.clientMutex.Lock()
 	c.metadataInfo.Data[k] = v
 	c.metadataInfo.Clock++
+	c.clientMutex.Unlock()
 
 	if len(c.attachments) == 0 {
 		return nil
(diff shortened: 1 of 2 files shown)


shmruin:
I append attach/detach client test, but it fails. I guess the reason is that gRPC not accept detach with not detached document. Is there any way to make gRPC ignore if document not exists?

dc7303 (changes requested):
Thank you for your work! 👍 
I want to help you solve this problem :D
I left a few suggestions and questions.

And I will continue to review carefully and ask questions.

**Question**
There is code to check a client's status and access Attachment in `UpdateMetadata`, `Sync`, and `sync` methods. I think RLock is also needed here, is there any reason for excluding it?

client/client.go
```diff
@@ -207,9 +217,12 @@ func (c *Client) Close() error {
 // and receives a unique ID from the agent. The given ID is used to distinguish
 // different clients.
 func (c *Client) Activate(ctx context.Context) error {
+	c.clientMutex.RLock()
 	if c.status == activated {
+		c.clientMutex.RUnlock()
 		return nil
 	}
+	c.clientMutex.RUnlock()
```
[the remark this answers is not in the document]
shmruin: Yes, I think your suggestion is reasonable. And I missed the point of other `UpdateMetadata `, `Sync`, `sync`. I will check that points. Thanks.

shmruin:
Hi, I twist my test that make 5 client attach in go routine and 5 dettach in other goroutine. The client pass the test. My question is, what if go routine on same document attach/detach concurrently? **Like if I attach/detach same document of `{"k1", "v1"}`, then it FAILS, because it cannot guarantee the order of go routine**. So, `Detach - Detach - Attach - Attach` surely fail as no document exists at first. Even, I'm not sure `Attach - Detach - Attach - Detach` and no document exists is what really the user want. I wonder what should I do for this intersect A/D case on same document.   
Plus, I'm working on `Watch` in client and make other tests soon.

client/client.go
```diff
@@ -225,17 +235,22 @@ func (c *Client) Activate(ctx context.Context) error {
 		return err
 	}
 
+	c.clientMutex.Lock()
 	c.status = activated
```
dc7303, line 239: Yorkie's client status is also managed by the DB.
I am currently concerned about the range of Mutex. This is because I expect that the client state of the actual DB and the state of the Client struct may be different.

The situation I assumed is the following:
image


So, I think that even if you call Activate and Deactivate in multiple routines, you need a test in which the database and instance state match.
What do you guys think?

test/integration/client_concurrency_test.go
```diff
@@ -0,0 +1,158 @@
...
+			}(i)
+		}
+
+		wg.Wait()
+
+		if !cli.IsActive() {
+			cli.Activate(ctx)
+		}
+
+		// All Activate in goroutine
```
dc7303, line 68: These codes are simply trying to call Activate() and Deactivate() from several routines. Can you explain a bit more about what you want to check with this test code?

dc7303 (commented):
Added some comments and questions. please check :)

client/client.go
shmruin: Yes, I think I need some Activate and Deactivate test in goroutines, and check the final condition. But the problem is, that even if I wait for all goroutines in activate/deactivate in that test, I'm not sure that it guarantees the client's state (ActivateClient or so on). Waiting for the goroutines is not enough. This makes me really hard even the other test cases that I try to append (like attach/detach test and Watch test). gRpc side is a whole black box in this test. I know that this gRpc itself is concurrent, but doubt it make this case to be solved.

test/integration/client_concurrency_test.go
shmruin: Sadly, I realize that this is meaningless for the time being. I guess the final test should be some kind of testing Activate/Deactivate several times and check the final state. Of course including the ActivateClient you argue above. I'm stuck in that stage.

client/client.go
dc7303: I sympathize with the problem you are experiencing. Let's discuss it together and solve it one by one. :)
The integration test is a black-box test, so there is no need to access the low-level (db) and compare directly.
Instead, it seems to be possible to compare using the API used by the user.
I wrote a temporary test to reproduce the problem I mentioned.

For Example:
```go
        MAX_ROUTINES := 5

	t.Run("concurrent activate/deactivate test", func(t *testing.T) {
		cli, err := client.Dial(defaultAgent.RPCAddr())
		assert.NoError(t, err)
		defer func() {
			err := cli.Close()
			assert.NoError(t, err)
		}()

		ctx := context.Background()
		wg := gosync.WaitGroup{}
		doc := document.New(helper.Collection, t.Name())

		// Activate <-> Deactivate in goroutine
		for i := 0; i <= 50; i++ {
			for i := 1; i <= MAX_ROUTINES; i++ {
				wg.Add(1)
				go func(idx int) {
					defer wg.Done()
					time.Sleep(time.Duration(rand.Intn(5)+4) * time.Millisecond)
					if idx%2 == 0 {
						assert.NoError(t, cli.Deactivate(ctx))
						return
					}
					assert.NoError(t, cli.Activate(ctx))
				}(i)
			}

			wg.Wait()
			if cli.IsActive() {
				assert.NoError(t, cli.Attach(ctx, doc))
				assert.NoError(t, cli.Detach(ctx, doc))
				assert.NoError(t, cli.Deactivate(ctx))
			}
		}
	})
```

In this test, using the `cli.Attach()` API to test the final condition of a client.
And this test currently throws the following failure:

```bash
.........
2021-09-25T17:24:32.550+0900	WARN	RPC : "/api.Yorkie/AttachDocument" 3.684915ms: "client_id:\"aN\\334\\300\\027\\220\\261\\232S\\347[V\" change_pack:<document_key:<collection:\"test-collection\" document:\"TestClientConcurrency/concurrent_activate/deactivate_test\" > checkpoint:<> > " => "client not activated"
2021-09-25T17:24:32.550+0900	ERROR	rpc error: code = FailedPrecondition desc = client not activated
github.com/yorkie-team/yorkie/client.(*Client).Attach
	/Users/scott/Workspace/dc7303/yorkie/client/client.go:303
command-line-arguments.TestClientConcurrency.func1
	/Users/scott/Workspace/dc7303/yorkie/test/integration/client_concurrency_test.go:67
testing.tRunner
	/Users/scott/.gvm/gos/go1.17/src/testing/testing.go:1259
2021-09-25T17:24:32.552+0900	INFO	RPC : "/api.Yorkie/DeactivateClient" 1.181214ms
--- FAIL: TestClientConcurrency (0.02s)
    --- FAIL: TestClientConcurrency/concurrent_activate/deactivate_test (0.02s)
        client_concurrency_test.go:67:
            	Error Trace:	client_concurrency_test.go:67
            	Error:      	Received unexpected error:
            	            	rpc error: code = FailedPrecondition desc = client not activated
            	Test:       	TestClientConcurrency/concurrent_activate/deactivate_test
FAIL
2021-09-25T17:24:32.556+0900	INFO	backend stoped: id: c57dpg6a0brql6r6i91g, rpc: localhost:21201
FAIL	command-line-arguments	0.313s
FAIL
```

It's not always a failing test. Because testing concurrency can sometimes be inconsistent.
Still, through these tests, I found that the issue I mentioned earlier can be reproduced and that this is a situation that users may encounter while using it.
The issue of adding this inconsistent test remains to be discussed further, but it is meaningful to know and deal with this situation.

How about improving by testing the API one after the other while writing the integration test from the user's point of view? After resolving `activate/deactivate`, I think that it will be possible to complete it while thinking about `attach/detach`, `watch`, etc. separately.

shmruin: Thanks! I really look forward to discuss this on sync time.

PR #243, opened 2021-08-28
make GarbageCollect function private

### LINKED ISSUE #125: Change Document.GarbageCollect to private method
https://github.com/yorkie-team/yorkie/blob/83881a34b6492535625f551d666103806331f0e5/pkg/document/document.go#L197

Currently, `Document.GarbageCollect()` is exposed to the outside. This method is not provided to the user, it needs to be hidden.

We are moving the responsibility for testing GarbageCollect to the JSON package through issue #118.
And I think `Document.GarbageCollect()` can be hidden when `document_text.go` is cleaned up. What do you think? I'm curious about your opinion.

codinggosu:
makes the GarbageCollect private for encapsulation

Fixes #125 

```release-note
NONE, the only place the function is used is internally, and in tests.
Internally, there are no logic changes, only name changes,
For tests, to use a private function, I created a new public function that simply wraps the private function. 
I don't know how helpful this is, since this is just an indirection, but I tried to follow the steps mentioned [here](https://newbedev.com/how-to-test-a-unexported-private-function-in-go-golang) and [here](https://stackoverflow.com/questions/36009327/how-to-access-private-functions-of-another-package-in-golang/36011850)
```

**Additional documentation**:

<!--
This section can be blank if this pull request does not require a relherelin h


codinggosu:
trying to fix the CI error (ERROR context deadline exceeded) now

codinggosu:
just a thought, can "reflection" be used to solve this issue, instead of doing what I did?

dc7303:
Thank you so much for the work! :)
The reason we want to make GarbageCollect private is that we don't want to expose it to users.
However, in the code you worked on, you can see that a public method with a new name has been added. So it is still exposed.

The reason we cannot easily change GarbageCollect to private is that we test this function by calling it directly in the test. So I think this problem has something to do with the test scope.

> just a thought, can "reflection" be used to solve this issue, instead of doing what I did?

Can you explain a bit more about this? It would be better if you could share the sample code. And I think it would be good to talk together later.

codinggosu:
@dc7303  The idea that I had in mind was that maybe golang had 'reflection' that was similar to 'reflection' in Java, which I thought could be used in this case. 
Upon further research, I discovered that reflection in Go is a little bit different. 
However, these two options seem promising:
1. using the 'unsafe' package: https://play.golang.org/p/K9PmMb7qTYb
2. something a lot more complicated than just using a package: http://www.alangpierce.com/blog/2016/03/17/adventures-in-go-accessing-unexported-functions/
I haven't tried any of these methods, and am unsure about either of them, partially because I don't have a lot of experience in Go.
If you think that these two methods can be applied or are appropriate, I will be happy to try them

PR #252, opened 2021-10-04
Thread safe client

### LINKED ISSUE #190: Improve the client to be safe in multiple Goroutines

shmruin:

In this PR, I tried to solve this problem in TWO STEPS.

> 1. Guarantee concurrency between Activate/Deactivate **in one client.**
> 2. (Will Support) Guarantee concurrency between Attach/Detach **in one client's one document**.

I think in this way, we can **make users re-use their client**.

`Because, the original way of supporting high-level concurrency is way too hard at this point. We need to make this problem smaller.`

Like, if they create once, then they can use forever before the concrete end. (Never be in [this situation](https://github.com/metis-labs/metis-server/blob/280633d9541f98c776df5b1508ede1d25d8c5762/server/projects/projects.go#L44-L78) again)

I draw some simple user situation below...

Client Concurrency User Scenario

This PR is step ONE(Activate/Deactivate), supporting client's concurrency and test is included.

I'm gonna jump into step TWO, but not sure I use another PR for this or make another PR for that (any opinion?)

Address #190 

```release-note
NONE
```

**Additional documentation**:

```docs

```

**Checklist**:
- [*] Added relevant tests or not required
- [*] Didn't break anything


client/client.go
```diff
@@ -207,7 +212,11 @@ func (c *Client) Close() error {
...
 // different clients.
 func (c *Client) Activate(ctx context.Context) error {
-	if c.status == activated {
+	// Apply critical section for whole Activate
+	defer c.clientMutex.Unlock()
+	c.clientMutex.Lock()
+
+	if c.IsActive() {
 		return nil
 	}
```
dc7303, line 221:
```suggestion
	if c.IsActive() {
		return nil
	}
	
        defer c.clientMutex.Unlock()
	c.clientMutex.Lock()
```
In my opinion, it is okay to set a read-lock in the c.IsActive() method, and then put a write-lock on the logic that changes the resource afterward. 
How about setting up your scope like this?

dc7303 (changes requested):
Thanks for the work! I've made a few suggestions, so please check them out!
1. I agree with the sentence below. I've been looking for ways to test this, but I think it's a hard problem. How about going through this thread without adding any tests?
```
Testing will show you that -- as long as you're not unlucky -- it all worked. But a test is a blunt instrument; it can fail to execute the exact right sequence to cause failure.
```

https://stackoverflow.com/a/2380892

2. Address keyword is used when PR doesn't completely solve the issue.
Fixes # 190 -> Address # 190

hackerwins:
Thanks for the work.

But this PR is not complete and it looks like it will be difficult to merge yet. It would be better to start over later.

PR #290, opened 2022-01-22
Patch loglevel flags

### LINKED ISSUE #279: Add `--log-level` flag to agent command.
**Description**:

Add `--log-level` flag to [agent command](https://github.com/yorkie-team/yorkie/blob/main/internal/cli/agent.go).

Configures log level with `debug`, `info`, `warn`, `error`, `panic`, or `fatal`.

**Why**:

- We can configure the log level on starting the agent.

zacscoding:

Add `--log-level` flags like below commands.
(I think config file is also need changes if provide more configs such as encoding)

```
$ yorkie agent --log-level warn
```

Fixes #279 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/internal/cli/agent.go
+++ b/internal/cli/agent.go
@@ -101,7 +101,9 @@ func newAgentCmd() *cobra.Command {
 				conf = parsed
 			}
 
-			log.SetLogLevel(flagLogLevel)
+			if err := log.SetLogLevel(flagLogLevel); err != nil {
+				return err
+			}
 
 			y, err := yorkie.New(conf)
 			if err != nil {
--- a/internal/cli/agent.go
+++ b/internal/cli/agent.go
@@ -28,6 +28,7 @@ import (
 	"github.com/yorkie-team/yorkie/yorkie"
 	"github.com/yorkie-team/yorkie/yorkie/backend/db/mongo"
 	"github.com/yorkie-team/yorkie/yorkie/backend/sync/etcd"
+	"github.com/yorkie-team/yorkie/yorkie/log"
 )
 
 var (
@@ -36,6 +37,8 @@ var (
 
 var (
 	flagConfPath string
+
+	flagLogLevel string
 
 	housekeepingInterval            time.Duration
 	housekeepingDeactivateThreshold time.Duration
@@ -97,6 +100,8 @@ func newAgentCmd() *cobra.Command {
 				}
 				conf = parsed
 			}
+
+			log.SetLogLevel(flagLogLevel)
 
 			y, err := yorkie.New(conf)
 			if err != nil {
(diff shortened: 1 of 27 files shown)


internal/cli/agent.go
```diff
@@ -18,6 +18,7 @@ package cli
 
 import (
 	"fmt"
+	"github.com/yorkie-team/yorkie/yorkie/log"
```
dc7303, line 21: Our import ordering is below.

- Standard library
- Third party library
- Yorkie

GoFmt helps you sort by each group. It's not perfect, so some corrections are needed.

yorkie/log/log.go
```diff
@@ -18,18 +18,53 @@ package log
...
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
-// Logger is the default logger used by Yorkie.
-var Logger *zap.SugaredLogger
+var defaultLogger *zap.SugaredLogger
+var coreLogger zapcore.Core
+var rawLogger *zap.Logger
```
dc7303, line 29: It seems rawLogger can only be used in the scope of the initialize() function, what do you think?

yorkie/log/log.go
```diff
@@ -18,18 +18,53 @@ package log
...
 
-// Core is a minimal, fast logger interface.
-var Core zapcore.Core
+var logLevel = zapcore.InfoLevel
+var loggerOnce sync.Once
 
-var rawLogger *zap.Logger
+// DefaultLogger returns the default logger used by Yorkie.
+func DefaultLogger() *zap.SugaredLogger {
```
dc7303, line 35: DefaultLogger() -> Logger()
CoreLogger() -> Core()

What do you think of this change?

dc7303 (changes requested):
Thank you so much for your contribution! 👍 
I've made some suggestions, so please leave your comments!

internal/cli/agent.go
zacscoding: I will follow this project style guide :)
Are there any scripts for checking this format is proper or not before commit?

yorkie/log/log.go
zacscoding: I removed it :)

yorkie/log/log.go
zacscoding: I changed it because `log` package name already implied logger and more simplified :) thx!

internal/cli/agent.go
dc7303: You can check the code style with `make lint`, but it doesn't check the import order very well. 😭

dc7303 (approved):
LGTM 👍 
It is judged that there are no functional problems. But there is one thing I want to discuss.

The default level is set to info, so even if the user makes a typo such as 'erro', it will work well with the info level. However, users are unaware of their mistakes.

How about adding validator logic to return an error when the user enters an incorrect level? What do you think?

zacscoding:
> How about adding validator logic to return an error when the user enters an incorrect level? What do you think?

At first, i just want to ignore invalid params because zap logger use `zap.NewNop()` if provide invalid configs.
So i checked other framework or application, where spring boot, docker(moby). They return an error if user use invalid log level values :)

I added validation of log level. thx @dc7303 👍

hackerwins:
@zacscoding Thanks for your contribution. 👍

PR #300, opened 2022-02-05
Load changes gradually when creating a snapshot

### LINKED ISSUE #228: Load changes incrementally when creating a snapshot

Load changes incrementally when creating a snapshot.

When a large number of changes are accumulated in a document, Agent replaces the changes with a snapshot and passing to the client(#9). When loading changes from DB to create snapshots, there may be many changes to be loaded at once.

https://github.com/yorkie-team/yorkie/blob/f7734e8a98eb5532d0ecec037c45f3a28ef0f318/yorkie/packs/packs.go#L319-L339

**Why is this needed**:
- Even when changes are accumulated a lot, Agent can reliably handle the client's request.

ejolie:
We have to load changes gradually from DB when a large number of changes are accumulated.

Fixes #228

I create a draft PR to ask for help.
I have two questions while trying to fix the issue. 😅 

1. How to determine the right chunk size of changes.
Is there any way to test the performance depending on the chunk size of changes in my local environment?

2. How to write test code to make it arrive at the code I changed.
I wrote test code referring to other one in `TestSnapshot`, but I noticed my code does not arrive at 195 line of `pushpull.go`. I don't know how to set up the situation in the test code.

Thanks in advance.

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Didn't break anything

### DIFF
--- a/internal/cli/agent.go
+++ b/internal/cli/agent.go
@@ -295,6 +295,12 @@ func init() {
 		yorkie.DefaultSnapshotInterval,
 		"Interval of changes to create a snapshot.",
 	)
+	cmd.Flags().Uint64Var(
+		&conf.Backend.ChangesChunkSize,
+		"backend-changes-chunk-size",
+		yorkie.DefaultChangesChunkSize,
+		"Chunk size of changes to load gradually when creating a snapshot",
+	)
 	cmd.Flags().StringVar(
 		&conf.Backend.AuthWebhookURL,
 		"auth-webhook-url",
--- a/test/helper/helper.go
+++ b/test/helper/helper.go
@@ -50,6 +50,7 @@ const (
 	MongoConnectionTimeout = "5s"
 	MongoPingTimeout       = "5s"
 	SnapshotThreshold      = 10
+	ChangesChunkSize       = 2
 	Collection             = "test-collection"
 
 	AuthWebhookMaxWaitInterval = 3 * gotime.Millisecond
@@ -111,6 +112,7 @@ func TestConfig(authWebhook string) *yorkie.Config {
 		},
 		Backend: &backend.Config{
 			SnapshotThreshold:          SnapshotThreshold,
+			ChangesChunkSize:           ChangesChunkSize,
 			AuthWebhookURL:             authWebhook,
 			AuthWebhookMaxWaitInterval: AuthWebhookMaxWaitInterval.String(),
 			AuthWebhookCacheSize:       AuthWebhookSize,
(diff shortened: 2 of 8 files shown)


hackerwins:
@ejolie Thanks for your contribution. Overall the code looks good.

> 1. How to determine the right chunk size of changes.
> Is there any way to test the performance depending on the chunk size of changes in my local environment?

I think we can test the performance by increasing the editing of the document. The benchmark below is an example running about 14,000 edits. If you use the code below, you can easily increase the amount of change.

https://github.com/yorkie-team/yorkie/blob/main/test/bench/text_editing_bench_test.go

If a snapshot already exists, the document is rebuilt from the snapshot, so it would be good to test by increasing the snapshot creation cycle.

https://github.com/yorkie-team/yorkie/blob/8e15126d3491dd514125cd699631217d26e832a1/internal/cli/agent.go#L292-L297

> 2. How to write test code to make it arrive at the code I changed.
> I wrote test code referring to other one in TestSnapshot, but I noticed my code does not arrive at 195 line of pushpull.go. I don't know how to set up the situation in the test code.

When I ran the test I wrote, it reached that line. How about printing a log on the line?

Screen Shot 2022-02-07 at 2 26 24 PM

ejolie:
> @ejolie Thanks for your contribution. Overall the code looks good.
> 
> > 1. How to determine the right chunk size of changes.
> >    Is there any way to test the performance depending on the chunk size of changes in my local environment?
> 
> I think we can test the performance by increasing the editing of the document. The benchmark below is an example running about 14,000 edits. If you use the code below, you can easily increase the amount of change.
> 
> https://github.com/yorkie-team/yorkie/blob/main/test/bench/text_editing_bench_test.go
> 
> If a snapshot already exists, the document is rebuilt from the snapshot, so it would be good to test by increasing the snapshot creation cycle.
> 
> https://github.com/yorkie-team/yorkie/blob/8e15126d3491dd514125cd699631217d26e832a1/internal/cli/agent.go#L292-L297
> 
> > 2. How to write test code to make it arrive at the code I changed.
> >    I wrote test code referring to other one in TestSnapshot, but I noticed my code does not arrive at 195 line of pushpull.go. I don't know how to set up the situation in the test code.
> 
> When I ran the test I wrote, it reached that line. How about printing a log on the line?
> 
> Screen Shot 2022-02-07 at 2 26 24 PM

Thanks for your reply.
I’d like to write benchmark test for “snapshot load changes test" to check the performance.
So I tried to create activated clients to be used in the test by referring to `createActivatedClients` function in `main_test.go` but an error occurred when activating a client.
```
rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp [::1]:21201: connect: connection refused
```
I wonder if it is possible to use the client(`client.Client`) in the benchmark test code.

hackerwins:
Unfortunately, I think it would be better to close this PR for now and open it when working again.

PR #314, opened 2022-04-30
fix: lru cache expired when same key update

ppeeou:

missing function updating ttl in lru cache
 

**Checklist**:
- [x] Didn't break anything


dc7303 (approved):
LGTM 👍

Issue #304, opened 2022-02-24
Update Go version to 1.18


hackerwins:
**Description**:

Update Go version to 1.18.

We need to check about changes from 1.17 to 1.18 and reflect the changes on the Yorkie.

https://tip.golang.org/doc/go1.18

Tasks

- [x] https://github.com/yorkie-team/yorkie/issues/234
- [x] Introduce Generics
  - [x] [cache.go](https://github.com/yorkie-team/yorkie/blob/main/pkg/cache/cache.go)
  - [x] [splay.go](https://github.com/yorkie-team/yorkie/blob/main/pkg/splay/splay.go)
  - [x] [priority_queue](https://github.com/yorkie-team/yorkie/blob/main/pkg/pq/priority_queue.go)
  - [x] [llrb.go](https://github.com/yorkie-team/yorkie/blob/main/pkg/llrb/llrb.go)
  - [x] [rga_tree_split](https://github.com/yorkie-team/yorkie/blob/main/pkg/document/json/rga_tree_split.go)

**Why**:

- We can reflect on the latest Go improvements and keep an up-to-date code.

dc7303:
We can use generics while updating to 1.18.

I think the list below would be good if generics were applied.

- [cache.go](https://github.com/yorkie-team/yorkie/blob/main/pkg/cache/cache.go)
- [splay.go](https://github.com/yorkie-team/yorkie/blob/main/pkg/splay/splay.go)
- [llrb.go](https://github.com/yorkie-team/yorkie/blob/main/pkg/llrb/llrb.go)
- [rga_tree_split](https://github.com/yorkie-team/yorkie/blob/main/pkg/document/json/rga_tree_split.go)

hackerwins:
@dc7303 Thanks for your comments. I added the tasks in the first comment.

PR #331, opened 2022-05-21
Improve the llrb tree using generic

dc7303:
Improve the `llrb.go` using generic.

Address #304

When a generic type is a return type, nil cannot be returned because of the zero-value issue.
It is possible to add a method that handles zero-value, but I think it is sufficient to handle it within the method. What do you think?

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/pkg/document/json/rga_tree_split.go
+++ b/pkg/document/json/rga_tree_split.go
@@ -286,7 +286,7 @@ func (t *RGATreeSplitNode) Value() RGATreeSplitValue {
 type RGATreeSplit struct {
 	initialHead *RGATreeSplitNode
 	treeByIndex *splay.Tree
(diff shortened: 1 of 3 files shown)


dc7303:
When generics are return types, we can't return nil, so we can handle zero-values ​​like this.
```go
var zeroK K
var zeroV V
return zeroK, zeroV
```

https://github.com/yorkie-team/yorkie/pull/331/files#diff-24b81e7ebed48258518bbf8ade164201ec590aec37112d003a3b05fd4126db15R133-R135

hackerwins (approved):
LGTM.

Thanks for your contribution.

PR #335, opened 2022-06-03
Introduce generics to cache.go

### LINKED ISSUE #304: Update Go version to 1.18

DongjinS:

Introduce generics to cache.go

Address #304 

I am not sure what I've done is applied generics correctly. 
Please review it and I will revise it.

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ fmt: ## applies format and simplify codes
 	gofmt -s -w $(GO_SRC)
 
 lint: ## runs the golang-ci lint, checks for lint violations
-	 golangci-lint run ./...
+	golangci-lint run ./...
 
 test: ## runs integration tests that require local applications such as MongoDB
 	go clean -testcache
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -24,8 +24,6 @@ import (
 	"errors"
 	"sync"
 	"time"
-
-	"github.com/yorkie-team/yorkie/api/types"
 )
 
 var (
@@ -35,7 +33,7 @@ var (
 
 // LRUExpireCache is a cache that ensures the mostly recently accessed keys are returned with
 // a ttl beyond which keys are forcibly expired.
-type LRUExpireCache[V types.AuthWebhookResponse] struct {
+type LRUExpireCache[V any] struct {
 	lock sync.Mutex
 
 	maxSize      int
@@ -44,7 +42,7 @@ type LRUExpireCache[V types.AuthWebhookResponse] struct {
 }
 
 // NewLRUExpireCache creates an expiring cache with the given size
-func NewLRUExpireCache[V types.AuthWebhookResponse](maxSize int) (*LRUExpireCache[V], error) {
+func NewLRUExpireCache[V any](maxSize int) (*LRUExpireCache[V], error) {
 	if maxSize <= 0 {
 		return nil, ErrInvalidMaxSize
 	}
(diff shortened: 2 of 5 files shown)


hackerwins:
Thanks for your first contribution. I left a brief comment.

It would be good to think about why we had to cast a value from the cache in the code below, and how Generics can solve it.
https://github.com/yorkie-team/yorkie/blob/main/server/rpc/auth/webhook.go#L65

Makefile
```diff
@@ -42,7 +42,7 @@ fmt: ## applies format and simplify codes
 	gofmt -s -w $(GO_SRC)
 
 lint: ## runs the golang-ci lint, checks for lint violations
-	 golangci-lint run ./...
+	golangci-lint run ./...
```
hackerwins, line 45: 👍

pkg/cache/cache.go
```diff
@@ -33,7 +35,7 @@ var (
 
 // LRUExpireCache is a cache that ensures the mostly recently accessed keys are returned with
 // a ttl beyond which keys are forcibly expired.
-type LRUExpireCache struct {
+type LRUExpireCache[V types.AuthWebhookResponse] struct {
```
hackerwins, line 38: It would be nice to be able to use various values ​​as well as `AuthwebhookResponse` for Cache. Since `LRUExpireCache` does not directly manipulate the Value type, it seems that the type constraint can be set in a wide range.
```suggestion
type LRUExpireCache[V any] struct {
```

hackerwins (commented):
Thanks for the correction. I left an additional comment.

pkg/cache/cache_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"github.com/stretchr/testify/assert"
 
 	"github.com/yorkie-team/yorkie/pkg/cache"
+
+	"github.com/yorkie-team/yorkie/api/types"
```
hackerwins, line 11:
```suggestion
	"github.com/yorkie-team/yorkie/api/types"
```

pkg/cache/cache_test.go
```diff
@@ -11,17 +11,17 @@ import (
...
 
-		lruCache, err = cache.NewLRUExpireCache(0)
+		lruCache, err = cache.NewLRUExpireCache[any](0)
 		assert.ErrorIs(t, err, cache.ErrInvalidMaxSize)
 		assert.Nil(t, lruCache)
 	})
 
 	t.Run("add test", func(t *testing.T) {
-		lruCache, err := cache.NewLRUExpireCache(1)
+		lruCache, err := cache.NewLRUExpireCache[any](1)
```
hackerwins, line 24: We can use `string` here to specify the type.
```suggestion
		lruCache, err := cache.NewLRUExpireCache[string](1)
```

hackerwins (commented):
Now overall it looks good.
I left an additional comment.

pkg/cache/cache_test.go
DongjinS: Thanks for comment!
I also try to use `string` here, but cache_test.go:40, `assert.Nil(t, response1)` makes error.
Error is `Expected nil, but got: ""`, so I use `any` to avoid this error.
Is there better way to solve this error?

hackerwins: It seems to be possible to modify it to assert empty string instead of `assert.Nil`.

DongjinS: Now, assert.Nil(t, response1) is replaced by assert.Empty(t, response1).
Thanks for nice advice!

DongjinS: Thanks for review:D
`assert.NotNil` is replaced with `assert.NotEmpty`

PR #336, opened 2022-06-03
Introduce generics to priority queue

### LINKED ISSUE #304: Update Go version to 1.18

Eithea:

Introduce generics to priority queue

To apply generics to priority queue, make interface "ComparableValue"
which is a 'Value' and comaparable and apply generic to all structs
and functions refer to 'Value'.

Address #304

First, I applyed generic to Value but there was un error in using '==' operator in func Release.
So I tried to push 'comparable' in interface Value but it didn't work.
So instead, I implemented interface ComparableValue which is a Value and comparable.
I'm not sure this is right way, please give an advice if there's a better.

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/pkg/pq/priority_queue.go
+++ b/pkg/pq/priority_queue.go
@@ -26,7 +26,7 @@ type PriorityQueue[V comparableValue] struct {
 }
 
 // NewPriorityQueue creates an instance of NewPriorityQueue.
-func NewPriorityQueue[V comparableValue] () *PriorityQueue[V] {
+func NewPriorityQueue[V comparableValue]() *PriorityQueue[V] {
 	pq := &internalQueue[V]{}
 	heap.Init(pq)
 
--- a/pkg/pq/priority_queue.go
+++ b/pkg/pq/priority_queue.go
@@ -21,43 +21,43 @@ import (
 )
 
 // PriorityQueue is a priority queue implemented with max heap.
-type PriorityQueue struct {
-	queue *internalQueue
+type PriorityQueue[V ComparableValue] struct {
+	queue *internalQueue[V]
 }
 
 // NewPriorityQueue creates an instance of NewPriorityQueue.
-func NewPriorityQueue() *PriorityQueue {
-	pq := &internalQueue{}
+func NewPriorityQueue[V ComparableValue] () *PriorityQueue[V] {
+	pq := &internalQueue[V]{}
 	heap.Init(pq)
 
-	return &PriorityQueue{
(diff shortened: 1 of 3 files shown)


pkg/pq/priority_queue.go
```diff
@@ -21,43 +21,43 @@ import (
...
+func (pq *PriorityQueue[V]) Len() int {
 	return pq.queue.Len()
 }
 
 // Release deletes the given value from this PriorityQueue.
-func (pq *PriorityQueue) Release(value Value) {
+func (pq *PriorityQueue[V]) Release(value V) {
 	idx := -1
 	for i, item := range *pq.queue {
 		if item.value == value {
```
hackerwins, line 63: It looks like we have the wrong code on this line. 

If all values ​​in the struct are comparable, we can still use `==`, but this can lead to mistakes. So we have to use `Less()` to check for equality.

```suggestion
		if !item.value.Less(value) && !value.Less(item.value) {
```

After changing this code we can use `Value` directly as a type parameter.

hackerwins (commented):

I left a short comment on the question.

pkg/pq/priority_queue.go
Eithea: It was a problem I hadn't thought of. I learned a good idea that using 'Less' in both ways can replace 'equal'. The type comparableValue which was created because of the '==' operation could also be removed. Thank you!

hackerwins (approved):
LGTM. 👍

PR #337, opened 2022-06-07
Add a document benchmark test for text deletion

Eithea:
Add benchmark tests which assume that large text deletes
at a time in a docs.

Fixes #

I made a function and add 2 test for 1,000 and 10,000 letters. If there is a better size, please tell me.

size 1000 : 
6946508 ns/op
1681464 B/op
58042 allocs/op

size 10000 :
83517332 ns/op
16463037 B/op
580707 allocs/op

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Didn't break anything


test/bench/document_bench_test.go
```diff
@@ -644,4 +676,4 @@ func benchmarkObject(cnt int, b *testing.B) {
 		})
 		assert.NoError(b, err)
 	}
-}
+}
```
hackerwins, line 679: Could you keep the newline at the end of the file?

https://stackoverflow.com/questions/5813311/whats-the-significance-of-the-no-newline-at-end-of-file-log

hackerwins (approved):
Overall it looks good.
I left a simple comment.

test/bench/document_bench_test.go
Eithea: Knowledge has increased. thank you!
I think it disappeared when I moved the function i made. 😅

PR #338, opened 2022-06-08
Implement an update project

DongjinS:
Co-authored-by: Yourim Cha <yourim.cha@navercorp.com>

**What this PR does / why we need it**:
- [ ] Implement an update project API
  - [x] In mongoDB, Implement UpdateProjectInfo  
  - [x] In memoryDB, Implement UpdateProjectInfo
  - [x] Add name field to UpdatableProjectFields
  - [x] Add authWebhookURL field to UpdatableProjectFields
  - [x] Add authWebhookMethod field to UpdatableProjectFields
    - [x] check exceptional cases
    - [x] check duplicated code
  - [x] Add updated_at field to project
  - [ ] Apply to yorkie-house

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Added relevant tests or not required
- [x] Didn't break anything

PR #345, opened 2022-06-23
Introduce validator for updatable_project_fields validation

DongjinS:
Introduce go-playground/validator for updatable_project_fields validation

In this commit, the validator is only applied for name and one
 regular expression is used because go-playground/validator does not provide
perfect tag for project's name.

In addition, AuthWebhookURL of updatable_project_fields also need to apply validator.

- go-playground/validator
  > https://github.com/go-playground/validator
- regular expression for name
  > https://datatracker.ietf.org/doc/html/rfc3986#section-2.3
  > https://gist.github.com/dpk/4757681

- what
   - Introduce go-playground/validator for updatable_project_fields validation
- why
   - Constraints for valid updatable_project_fields(include name) was needed
   - go-playground/validator provide proper validation

Fixes #338 
- This PR is extended from https://github.com/yorkie-team/yorkie/pull/338

I think there could be more reserved name. 
Currently, there are just two reserved names: "new" and "default".

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins (commented):
Thank you for suggesting a library that is easy to express constraints.

Let's find a way to use `go-playground/validator` well.

api/types/updatable_project_fields.go
```diff
@@ -54,5 +75,29 @@ func (i *UpdatableProjectFields) Validate() error {
 			}
 		}
 	}
+
+	v := validator.New()
```
hackerwins, line 79: `Validate` is designed to be thread-safe and used as a singleton. So I think creating an instance every time Validate is called is not fit the design intent.

https://github.com/go-playground/validator/blob/9e2ea4038020b5c7e3802a21cfa4e3afcfdcd276/validator_instance.go#L94-L99

DongjinS: Thanks for review.
Now I make new file `validator.go` and use `Validate` as a package global variable.
- https://github.com/yorkie-team/yorkie/pull/345/files#diff-ad4ef0932407ad66ec522d5674a6995b785ac92d716d6c9ecd146ff870b181e5:~:text=var%20(-,defaultValidator,-%3D%20initValidator()

Please check it and let me know if you have anything in mind.

hackerwins (approved):
Overall it looks good. Thanks for your contribution.

PR #351, opened 2022-06-30
rebuild : applying split tree range isolation

Eithea:

From the stable version, range isolation in splay tree is applied.
By not applying node deletion yet, it is expected that removed
nodes return correct index.

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Didn't break anything


Eithea:
I will pull it again after organizing well.

PR #364, opened 2022-07-20
Fix incorrect index for nodes newly created then concurrently removed

Eithea:

Nodes newly created then concurrently removed in the range to be deleted
had incorrect weight since they were not included in either nodesToKeep
or nodesToDelete. This problem causes incorrect indexing, and this commit
fixes that missing weight updates.

all changes are same with https://github.com/yorkie-team/yorkie-js-sdk/pull/334

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins (approved):
LGTM. 👍

PR #360, opened 2022-07-11
Implement pagination flags to history command

chacha912:
**What this PR does / why we need it**:

- Add `previous_seq`, `size`, `forward` flags to history command
- Apply generics to type `Paging`
- Add Makefile script that generates protoset file to test grpcurl

**Which issue(s) this PR fixes**:

Fixes #352

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

- Users can use the history command 

  ```bash 
  $ bin/yorkie history projectName documentKey
  $ bin/yorkie history projectName documentKey --previous-seq=4 --size=3 --forward
  ```
  image

- Users can use [gRPCurl](https://github.com/fullstorydev/grpcurl) (curl for gRPC servers) 

  ```bash 
  $ make protoset
  
  $ grpcurl -protoset ./yorkie.protoset -plaintext localhost:11103 api.Admin/ListProjects 
  $ grpcurl -protoset ./yorkie.protoset -plaintext -d '{"project_name":"drawing-world", "document_key":"world-docs11"}' localhost:11103 api.Admin/ListChanges
  ```
  image

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Added relevant tests or not required
- [x] Didn't break anything

PR #370, opened 2022-07-25
Add `--backend-snapshot-with-purging-changes` flag

### LINKED ISSUE #288: Add `--backend-snapshot-with-purging-changes` flag
**What happened**:
Hey there! Yorkie seems really really cool - exactly what I've been looking for for a niche project of mine. Very cool to finally find a project that has support for realtime collaborative editing but isn't just the client side with everything else left to create from scratch.  

I've got it running remotely behind envoy and am able to sync docs etc. Trying now to look at some performance and such for the database (using mongodb) and noticed that it doesn't look like garbage collection is happening after a snapshot is taken. Or it's possible I don't fully understand the [GC](https://yorkie.dev/docs/garbage-collection) behavior properly 😓 😄 

I've been using the Text CRDT type to do collaborative text editing, like so:

```
doc.update((root) => {
        if (content) {
            if (!root.content) {
                const text = root.createText("content");
                text.edit(content.start, content.end, content.text);
            } else {
                root.content.edit(content.start, content.end, content.text);
            }
        }
    });
```

After 1000 edits, there are 1000 docs in the `changes` collection. I can see a snapshot being taken like so:

```
INFO	SNAP: '61d5e45922aadcfafba4606f$61d94aa4a3ff7fe9024d9cc9', serverSeq: 1010
INFO	RPC : "/api.Yorkie/PushPull" 653.961929ms
INFO	PUSH: '61d94aab0035805841153d7a' pushes 12 changes into '61d5e45922aadcfafba4606f$61d94aa4a3ff7fe9024d9cc9', rejected 0 changes, serverSeq: 1028 -> 1040, cp: serverSeq=1040, clientSeq=1040
```

However, it still seems like there are the same number of documents in the collection, which is surprising to me. I would've thought that the snapshot would remove the need for the entire history of changes made so far (maybe I'm off on that) or at least that some changes would've been removed as unnecessary. I noticed that there was a recent release that specifically enabled this to happen on snapshots, so wanted to file this issue in case either I'm missing something or it's a real bug. Thank you!

_see above_

**Environment**:
- Operating system: linux (amd64 docker image)
- Browser and version: n/a
- Yorkie version (use `yorkie version`):  0.2.1
- Yorkie JS SDK version: 0.2.0

chromato99:

This commit implements the `--backend-snapshot-with-purging-changes` flag.
If the above flag is used, changes that are not needed because they are recorded in the snapshot are removed to save storage.

This enables the option to delete old changes when creating a snapshot.

Fixes #288

More specifically, the changes that are deleted are before the lowest serverSeq recorded in the syncedseqs collection.
So it is based on the most lastly updated value in syncedseqs, not the changes just before the snapshot. 
The reason is to avoid an error when there is a process of finding change when working UpdateAndFindMinSyncedTicket function.

Users can now use the '--backend-snapshot-with-purging-changes' flag when run yorkie server. This can help save db storage.

```release-note
yorkie server --backend-snapshot-with-purging-changes
```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins:
@chromato99 Thanks for your contribution.

Does `history` command work after purging changes?
https://github.com/yorkie-team/yorkie/pull/360

chromato99:
Ah, I forgot the history command and didn't test it..

When I rebased with the newly added `document ls` command commit and test it now, it seems that the ListChangeSummaries is assuming that there are all changes, so there is an error.

I think it can fix, when the `--backend-snapshot-with-purging-changes` flag exists, just print the existing changes not from seq 1.

I'll edit the code and commit it.

admin/client.go
```diff
@@ -201,13 +201,14 @@ func (c *Client) ListChangeSummaries(
...
-		IsForward: isForward,
-	}, lastSeq)
-	seq := from - 1
+	// lastSeq := changes[len(changes)-1].ID().ServerSeq()
+	// from, _ := types.GetChangesRange(types.Paging[uint64]{
+	// 	Offset:    previousSeq,
+	// 	PageSize:  int(pageSize),
+	// 	IsForward: isForward,
+	// }, lastSeq)
+	// seq := from - 1
```
chromato99: The above ListChanges function is already calling the GetChangesRange function, so I omitted it.

chromato99:
Screenshot from 2022-07-25 22-06-26
Screenshot from 2022-07-25 22-06-54

As shown in the screenshot above, when the `--backend-snapshot-with-purging-changes` flag is used, only changes since the latest snapshot are displayed.

However, I think that too many functions have been added for exception handling. If you have any good comments, please feel free to tell:)

hackerwins:
Thanks for the test.
The history will be reduced after introducing `revision`s that point to meaningful changes in the future.

Could you fix the conflict by rebasing or merging the main branch?

chromato99:
Yes, I resolved the conflict!

chromato99:
There was a problem with the auto merged part, so I rebase it.

admin/client.go
```diff
@@ -201,13 +201,7 @@ func (c *Client) ListChangeSummaries(
...
 	}
 
-	lastSeq := changes[len(changes)-1].ID().ServerSeq()
-	from, _ := types.GetChangesRange(types.Paging[uint64]{
-		Offset:    previousSeq,
-		PageSize:  int(pageSize),
-		IsForward: isForward,
-	}, lastSeq)
-	seq := from - 1
+	seq := changes[0].ServerSeq()
```
hackerwins, line 204: Is there any problem with changing this code?
CC) @chacha912

server/backend/database/memory/database.go
```diff
@@ -547,6 +547,44 @@ func (d *DB) CreateChangeInfos(
...
+	defer txn.Abort()
+
+	// 01. Find the smallest serverSeq recorded in colSyncedSeqs.
+	// This is to avoid the error that the change could not be found when doing UpdateAndFindMinSyncedTicket.
+	it, err := txn.Get(tblSyncedSeqs, "id")
+	if err != nil {
+		return err
+	}
+
+	var minSyncedServerSeq uint64 = 18446744073709551615
```
hackerwins, line 565:
```suggestion
	var minSyncedServerSeq uint64 = math.MaxUint64
```

hackerwins (changes requested):
Thanks for your contribution.

How about adding some test code for the purging logic?
https://github.com/yorkie-team/yorkie/blob/main/test/integration/history_test.go#L33

We also need to fix the lint error.
https://github.com/yorkie-team/yorkie/runs/7512228820?check_suite_focus=true

server/backend/database/memory/database.go
chromato99: That looks more clean!

chromato99:
I think I overlooked the test. Thanks for your comments.
I will work on the test part more.

admin/client.go
chromato99: To explain why I've implemented this, when there's a `--backend-snapshot-with-purging-changes` flag, the changes of the front are deleted, so I want to create a history based on the changes that the ListChanges succeeded in loading from db .
I thought there would be no problem because the change is imported from the db based on GetChangesRange in ListChanges.

I'd appreciate it if you could point out something that I missed.

admin/client.go
```diff
@@ -201,7 +201,7 @@ func (c *Client) ListChangeSummaries(
 		return nil, err
 	}
 
-	seq := changes[0].ServerSeq()
+	seq := changes[0].ServerSeq() - 1
```
chromato99, line 209: > Is there any problem with changing this code? CC) @chacha912



> To explain why I've implemented like this, when there's a `--backend-snapshot-with-purging-changes` flag, the changes of the front are deleted, so I want to create a history based on the changes that the ListChanges succeeded in loading from db . I thought there would be no problem because the change is imported from the db based on GetChangesRange in ListChanges.
> 
> I'd appreciate it if you point out something that I missed.

I updated it little bit, check here please!
CC) @chacha912

chromato99:
I added simple handling for hisotry_test!

admin/client.go
chacha912: I like it. 
It didn't occur to me that I can get sequence by `changes[0].ServerSeq()`. 
Thanks :)

admin/client.go
chacha912: I like it. 
It didn't occur to me that I can get start-sequence by `changes[0].ServerSeq()`.  Thanks :)

server/admin/server.go
```diff
@@ -266,6 +266,23 @@ func (s *Server) GetSnapshotMeta(
...
 	}
 
+	// When the `--backend-snapshot-with-purging-changes` flag exists, there is no changes before the latest snapshot,
+	// so set the latest snapshot to req.serverSeq.
+	if s.backend.Config.SnapshotWithPurgingChanges {
+		docInfo, err := s.backend.DB.FindDocInfoByKey(ctx, project.ID, key.Key(req.DocumentKey))
+		if err != nil {
+			return nil, err
+		}
+		latestSnapshot, err := s.backend.DB.FindLatestSnapshotInfo(ctx, docInfo.ID)
```
hackerwins, line 276: Can we replace this code with the code below?
```suggestion
		latestSnapshot, err := s.backend.DB.FindClosestSnapshotInfo(ctx, docInfo.ID, uint64(math.MaxUint64))
```

admin/client.go
```diff
@@ -218,13 +217,22 @@ func (c *Client) ListChangeSummaries(
...
+	)
 
 	if err != nil {
 		return nil, err
 	}
 	var summaries []*types.ChangeSummary
 	for _, c := range changes {
+		if c.ServerSeq() <= newDoc.Checkpoint().ServerSeq {
+			continue
+		}
```
hackerwins, line 234: May I know what situation this code was added to handle?

test/integration/history_test.go
```diff
@@ -67,8 +68,14 @@ func TestHistory(t *testing.T) {
 
 		changes, err := adminCli.ListChangeSummaries(ctx, "default", d1.Key(), 0, 0, true)
 		assert.NoError(t, err)
-		assert.Len(t, changes, 3)
 
+		if helper.SnapshotWithPurgingChanges {
+			// There is no change when SnapshotInterval is not set.
+			assert.Len(t, changes, 0)
+			return
+		}
```
hackerwins, line 76: Is this code ever actually executed?

hackerwins (commented):
Thanks for the correction. 👍 

I left comments and questions.

server/admin/server.go
chromato99: I'm missing a way to reuse an existing function. Thanks for your comment.
I'll update and commit it!

admin/client.go
chromato99: Not required in normal situations, but required if the '--backend-snapshot-with-purging-changes' flag is used!

In the current implementation, the changes that are deleted are before the lowest serverSeq recorded in the syncedseqs collection.
So it is based on the most lastly updated value in syncedseqs, not the changes just before the snapshot.

In this case, changes that are older than the latest snapshot may remain. I added it to skip remained changes.

test/integration/history_test.go
chromato99: I added this in case the developer sets SnapshotWithPurgingChanges in helper.go to true and run test.
But on second thought, I think it would be better to add a separate test case for test automation..!

server/admin/server.go
chromato99: I got the error `cannot transform type primitive.M to a BSON Document: 18446744073709551615 overflows int64` so I'll update it to use MaxInt64 instead.

hackerwins: Thanks for letting me know. 👍

We need to define `MaxServerSeq` with `math.MaxInt64` and replace all types for `ServerSeq` with int64 in the system.

https://github.com/yorkie-team/yorkie/blob/main/pkg/document/change/checkpoint.go#L36

chromato99: If I change this now, almost all uint64s will be replaced with int64s for consistency, but I think I'll have to edit a lot of code.
Is it ok if I make batch edits in this pr?

hackerwins: It would be better to post a separate issue and handle this in another PR.

chromato99: I'll record it as a new issue!

admin/client.go
```diff
@@ -218,13 +217,22 @@ func (c *Client) ListChangeSummaries(
 		return nil, err
 	}
 
-	newDoc, err := document.NewInternalDocumentFromSnapshot(key, seq, snapshotMeta.Lamport, snapshotMeta.Snapshot)
+	newDoc, err := document.NewInternalDocumentFromSnapshot(
```
dc7303, line 220: In the `history_test_with_purging_changes` test, intermittently the sequence of newDoc is zero. If you look at the snapshotMeta information, the lamport and seq information are set to the initial values.

dc7303 (changes requested):
Thank you for your great work! 👍 
I leave a few comments about the work.

### In the `history_test_with_purging_changes` test, intermittently the sequence of `newDoc`(see code comments) is zero.
Intermittently, failure is occurring as shown in the bash below.
```bash
snapshotMeta :: snapshot:"\n\020\022\016\032\014\000\000\000\000\000\000\000\000\000\000\000\000" lamport:1 
    history_test.go:144: 
                Error Trace:    history_test.go:144
                Error:          "[%!s(*types.ChangeSummary=&{{3 3 3 0xc00028ae40} buy bread {"todos":["buy coffee","buy bread"]}}) %!s(*types.ChangeSummary=&{{2 2 2 0xc00028ad80} buy coffee {"todos":["buy coffee"]}}) %!s(*types.ChangeSummary=&{{1 1 1 0xc00028ace0} create todos {"todos":[]}})]" should have 0 item(s), but has 3
                Test:           TestHistory/history_test_with_purging_changes
encodedDocID ObjectID("62e3909c97bc128e9c393a59")
--- FAIL: TestHistory (0.21s)
    --- PASS: TestHistory/history_test (0.07s)
    --- FAIL: TestHistory/history_test_with_purging_changes (0.13s)
FAIL
```

### It seems that you are not testing the actual changes being deleted.
While testing to analyze the intermittent test errors mentioned earlier, I found that the tests were not actually deleting Changes. Nevertheless, the reason the test succeeds seems to be getting the changes filtered by 'ListChangeSummaries' rather than actually being deleted.

test command: `go test -tags integration -race ./test/integration/main_test.go ./test/integration/history_test.go `

line where the logic stops: https://github.com/yorkie-team/yorkie/pull/370/files#diff-b4e3df8e62ae60e44935bd2f072a98591f14aace084ef4214b0a2cad6a0d6a68R661-R663


### In conclusion
Intermittent issues require further analysis. This is because the integration test environment can cause unexpected problems (eg concurrency problems). I will also analyze this issue.

But first, it seems like a good idea to fix the test to see if you can delete old Changes.

admin/client.go
chromato99: It seems that I have found a suspicious situation.

At the server-side implementation, when the yorkie server is synced, the creation or reflection of changes is synchronous, but the snapshot creation seems to be a goroutine. 
If so, it seems that a history request is made before the snapshot creation is completed, and the snapshotMeta may be created based on ServerSeq 0.
In that case, newDoc will be created from 0 since there is no snapshot yet.

dc7303: I did the same analysis. You used `time.Sleep` in the current commit, but I don't think this completely solves the cause of the test failure.

How about testing using monkey-patch? 
example:
```go
	t.Run("history test with purging changes", func(t *testing.T) {
	        ...
		var b *background.Background
		// "bou.ke/monkey"
		monkey.PatchInstanceMethod(
			reflect.TypeOf(b),
			"AttachGoroutine",
			func(_ *background.Background, f func(c context.Context)) {
				f(context.Background())
			},
		)
		...
	}
```

But the problem is that bou.ke/monkey has an issue in an ARM machine.
reference: https://github.com/jenkins-x/jx/issues/2081

In my case, I checked the improvement by running the test with the command below referring to the reference above.

`GOARCH=amd64 go test -tags integration -gcflags=all=-l ./test/integration/main_test.go ./test/integration/history_test.go`

cc. @hackerwins

chromato99: Thanks for the nice comment .
It seems like a very good way to explicitly handle it!!

https://github.com/yorkie-team/yorkie/blob/4b333b949e03ae3f60fdbeb406a45c0c1a415541/test/integration/snapshot_test.go#L59-L63

However, other tests also seem to use `time.Sleep`. It seems good to apply monkey-patch consistently for these tests as well. 
Would it be better to treat this as a seperate issue?

cmd/yorkie/server.go
```diff
@@ -334,6 +334,12 @@ func init() {
 		server.DefaultSnapshotInterval,
 		"Interval of changes to create a snapshot.",
 	)
+	cmd.Flags().BoolVar(
+		&conf.Backend.SnapshotWithPurgingChanges,
+		"backend-snapshot-with-purging-changes",
+		server.DefaultSnapshotWithPurgingChanges,
+		"Whether to delete previous changes when the snapshot is created. ",
```
hackerwins, line 341:
```suggestion
		"Whether to delete previous changes when the snapshot is created.",
```

server/backend/database/memory/database.go
```diff
@@ -666,6 +667,47 @@ func (d *DB) CreateChangeInfos(
...
+	defer txn.Abort()
+
+	// 01. Find the smallest serverSeq recorded in colSyncedSeqs.
+	// This is to avoid the error that the change could not be found when doing UpdateAndFindMinSyncedTicket.
+	it, err := txn.Get(tblSyncedSeqs, "id")
+	if err != nil {
+		return err
+	}
+
+	var minSyncedServerSeq int64 = math.MaxInt64
```
hackerwins, line 685:
```suggestion
	minSyncedServerSeq := math.MaxInt64
```

https://github.com/uber-go/guide/blob/master/style.md#local-variable-declarations

server/backend/database/mongo/client.go
```diff
@@ -761,6 +761,57 @@ func (c *Client) CreateChangeInfos(
...
+
+	// 01. Find the smallest serverSeq recorded in colSyncedSeqs.
+	// This is to avoid the error that the change could not be found when doing UpdateAndFindMinSyncedTicket.
+	result := c.collection(colSyncedSeqs).FindOne(
+		ctx,
+		bson.M{
+			"doc_id": encodedDocID,
+		},
+		options.FindOne().SetSort(bson.M{"server_seq": 1}),
+	)
```
hackerwins, line 782: We need to compare the order of changes with `lamport` and `actor_id`(logical clock), not `server_seq`, which is a physical storage order in the server.

How about extracting the below codes with a function and use it together?
https://github.com/yorkie-team/yorkie/blob/4b333b949e03ae3f60fdbeb406a45c0c1a415541/server/backend/database/mongo/client.go#L904-L910

In fact, We're using Lamport timestamp, so we don't know whether the two changes are causal relations or concurrent relations. It would be better to deal with the problem about the relation of change in another issue later.

hackerwins (changes requested):
Overall, the code looks good.
But I think a few logic needs to be fixed.

server/backend/database/memory/database.go
chromato99: Thank you so much for commenting.
I should have checked the guide more carefully, but I think I missed it..
We will update it.

chromato99: https://github.com/chromato99/yorkie/blob/7f04224d979743882d885bb42dce4447af5d3fef/server/backend/database/memory/database.go#L685

Go recognized MaxInt64 as an int instead of an int64, so I got an error in the comparison statement below.
Would this method be okay?

admin/client.go
hackerwins: It would be good to deal with other issues.

hackerwins: It would be good to consider the following:

1. Call `unpatch` explicitly by returning guard.: https://github.com/bouk/monkey#using-monkey
2. Add `amd64` tag to run only on `amd64` machine: https://github.com/yorkie-team/yorkie/issues/388

server/backend/database/memory/database.go
```diff
@@ -667,6 +668,47 @@ func (d *DB) CreateChangeInfos(
 	return nil
 }
 
+// DeleteOldChangeInfos delete old change infos before checpoint to save storage
+func (d *DB) DeleteOldChangeInfos(
+	ctx context.Context,
+	docInfo *database.DocInfo,
```
hackerwins, line 674: We call this function with `docInfo`, but inside the function, we only use the `ID` of it.
How about only passing the ID to further reveal the intent of the function?

chromato99: I updated it!

admin/client.go
chromato99: Thank you for comment!
I updated the code.
Can you check if the tag is working properly in the ARM environment?

server/backend/database/memory/database.go
```diff
@@ -786,6 +828,35 @@ func (d *DB) FindClosestSnapshotInfo(
 	return snapshotInfo, nil
 }
 
+// FindMinSyncedSeqInfo finds the minimum synced sequence info.
+func (d DB) FindMinSyncedSeqInfo(
```
hackerwins, line 832:
```suggestion
func (d *DB) FindMinSyncedSeqInfo(
```

server/config.sample.yml
```diff
@@ -50,6 +50,9 @@ Backend:
   # SnapshotInterval is the number of changes to create a snapshot.
   SnapshotInterval: 1000
 
+  # SnapshotWithPurgingChanges is whether to delete previous changes when the snapshot is created.
+  SnapshotPurgingChanges: false
```
hackerwins, line 54:
```suggestion
  SnapshotWithPurgingChanges: false
```

test/integration/snapshot_test.go
```diff
@@ -21,18 +21,42 @@ package integration
...
+		reflect.TypeOf(b),
+		"AttachGoroutine",
+		func(_ *background.Background, f func(c context.Context)) {
+			f(context.Background())
+		},
+	)
+	defer monkey.UnpatchInstanceMethod(
+		reflect.TypeOf(b),
+		"AttachGoroutine",
+	)
```
hackerwins, line 58: How about using unpatch directly?

```suggestion
	guard := monkey.PatchInstanceMethod(
		reflect.TypeOf(b),
		"AttachGoroutine",
		func(_ *background.Background, f func(c context.Context)) {
			f(context.Background())
		},
	)
	defer func() {
		guard.Unpatch()
	}()
```

test/integration/history_test.go
```diff
@@ -20,17 +20,40 @@ package integration
...
+		reflect.TypeOf(b),
+		"AttachGoroutine",
+		func(_ *background.Background, f func(c context.Context)) {
+			f(context.Background())
+		},
+	)
+	defer monkey.UnpatchInstanceMethod(
+		reflect.TypeOf(b),
+		"AttachGoroutine",
+	)
```
hackerwins, line 55:
```suggestion
	guard := monkey.PatchInstanceMethod(
		reflect.TypeOf(b),
		"AttachGoroutine",
		func(_ *background.Background, f func(c context.Context)) {
			f(context.Background())
		},
	)
	defer func() {
		guard.Unpatch()
	}()
```

hackerwins (changes requested):
Thank you for applying the requests.
It looks like the work is almost done.

server/backend/database/memory/database.go
chromato99: There was a fatal typo...
I'll update it.

server/config.sample.yml
chromato99: Thank you for comment.
I'll update it.

test/integration/snapshot_test.go
hackerwins: Oh, I see. Thanks for the explanation.

test/integration/snapshot_test.go
```diff
@@ -21,18 +21,42 @@ package integration
...
+	"github.com/yorkie-team/yorkie/server"
+	"github.com/yorkie-team/yorkie/server/backend/background"
+	"github.com/yorkie-team/yorkie/server/backend/database/mongo"
+	"github.com/yorkie-team/yorkie/server/logging"
 	"github.com/yorkie-team/yorkie/test/helper"
 )
 
 func TestSnapshot(t *testing.T) {
+	var b *background.Background
+	// "bou.ke/monkey"
```
hackerwins, line 47:
```suggestion
```

test/integration/history_test.go
```diff
@@ -20,17 +20,40 @@ package integration
...
 	"github.com/yorkie-team/yorkie/pkg/document/key"
+	"github.com/yorkie-team/yorkie/server"
+	"github.com/yorkie-team/yorkie/server/backend/background"
+	"github.com/yorkie-team/yorkie/server/logging"
 	"github.com/yorkie-team/yorkie/test/helper"
 )
 
 func TestHistory(t *testing.T) {
+	var b *background.Background
+	// "bou.ke/monkey"
```
hackerwins, line 44:
```suggestion
```

hackerwins (approved):
LGTM. 👍

PR #368, opened 2022-07-24
Introduce`buf`

heka1024:
**What this PR does / why we need it**:
- Introduce [buf](https://buf.build/)
- There is a violation of the [protocol-buffers style guide](https://developers.google.com/protocol-buffers/docs/style#enums). But it's hard to remember all these rules. So, it's better to have linter, like `golangci-lint`.

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:
- [Slack thread](https://dev-yorkie.slack.com/archives/C014BTGCBJS/p1658576462364749?thread_ts=1658560179.994979&cid=C014BTGCBJS)
- [Buf Introduction](https://docs.buf.build/tour/introduction/)
- [Avoiding Common Protobuf’s Pitfalls with Buf](https://earthly.dev/blog/buf-protobuf/)

FYI, `buf lint` says
```
api/admin.proto:18:1:Package name "api" should be suffixed with a correctly formed version, such as "api.v1".
api/admin.proto:23:9:Service name "Admin" should be suffixed with "Service".
api/cluster.proto:18:1:Package name "api" should be suffixed with a correctly formed version, such as "api.v1".
api/cluster.proto:23:9:Service name "Cluster" should be suffixed with "Service".
api/resources.proto:21:1:Package name "api" should be suffixed with a correctly formed version, such as "api.v1".
api/resources.proto:173:9:Oneof name "Body" should be lower_snake_case, such as "body".
api/resources.proto:285:3:Enum value name "NULL" should be prefixed with "VALUE_TYPE_".
api/resources.proto:285:3:Enum zero value name "NULL" should be suffixed with "_UNSPECIFIED".
api/resources.proto:286:3:Enum value name "BOOLEAN" should be prefixed with "VALUE_TYPE_".
api/resources.proto:287:3:Enum value name "INTEGER" should be prefixed with "VALUE_TYPE_".
api/resources.proto:288:3:Enum value name "LONG" should be prefixed with "VALUE_TYPE_".
api/resources.proto:289:3:Enum value name "DOUBLE" should be prefixed with "VALUE_TYPE_".
api/resources.proto:290:3:Enum value name "STRING" should be prefixed with "VALUE_TYPE_".
api/resources.proto:291:3:Enum value name "BYTES" should be prefixed with "VALUE_TYPE_".
api/resources.proto:292:3:Enum value name "DATE" should be prefixed with "VALUE_TYPE_".
[...]

PR #372, opened 2022-07-28
Refactor `proto`

heka1024:
#368, We decide to lint our `proto` files.
Lint our `proto` files as linter suggests, and follow `proto` convention like `karrot` or `uber` does.

Fixes lint errors in #368.

```release-note
- Breaking changes in protocol.
- From now, generated files like `admin.pb.go` are located in `gen/go`.
```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Didn't break anything (Actually, break everything!)


admin/client.go
```diff
@@ -18,12 +18,12 @@ package admin
 
 import (
 	"context"
+	"github.com/yorkie-team/yorkie/gen/go/yorkie/v1"
```
hackerwins, line 21: We organized directories according to the [Go Project Layout](https://github.com/golang-standards/project-layout#api).

How about changing this like the below?
```suggestion
	api "github.com/yorkie-team/yorkie/api/v1"
```

hackerwins (changes requested):

I left a simple question about the package location.

We also need to edit the codes of `make proto` command.
https://github.com/yorkie-team/yorkie/blob/0cca5e25b8b83ad5b389f84735324a016f9691f9/Makefile#L29-L37

PR #375, opened 2022-07-29
Fix the problem local changes were applied twice

Eithea:

When a client receives a remote snapshot, the concurrent local changes
were apllied twice.
client.syncInternal first sends local changes to server, and then
receives remote ChangePack. So the snapshot in given ChangePack
is of the time when the local changes were already applied.

By deleting change.execute in applySnapshot, fix above problem.

https://github.com/yorkie-team/yorkie-js-sdk/pull/356

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins (approved):
LGTM. 👍

PR #379, opened 2022-08-04
Change the data type of data inserted into the db from uint64 to int64

### LINKED ISSUE #373: Change the data type of data inserted into the db from uint64 to int64.
**Description**:

https://github.com/yorkie-team/yorkie/pull/370#discussion_r931726563

It would be better to change uint64 to int64 among data types that are inserted into db.
For example, the data type of ServerSeq defined in `pkg/document/change/checkpoint.go` need to change from uint64 to int64. 

**Why**:

https://pkg.go.dev/go.mongodb.org/mongo-driver/bson#hdr-Native_Go_Types

According to the above document, uint64 is marshaled to int64, when marshalling a D or M to BSON.
Therefore, there is no problem in the case of a number smaller than math.MaxInt64 even now, however changing this can make the data types more consistent.

chromato99:

This PR changes uint64 to int64 among the data types of data inserted into the db.

https://pkg.go.dev/go.mongodb.org/mongo-driver/bson#hdr-Native_Go_Types

According to the above document, uint64 is marshaled to int64, when marshalling a D or M to BSON.
Therefore, there is no problem in the case of a number smaller than math.MaxInt64 even now, however changing this can make the data types more consistent.

Fixes #373

Changes two data's data type, ServerSeq and Lamport, which are uint64 data inserted into the db.
There are other uint64 data, but they don't seem to be inserted into the db, so I haven't modified them.

```release-note
NONE
```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins:
Continue on #381.

PR #380, opened 2022-08-05
Add `[jstype = JS_STRING]` option in admin.proto

chacha912:

> The jstype option determines the JavaScript type used for values of the
> field.  The option is permitted only for 64 bit integral and fixed types
> (int64, uint64, sint64, fixed64, sfixed64).  A field with jstype JS_STRING
> is represented as JavaScript string, which avoids loss of precision that
> can happen when a large value is converted to a floating point JavaScript.

https://github.com/protocolbuffers/protobuf/blob/79c3d07c63e711255e0d566c9b876706e978686a/php/src/Google/Protobuf/Internal/FieldOptions.php#L37-L51

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins (approved):
LGTM. 👍

PR #401, opened 2022-09-03
improve commit and buildDate

Limm-jk:
- improve version cmd's commitVersion and buildTime (use debug/buildInfo (add [this feature](https://tip.golang.org/doc/go1.18#debug/buildinfo) in go 1.18)).
- rm make script depending on os.

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,12 @@ YORKIE_VERSION := 0.2.17
 
 GO_PROJECT = github.com/yorkie-team/yorkie
 
+ifeq ($(OS),Windows_NT)
+    EXECUTABLE = ./bin/yorkie.exe
+else
+    EXECUTABLE = ./bin/yorkie
+endif
+
 # inject the version number into the golang version package using the -X linker flag
 GO_LDFLAGS ?=
 GO_LDFLAGS += -X ${GO_PROJECT}/internal/version.Version=${YORKIE_VERSION}
--- a/Makefile
+++ b/Makefile
@@ -1,23 +1,10 @@
 YORKIE_VERSION := 0.2.17
 
-GIT_COMMIT := $(shell /usr/bin/git rev-parse --short HEAD)
 GO_PROJECT = github.com/yorkie-team/yorkie
-
-ifeq ($(OS),Windows_NT)
-    BUILD_DATE := $(shell echo %date:~6,4%-%date:~0,2%-%date:~3,2%)
-    GO_SRC := $(shell dir /s /b *.go | findstr /v "vendor")
-    EXECUTABLE = ./bin/yorkie.exe
-else
-    BUILD_DATE := $(shell date "+%Y-%m-%d")
-    GO_SRC := $(shell find . -path ./vendor -prune -o -type f -name '*.go' -print)
-    EXECUTABLE = ./bin/yorkie
-endif
 
 # inject the version number into the golang version package using the -X linker flag
 GO_LDFLAGS ?=
-GO_LDFLAGS += -X ${GO_PROJECT}/internal/version.GitCommit=${GIT_COMMIT}
 GO_LDFLAGS += -X ${GO_PROJECT}/internal/version.Version=${YORKIE_VERSION}
-GO_LDFLAGS += -X ${GO_PROJECT}/internal/version.BuildDate=${BUILD_DATE}
 
 default: help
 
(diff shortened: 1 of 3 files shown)


cmd/yorkie/version.go
```diff
@@ -19,26 +19,43 @@ package main
...
 	"runtime"
+	"runtime/debug"
 
 	"github.com/spf13/cobra"
 
 	"github.com/yorkie-team/yorkie/internal/version"
 )
 
+const commitRevisionKey = "vcs.revision"
+const buildTimeKey = "vcs.time"
```
hackerwins, line 30: `vcs.time` might be commit time, not build time. I think it would be better to keep the existing build time. What do you think? 🤔

https://cs.opensource.google/go/go/+/refs/tags/go1.18.5:src/cmd/go/internal/vcs/vcs.go;l=60

cmd/yorkie/version.go
```diff
@@ -19,26 +19,43 @@ package main
...
+		if info, ok := debug.ReadBuildInfo(); ok {
+			for _, setting := range info.Settings {
+				if setting.Key == key {
+					return setting.Value
+				}
+			}
+		}
+		return ""
+	}()
+}
```
hackerwins, line 57: Is there any reason to wrap the logic in an anonymous function?

cmd/yorkie/version.go
```diff
@@ -19,26 +19,43 @@ package main
...
+			fmt.Printf("Commit: %s\n", getBuildInfoByKey(commitRevisionKey))
 			fmt.Printf("Go: %s\n", runtime.Version())
-			fmt.Printf("Build date: %s\n", version.BuildDate)
+			fmt.Printf("Build date: %s\n", getBuildInfoByKey(buildTimeKey))
 			return nil
 		},
 	}
 }
 
+func getBuildInfoByKey(key string) string {
```
hackerwins, line 46: It's like a function that returns a value from `BuildInfo.Settings` by key. So it seems more accurate to name it `getSettingValueByKey`. But I think `settingValueByKey` is appropriate because Go doesn't name getters like `GetXXX`.

https://go.dev/doc/effective_go#Getters

cmd/yorkie/version.go
```diff
@@ -19,26 +19,43 @@ package main
...
+const buildTimeKey = "vcs.time"
+
 func newVersionCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:   "version",
 		Short: "Print the version number of Yorkie",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			fmt.Printf("Yorkie: %s\n", version.Version)
-			fmt.Printf("Commit: %s\n", version.GitCommit)
+			fmt.Printf("Commit: %s\n", getBuildInfoByKey(commitRevisionKey))
```
hackerwins, line 38: Previously, `GitCommit` is a Short Hash of the revision: `git rev-parse --short HEAD`. How about assigning a short hash here?

Users can still check the full value of the revision using `go version`.
```
$ go version -m ./bin/yorkie
...
	build	vcs.revision=6096695f50be41580f86ea7c017f52d85b400019
	build	vcs.time=2022-08-29T10:42:58Z
	build	vcs.modified=true
```

hackerwins (changes requested):
Thanks for your contribution. I didn't know that Go 1.18 provides version control information. I learned from this PR.

I left a few comments.

Makefile
```diff
@@ -1,23 +1,16 @@
 YORKIE_VERSION := 0.2.17
 
-GIT_COMMIT := $(shell /usr/bin/git rev-parse --short HEAD)
 GO_PROJECT = github.com/yorkie-team/yorkie
 
 ifeq ($(OS),Windows_NT)
-    BUILD_DATE := $(shell echo %date:~6,4%-%date:~0,2%-%date:~3,2%)
-    GO_SRC := $(shell dir /s /b *.go | findstr /v "vendor")
```
hackerwins, line 8: Is it okay to remove `GO_SRC`? (We are using the value in `make fmt`)

cmd/yorkie/version.go
Limm-jk: It may not fit as much as the time of vcs(commit). However, I thought it would be consistent in the main-based release environment.
This information is important in most deployment environments.

hackerwins:
This continues at #419.

PR #427, opened 2022-10-10
Remove use of bou.ke/monkey library

### LINKED ISSUE #416: Golang monkey patching library utilized against terms of license
The Golang monkey patching library, https://github.com/bouk/monkey, is being utilized directly against [its license](https://github.com/bouk/monkey/blob/master/LICENSE.md):

> Copyright Bouke van der Bijl
> 
> I do not give anyone permissions to use this tool for any purpose. Don't use it.
> 
> I’m not interested in changing this license. Please don’t ask.

It is also archived and not maintained.

chromato99:

Remove use of bou.ke/monkey library.

Recently there was an issue pointing to the license of the monkey patch library.
I looked for several ways to replace it, but other monkey patch libraries have the same problem that there are too few users and can not used in the arm architecture.

So, I thought about changing it in a way that avoids the need for a monkey patch, and I tried it. 
But, this method may be considered unclean.
I'd appreciate it if you could comment on what I've been working on

<b>Edit: </b>
I changed it to use the undefinedlabs/go-mpatch library instead of the above method.
Same problem not working on ARM architecture, but I was able to completely replace the bou.ke/monkey library.

Fixes #416 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/yorkie-team/yorkie
 go 1.19
 
 require (
-	bou.ke/monkey v1.0.2
 	github.com/go-playground/locales v0.14.0
 	github.com/go-playground/universal-translator v0.18.0
 	github.com/go-playground/validator/v10 v10.11.1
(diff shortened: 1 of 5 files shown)


hackerwins:
Thanks for your PR.

However, I am concerned because the structure of the test code seems to add complexity to the original code. 🤔

chromato99:
Thanks for your comment. 

As you mentioned, this method seems to increase the complexity of the original code for testing. I'll have to work on using other monkey patch libraries.

chromato99:
I changed it to use the undefinedlabs/go-mpatch library instead of the above method.
Same problem not working on ARM architecture, but I was able to completely replace the bou.ke/monkey library.

hackerwins (approved):

blurfx (approved):
LGTM 👍

PR #430, opened 2022-10-24
feat: replace etcd image source from bitnami to quay

blurfx:
To resolve the issue of etcd not running properly in the arm64 architecture.

@dc7303 discovered and raised this issue before(etcd-io/etcd#14321), but the problem remains unresolved. For now, When we run `bitnami/etcd` image on arm64 environment, the first time the image run always fails and then we must restart image ourselves.

There was a [comment](https://github.com/etcd-io/etcd/issues/14321#issuecomment-1277240307) on the issue that said using the `quay.io/coreos/etcd` image would solve the problem, and it really works on my arm64 environment without any problems. So I made this pr because I thought it would be better to use the `quay.io/coreos/etcd` image rather than `bitnami/etcd`.

no issues related.

```release-note
NONE
```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/docker/docker-compose-full.yml
+++ b/docker/docker-compose-full.yml
@@ -18,11 +18,18 @@ services:
     depends_on:
       - prometheus
   etcd:
-    image: bitnami/etcd:latest
+    image: quay.io/coreos/etcd:latest
     container_name: etcd
-    environment:
-      - ALLOW_NONE_AUTHENTICATION=yes
-      - ETCD_ADVERTISE_CLIENT_URLS=http://etcd:2379
+    entrypoint: /usr/local/bin/etcd
+    command:
+      - '--name=etcd'
+      - '--initial-advertise-peer-urls=http://etcd:2379,http://etcd:2380'
+      - '--advertise-client-urls=http://etcd:2379,http://etcd:2380'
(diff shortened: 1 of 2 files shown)


Krapi0314:
I had problem with this etcd `m1/arm64` issue, and I have solved with your solution, thank you!

BTW, I'm curious about your etcd configuration flag args, that `http://etcd:2379,http://etcd:2380` value is repeating in several flags. I thought port 2379 was for client requests and 2380 for peer communication, therefore some values are unnecessary.

Maybe configuration should look like this?
```yml
  etcd:
    image: quay.io/coreos/etcd:latest
    container_name: etcd
    entrypoint: /usr/local/bin/etcd
    command:
      - "--name=etcd"
      - "--initial-advertise-peer-urls=http://etcd:2380"
      - "--advertise-client-urls=http://etcd:2379"
      - "--listen-client-urls=http://0.0.0.0:2379"
      - "--listen-peer-urls=http://0.0.0.0:2380"
      - "--initial-cluster=etcd=http://etcd:2380"
      - "--initial-cluster-token=etcd"
      - "--initial-cluster-state=new"
```

It's just a small change :)

hackerwins (approved):

PR #454, opened 2023-01-27
Add ClientDeactivateThreshold in Project

### LINKED ISSUE #444: Regarding `AttachDocument()`'s performance with Housekeeping and Garbage Collection
**Description**:

There is a `AttachDocument()` API performance issue on current [react-tldraw example](https://yorkie.dev/examples/tldraw).
Attaching document is taking almost **14 seconds** to complete, and this is severally harming Yorkie's user experience.

I have investigated this issue and I think I have found the root cause of this problem.
I will explain **why** this is happening, and then mention some **suggestions** to solve this issue.

**Why**:

There were two main causes of this problem:
1. Housekeeping's long period of `housekeeping-deactivate-threshold`
2. Absence of `detachDocument()`or `deactivateClient()` for several reasons

I will explain those two causes with react-tldraw example mentioned above.
It's a long story, so you can scroll down and just see final timeline for this incident.

in my react-tldraw example, I called two functions `client.detach()` and `client.deactivate()` in `handleDisconnect()` on window close to successfully close Yorkie connections.
```typescript
function handleDisconnect() {
   if (client === undefined || doc === undefined) return;

   client.detach(doc);
   client.deactivate();
 }

...

window.addEventListener('beforeunload', handleDisconnect);

```
However `handleDisconnect()` function was not invoked on window close. So `detachDocument()` or `deactivateClient()` were failed. I will call this failed client, a `dead` client.

But still this incident was in control because Yorkie server's Housekeeping service will deactivate clients that have not been used for a long time.
But, current Yorkie server's default configuration value of deactivate threshold `DefaultHousekeepingDeactivateThreshold` was `7 * 24 * time.Hour` = `7 days` (I assume that current Yorkie server on Yorkie SaaS is running on this value). 
```go
DefaultHousekeepingDeactivateThreshold = 7 * 24 * time.Hour
```
Therefore, housekeeping(deactivation) of client on react-tldraw did not happen. This made Yorkie server to keep storing `dead` client's `syncedseqs` because `updateSyncedSeq()` was not performed. 
```go
// UpdateSyncedSeq updates the syncedSeq of the given client.
func(c * Client) UpdateSyncedSeq(
...

if!isAttached {
    if _, err = c.collection(colSyncedSeqs).DeleteOne(ctx, bson.M {
        "doc_id": encodedDocID,
        "client_id": encodedClientID,
    }, options.Delete());
    err != nil {
        logging.From(ctx).Error(err)
        return fmt.Errorf("delete synced seq: %w", err)
    }
    return nil
}
```
After that, in every `PushPull()`, `minSyncedTicket` in `UpdateAndFindMinSyncedTicket()` was not updated because of this `dead` client's `syncedseqs`, which will always be the smallest value. To keep in mind, `syncedseqs` is used to perform Garbage Collection on document. For more information, check https://yorkie.dev/docs/internals#garbage-collection.
```go
// 04. update and find min synced ticket for garbage collection.
// NOTE(hackerwins): Since the client could not receive the response, the
// requested seq(reqPack) is stored instead of the response seq(resPack).
minSyncedTicket, err: = be.DB.UpdateAndFindMinSyncedTicket(
...
```

Therefore, GC was not performed even several `changes` were made in the document because `minSyncedTicket` was always the same.
This made `document`/`snapshot` get bigger and bigger as more changes are made in the document.
```go
// ApplyChangePack applies the given change pack into this document.
func (d *Document) ApplyChangePack(pack *change.Pack) error {
...

// 04. Do Garbage collection.
d.GarbageCollect(pack.MinSyncedTicket)
```
So, response snapshot of `AttachDocument()` got huge (in react-tldraw exmaple's case, `10.9MB`), and this delayed network transfer but also slowed down browser's resource loading (in react-tldraw exmaple's case, `2.05 s` network transfer + `7.12 s` resource loading). This situation is especially common in whiteboard because tons of changes(shape change, coordinate move, etc) are made in collaborative whiteboard.

**To conclude:**
1. `detachDocument()` and `deactivateDocument()` were not performed on tldraw close.
2. Yorkie server's `DefaultHousekeepingDeactivateThreshold` was 7 days, so housekeeping(deactivation) on dead client was not performed.
3. `syncedseqs` of dead client was preserved because `updateSyncedSeq()` was not performed.
4. `minSyncedTicket` in `UpdateAndFindMinSyncedTicket()` were not changed.
5. Garbage Collection on further document changes was not performed.
6. Document -> Snapshot -> response of `AttachDocument()` got huge.
7. Huge response delayed network transfer and slowed down browser's resource loading.

**Suggestions**:

Deactivating(Housekeeping) dead client is very crucial because just one dead client will block `minSyncedTicket`'s update, and the whole document will grow due to GC failure. And this will effect other peer's attach performance. 

I think there are several ways to solve this problem.
- Server side: Make Yorkie server's **housekeeping deactivate threshold shorter** (setting it too short will make client deactivate too quickly, so adjustments should be made).
- Client side: **Ensure client to perform detach/deactivate**. ex: `navigator.sendBaecon(deactivate)` in web browser.

But eventually, server is responsible for this, because client can be unexpectedly terminated due to several reasons (web browser force close, or device crash/blue screen).

I noticed that server can detect `WatchDocuments()` stream close on client's dead state. Maybe we can perform some actions when  `WatchDocuments()` stream is closed. I will keep looking for the solution to solve this problem.

**P.S**:

- There may be other reasons of `attachDocument()` performance issue on `yorkie-js-sdk`: https://github.com/yorkie-team/yorkie-js-sdk/issues/395, so after we fix this issue, I should look for this issue too.
- This issue will explain another cause of https://github.com/yorkie-team/yorkie/issues/391 issue.

krapie:

Add `ClientDeactivateThreshold` in `Project` to set client’s deactivation threshold in individual projects for housekeeping.

This will subtitute preview `DefaultHousekeepingDeactivateThreshold` in global  housekeeping configuration.

Current housekeeping mechanism deactivates client with provided `DefaultHousekeepingDeactivateThreshold` global configuration. This deactivates all clients without considering each client’s workloads.

In Yorkie, `Project` repesents specific workloads, and each `Project` has different **configurations** and **workloads**. Therefore, it will be more efficient to set deactivation conditions in `Project` rather then global housekeeping deactivation configuration.

For more information & context, follow this issue: #444

Related #444

Below is the list to be considered before merge.

1. Both implementation of memory and mongo of `FindDeactivateCandidates()` have changed to fetch all projects first, and then fetch clients by projects using `get()` and `find()`. There is limit(`candidatesLimit`) on clients to prevent overloading on memory, but pagination/limit on projects may also be needed. Then, what is a best pagination/limit strategy for projects on housekeeping? Any suggestions will be helpful. (Or, this may not be needed because we can shard project collections so that server can able to load all projects with no problem?)

```
Users can set client's deactivation threshold in each project using Yorkie dashboard (Yorkie admin CLI)
```

**Additional documentation**:

```

```

**Checklist**:

- [x]  Added relevant tests or not required
- [x]  Didn't break anything


hackerwins:
Thanks for your contribution.

> 1. Previous `DefaultHousekeepingDeactivateThreshold` in global configuration is removed. Do we need to keep this option alive or not?

It would be ok to remove it.

> 2. MongoDB’s aggregation pipeline is used to avoid memory consumptions on server during FindDeactivateCandidates(), is this over engineered?

We need to check that the query(aggregation) is not putting too much load on the DB. In a production environment, the number of projects may be small, but the number of clients may be large. It would be better to process the deactivation by project and check if the index is properly set.

> 3. Follow above, there are mismatch in go’s time.duration (microsecond unit) and mongoDB’s time (microsecond unit). I set time unit to microsecond, but the conversation is quite cumbersome, and this made explicit type convertion between mongoDB and go. Now ConvertBsonMillisceondToDurationMicrosecond() is needed in every projectInfo fetch. (Also ClientDeactivateThreshold in UpdatableProjectFields should be in microseconds…)
4. UpdatableProjectFields has changed to get ClientDeactivateThreshold value for Project ClientDeactivateThreshold update. Therefore code to update project ClientDeactivateThreshold should be added, and protobuf files should be updated in Yorkie dashboard.

Because We tried to use aggregation, I think there was a restriction on the type of the field. It would be good to change the structure to run the deactivation on a per-project basis as mentioned above, and change the type of the field to the user's input format(probably a string).

> 5. FindDeactivateCandidates() implementation of memory has changed to get all data get(), and iterate all data continue instead of reverseLowerBound with global offset. This should be optimized.

Yes, we need to apply proper pagination when fetching candidate clients from DB.

After this work, I think we can add the deactivation setting to the Dashboard and CLI.

Krapi0314:
Thank you for your feedbacks.

> It would be better to process the deactivation by project and check if the index is properly set

Do you mean iterating project with `FindDeactivateCandidatesInProject()` on housekeeping? (indexing `colClients` with project, and performing `FindDeactivateCandidatesInProject()`, which will use same algorithm(query) as previous `FindDeactivateCandidates()`)

If my understanding is correct, I think your suggestion is much better then current implementation. This will definitely reduce cumbersome code to perform deactivateThreshold on clients, and keep previous query to get deactivate candidates!

hackerwins:
> Both implementation of memory and mongo of FindDeactivateCandidates() have changed to fetch all projects first, and then fetch clients by projects using get() and find(). There is limit(candidatesLimit) on clients to prevent overloading on memory, but pagination/limit on projects may also be needed. Then, what is a best pagination/limit strategy for projects on housekeeping? Any suggestions will be helpful. (Or, this may not be needed because we can shard project collections so that server can able to load all projects with no problem?)

The number of projects is very small compared to the number of clients, and there are not many Yorkie users and projects yet. Therefore, it seems that there is no need to introduce pagination to the project now. How about leaving a TODO comment about the pagination of projects for now?

server/backend/database/memory/database.go
```diff
@@ -559,6 +584,29 @@ func (d *DB) FindDeactivateCandidates(
...
+	candidatesLimit int,
+) ([]*database.ClientInfo, error) {
+	projects, err := d.listAllProjectInfos(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var candidates []*database.ClientInfo
+	for _, project := range projects {
+		infos, err := d.findDeactivateCandidatesPerProject(ctx, project, candidatesLimit)
```
hackerwins, line 599: The meaning of `candidatesLimit` has changed to limit per project. How about changing the variable name to `candidatesLimitPerProject`?

server/backend/database/memory/housekeeping_test.go
```diff
@@ -28,38 +28,38 @@ import (
...
-	projectID := database.DefaultProjectID
 
 	t.Run("housekeeping test", func(t *testing.T) {
 		ctx := context.Background()
 
-		yesterday := gotime.Now().Add(-24 * gotime.Hour)
+		_, project, err := memdb.EnsureDefaultUserAndProject(ctx, "test", "test")
+		assert.NoError(t, err)
+
+		yesterday := gotime.Now().Add(-25 * gotime.Hour)
```
hackerwins, line 44: Have you had any problems here?

server/backend/database/project_info.go
```diff
@@ -30,6 +30,10 @@ var DefaultProjectID = types.ID("000000000000000000000000")
 // DefaultProjectName is the default project name.
 var DefaultProjectName = "default"
 
+// DefaultClientDeactivateThreshold is the default
+// deactivate threshold of clients in specific project for housekeeping.
+var DefaultClientDeactivateThreshold = 24 * time.Hour
```
hackerwins, line 35: How about receiving this value from the user when running the server with the CLI?
https://github.com/yorkie-team/yorkie/pull/454/files#diff-8563d4c8ba43b628fe4115162226d50749db653fdcd3dd2e661fb42a4671c3edL227

hackerwins (commented):
Thanks for your contribution. It looks good overall.

I left a trivial comment.

server/backend/database/memory/database.go
krapie: Thank you for your suggestion.
`candidatesLimit` has changed to `candidatesLimitPerProject` to clarify the meaning.

server/backend/database/memory/housekeeping_test.go
krapie: `findDeactivateCandidatesPerProject` inside `FindDeactivateCandidates()` needs `project` object to find deactivation candidates based on `project`'s `clientDeactivateThreshold`.

But previous test only provided `project ID`, not the whole `project`. So EnsureDefaultUserAndProject() was needed to get proper `project`.

server/backend/database/project_info.go
krapie: Thank you for your suggestion!

I added `--client-deactivate-threshold` in CLI to set default `Project`'s `ClientDeactivateThreshold` during project creation. Therefore, there are some changes in project creation code to receive CLI's `ClientDeactivateThreshold`.

hackerwins (approved):
Thanks for your contribution.

This PR will be helpful to other developers as they develop other global settings that need to be moved to project settings.

PR #439, opened 2023-01-05
Fix wrong string escape in RichText's attrs

cozitive:
**What this PR does / why we need it**:
In RichText's attrs, missing backslash before special characters caused JSON parse error. RHT.Marshal() now inserts escape characters when constructing JSON encoding.

**Which issue(s) this PR fixes**:

Fixes #437

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Added relevant tests or not required
- [ ] Didn't break anything

Issue #437, opened 2022-12-22
Handle `JSON.parse` error properly


chacha912:
**Description**:

When `JSON.parse` fails, an error that the document failed to load should be displayed.

<img width="1290" alt="image" src="https://user-images.githubusercontent.com/81357083/209075597-c217c77b-f59b-4745-be73-f8c03947bd6a.png">

An error occurs when accessing the `examples$20221216` document in the default project from the admin account.
https://yorkie.dev/dashboard/projects/default/documents/examples$20221216

**Why**:
If only an empty screen is displayed, users don't know why the page is not working.

cozitive:
Fixed by #443

PR #458, opened 2023-02-07
Remove duplicated backslash in string escaping

cozitive:
At PR #439, I escaped RHT key and value strings to fix issue #437.
When checking `EscapeString()` function at that time, I thought it should put two backslashes in front of special characters, because `JSON.stringify()` in JavaScript added two backslashes. For example, the return value of `JSON.stringify({"key": "value\n"})` is `'{"key":"value\\n"}'`.
However, the additional backslash in the example above is just to represent a single backslash in JS string. Valid JSON string uses only a single backslash for control characters. 
Since the updated `EscapeString()` makes invalid JSON strings, it should be reverted.

Fixes #439 and #443

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,9 @@ test: ## runs integration tests that require local applications such as MongoDB
 	go test -tags integration -race ./...
 
 bench: ## runs benchmark tests
-	set -o pipefail && go test -tags bench -benchmem -bench=. ./test/bench -memprofile=mem.prof -cpuprofile=cpu.prof | tee output.txt
+	mkfifo pipe
+	tee output.txt < pipe &
+	go test -tags bench -benchmem -bench=. ./test/bench -memprofile=mem.prof -cpuprofile=cpu.prof > pipe
 
 docker: ## builds docker images with the current version and latest tag
 	docker buildx build --push --platform linux/amd64,linux/arm64,linux/386 -t yorkieteam/yorkie:$(YORKIE_VERSION) -t yorkieteam/yorkie:latest .
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ test: ## runs integration tests that require local applications such as MongoDB
 	go test -tags integration -race ./...
 
 bench: ## runs benchmark tests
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -59,7 +59,7 @@ test: ## runs integration tests that require local applications such as MongoDB
 	go test -tags integration -race ./...
 
 bench: ## runs benchmark tests
-	go test -tags bench -benchmem -bench=. ./test/bench -memprofile=mem.prof -cpuprofile=cpu.prof | tee output.txt
+	set -o pipefail && go test -tags bench -benchmem -bench=. ./test/bench -memprofile=mem.prof -cpuprofile=cpu.prof | tee output.txt
```
cozitive, line 62: `pipefail` is added to preserve exit status code of benchmark test.

Makefile
```diff
@@ -59,7 +59,9 @@ test: ## runs integration tests that require local applications such as MongoDB
 	go test -tags integration -race ./...
 
 bench: ## runs benchmark tests
-	go test -tags bench -benchmem -bench=. ./test/bench -memprofile=mem.prof -cpuprofile=cpu.prof | tee output.txt
```
cozitive, line 62: The exit status of benchmark test has been ignored, because the exit status of a pipeline is the exit status of the last command (`tee`) in the pipeline.
To reveal the original exit status, piping command is separated.

Makefile
```diff
@@ -59,7 +59,9 @@ test: ## runs integration tests that require local applications such as MongoDB
...
+	mkfifo pipe
+	tee output.txt < pipe &
+	go test -tags bench -benchmem -bench=. ./test/bench -memprofile=mem.prof -cpuprofile=cpu.prof > pipe
```
cozitive, line 64: The exit status of benchmark test has been ignored, because the exit status of a pipeline is the exit status of the last command (tee) in the pipeline.
To reveal the original exit status, piping command is separated.

hackerwins: Thanks for the detailed description.

hackerwins (approved):
Thanks for your contribution.

PR #462, opened 2023-02-09
feat: remove priority queue from RHTPQMap and entire project

### LINKED ISSUE #423: Remove `Heap(Priority queue)` from `RHTPQMap`
We need to consider removing `Heap` from `RHTPQMap/elementQueueMapByKey`.

### Current
```swift
var elementQueueMapByKey: [String: Heap<TimeTicket, RHTPQMapNode>] = [:]
```

### Suggestion
```swift
var elementMapByKey: [String: (TimeTicket, RHTNode)] = [:]
```

**Why is this needed**:
`RHTPQMap` always uses the max value of `Heap`. We only need to hold the maximum value.

blurfx:
- `RHTPQMap` can have multiple elements per key, but we need only the most recently created element.
- Therefore, change the `RHTPQMap` to have only one recently created element.
- `RHTPQMap` was the only reference of `PriorityQueue`, but now it's not, so, remove `PriorityQueue` from the entire project.
- Renamed `RHTPQMap` to `ElementRHT`

Address #423 

I don't know how to write a test case to cover this change properly. need help.

```release-note

```

**Additional documentation**:

https://github.com/yorkie-team/yorkie/pull/462/files#diff-66703d803f9ced520eb5b0ba16e3f2af6c933fa9669fb90109329766e6b208e1

**Checklist**:
- [x] Didn't break anything


hackerwins:
Before:
Screenshot 2023-02-15 at 2 53 06 PM

After:
Screenshot 2023-02-15 at 2 54 29 PM

hackerwins:
After changing `RHTPQMap` to `RHTElement`, the previous value will be physically removed and not remain with the tombstone when we overwrite a value on the same key of the object.

How about adding the relevant test code for this case and making sure everything goes ok?

```
Initial State: { a: [1, 2] }

Client A: { a: [1, 2, 3] } // push 3 to the array
Client B: { a: [] } // delete previous array by setting a new array

Final State: { a: [] }
```

blurfx:
I wrote a small test code in `integration/document_test.go` to verify that the element was correctly removed.
Please let me know if I've written it differently.

blurfx:
@hackerwins I wrote a small test code in `integration/document_test.go` to verify that the element was correctly removed.
Please let me know if I've written it differently.

pkg/document/crdt/element_rht.go
```diff
@@ -0,0 +1,211 @@
...
+	if !ok || node == nil {
+		return nil
+	}
+
+	if !node.Remove(deletedAt) {
+		return nil
+	}
+
+	return node.elem
+}
```
hackerwins, line 143: Is there any place to use this code? `DeleteByCreatedAt`

pkg/document/crdt/element_rht.go
```diff
@@ -0,0 +1,211 @@
...
+
+// purge physically purge child element.
+func (rht *ElementRHT) purge(elem Element) {
+	node, ok := rht.nodeMapByCreatedAt[elem.CreatedAt().Key()]
+	if !ok {
+		panic("fail to find: " + elem.CreatedAt().Key())
+	}
+	if node == nil {
+		return
+	}
```
hackerwins, line 177: We are checking existence with `ok`, but is there a reason to separately check if `node` is nil?

pkg/document/crdt/element_rht.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// Set sets the value of the given key. If there is an existing value, it is removed.
+func (rht *ElementRHT) Set(k string, v Element) Element {
+	node, ok := rht.nodeMapByKey[k]
+	var removed Element
+	if ok && !node.isRemoved() && v.CreatedAt().After(node.elem.CreatedAt()) {
+		node.Remove(v.CreatedAt())
+		removed = node.elem
+	}
```
hackerwins, line 107: It seems that some of the logic is redundant. Is it possible to change it like the below?

```suggestion
	if ok && node.Remove(v.CreatedAt()) {
		removed = node.elem
	}
```

pkg/document/crdt/element_rht.go
```diff
@@ -0,0 +1,211 @@
...
+
+	return node.elem
+}
+
+// DeleteByCreatedAt deletes the Element of the given creation time.
+func (rht *ElementRHT) DeleteByCreatedAt(createdAt *time.Ticket, deletedAt *time.Ticket) Element {
+	node, ok := rht.nodeMapByCreatedAt[createdAt.Key()]
+	if !ok || node == nil {
+		return nil
+	}
```
hackerwins, line 136:
```suggestion
	if !ok {
		return nil
	}
```

pkg/document/crdt/element_rht.go
```diff
@@ -0,0 +1,211 @@
...
+	return n.elem
+}
+
+// ElementRHT is a hashtable with logical clock(Replicated hashtable).
+// The difference from RHT is that it keeps multiple values in one key. Using
+// Max Heap, the recently inserted value from the logical clock is returned
+// to the outside.
+type ElementRHT struct {
+	nodeMapByKey       map[string]*ElementRHTNode
+	nodeMapByCreatedAt map[string]*ElementRHTNode
```
hackerwins, line 70:
```suggestion
	// nodeMapByCreatedAt is a map with values ​​of nodes by creation time.
	// Even if an element is removed by `set` or `delete`, it remains in
	// nodeMapByCreatedAt and will be deleted physically by GC.
	nodeMapByCreatedAt map[string]*ElementRHTNode
```

hackerwins (commented):
Thanks for adding the test code. I left a few comments.

pkg/document/crdt/element_rht.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// Element returns the element of this node.
+func (n *ElementRHTNode) Element() Element {
+	return n.elem
+}
+
+// ElementRHT is a hashtable with logical clock(Replicated hashtable).
+type ElementRHT struct {
+	nodeMapByKey       map[string]*ElementRHTNode
```
hackerwins, line 66:
```suggestion
	// nodeMapByKey is a map with values of nodes by key.
	nodeMapByKey       map[string]*ElementRHTNode
```

pkg/document/crdt/element_rht.go
blurfx: Yes. `DeleteByCreatedAt` method of `crdt.Object` is still using it.

```go
// pkg/document/crdt/object.go

func (o *Object) DeleteByCreatedAt(createdAt *time.Ticket, deletedAt *time.Ticket) Element {
	return o.memberNodes.DeleteByCreatedAt(createdAt, deletedAt)
}
```

pkg/document/crdt/element_rht.go
blurfx: I had a nil check to handle exceptions while I was writing this pr, but I don't think we need it(And other commented things too) now.

hackerwins (approved):

PR #467, opened 2023-02-14
support validation for DocumentKey

### LINKED ISSUE #457: Implement validation for Document Key
**Description**: 
Currently, there are no restrictions on the document key when creating a document. 
Any characters can be used and no errors will occur. 

However, this causes issues when using the [Dashboard](https://github.com/yorkie-team/dashboard),
as the key is used in the URL and inappropriate characters can cause problems.
- Dashboard URL example - document search: 
`https://yorkie.dev/dashboard/projects/project-name/documents/document-key`

---

1. It is necessary to add rules for the allowed characters in the document key.
  - maximum length restrictions: ex) `From 4 to 30 characters in length`
  - can be used as a slug: ex) `Only contains letters, numbers, and the following characters: - . _ ~`
  - related https://github.com/yorkie-team/yorkie/pull/345
2. It would be helpful to provide a method to shorten the key.
  - For example, if the key is made by combining project, block, edit, file like `prj-block-edit-file`, it is inconvenient because the key is too long. If there is a way to shorten the key, it would be helpful for users.

**Why**: Currently, any characters can be used as the document key. It can cause problems when using the dashboard.

easylogic:

The document key `must be valid` because it must be usable as the `path segment` of the `url`.

Fixes #457

* The Documentkey is used as the path segment of the url.
* DocumentKey takes the form of a regular expression. (slugRegexString = `^[a-z0-9\-._~]+$`)
* The length of DocumentKey is limited to 4 ~ 30.

```release-note
* support validation for DocumentKey
```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Didn't break anything


test/integration/array_test.go
```diff
@@ -57,8 +58,8 @@ func TestArray(t *testing.T) {
 
 	t.Run("concurrent array add/delete simple test", func(t *testing.T) {
 		ctx := context.Background()
-
-		d1 := document.New(key.Key(t.Name()))
+		testDocumentKey := "concurrent-array-add-simple"
```
hackerwins, line 61: We used `t.Name` to generate a unique document key for each test and used it to prevent documents created in different tests from interfering with each other.

If we manually generate unique names per test, we don't know if they interfere with each other or not.

How about adding a function in `helper` package that changes the name generated by t.Name to valid name of key.Key?

https://github.com/yorkie-team/yorkie/blob/main/test/helper/helper.go#L75

pkg/document/key/key_test.go
```diff
@@ -0,0 +1,67 @@
...
+	t.Run("invalid key", func(t *testing.T) {
+		key := Key("invalid key") // space is not allowed
+
+		err := key.Validate()
+
+		assert.NotNil(t, err, "key should be invalid: with space")
+
+		key = Key("invalid-key-~$a") // last character should be alphanumeric
+		err = key.Validate()
+		assert.NotNil(t, err, "key should be invalid: with -")
```
hackerwins, line 53: How about comparing error values?

```suggestion
		assert.Equal(t, err, key.ErrInvalidKey, "key should be invalid: with -")
```

hackerwins (changes requested):
Thanks for your contribution and congratulations on your first PR. 🍻

I leave a few questions.

It seems that there are duplicate codes with [validation.go](https://github.com/yorkie-team/yorkie/blob/main/api/types/fields_validation.go), how should we solve it?

server/rpc/yorkie_server.go
```diff
@@ -129,6 +129,11 @@ func (s *yorkieServer) AttachDocument(
 		return nil, err
 	}
 
+	// check DocumentKey is valid before it is attached
+	if err := pack.DocumentKey.Validate(); err != nil {
```
cozitive, line 133: Is document key validation needed per every attachment?
I guess that it is because `AttachDocuent()` does both  attachment and creation by DB upsert.

test/integration/array_test.go
easylogic: Using the test name as the document key seems more inappropriate. To match the current rules, we need to recreate a new function.

In this case, eventually, the test code name must be uniquely matched to the document key rule.

It takes more dependencies.

I think the test name itself has nothing to do with the document key.

pkg/document/key/key_test.go
easylogic: I bundled the error occurrences together and checked them with NotNil. 

It doesn't matter if you do a direct comparison.

server/rpc/yorkie_server.go
easylogic: That's a good comment.

I thought a lot about this part.

I also don't think it's appropriate to check every time in attach.

However, the operation of mongo and memory database is different.

In the case of mongo, upsert is one action, so we need to separate it into two actions.

I didn't make a clear decision on the state that should be detached, so I checked it before Attach.

I think this part needs more discussion.

easylogic:
@hackerwins  

ci build test runner is different to local test runner.

There are no errro in local test runner.
But ci test runner has errors when test is build.

For example 

last ci build has a test error. 

It is a `snapshot_test.go` file. 
But local test runner has no error.

Why it happened?

easylogic:
New validator 

```go
	err := validation.ValidateDynamically(k.String(), []any{
		"required",
		"slug",
		"min=4",
		"max=30",
		validation.CustomRule{
			Func: func(fl validator.FieldLevel) bool {
				return fl.Field().String()[0] == '$'
			},
		},
	})

	if err != nil {
		return ErrInvalidKey
	}
```

test/integration/array_test.go
hackerwins: I thought the helper function was something like the below:

The helper function appears to be designed to return a valid document key for testing purposes. The code takes the current test's name and converts it into a string that is safe to use as a document key. The resulting string is lowercase, alphanumeric, and any non-alphanumeric characters are replaced with hyphens.

```go
// TestDocKey returns the valid document key for test.
func TestDocKey(t *testing.T) string {
	name := t.Name()

	if len(name) < 3 {
		log.Fatal("test name should be longer than 3")
	}

	if len(name) > 60 {
		name = name[:60]
	}

	sb := strings.Builder{}
	for _, c := range name {
		if c >= 'A' && c <= 'Z' {
			sb.WriteRune(c + ('a' - 'A'))
		} else if c >= 'a' && c <= 'z' {
			sb.WriteRune(c)
		} else if c >= '0' && c <= '9' {
			sb.WriteRune(c)
		} else {
			sb.WriteRune('-')
		}
	}

	return sb.String()
}

// Replace `t.Name()` with `helper.TestDocKey(t)` in tests
// before: `TestProjectAuthWebhook/authorization_webhook_test`
// after: `testprojectauthwebhook-authorization-webhook-test`
doc := document.New(key.Key(helper.TestDocKey(t)))
```

hackerwins:
We use `Build Constraints` for organizing the testing phase such as integration tests, benchmark tests, and so on. Additionally, `Build Constraints` can be used for specific machine architectures. As a result, tests could differ between local and CI.

You can find more information about `Build Constraints` in Go at https://pkg.go.dev/go/build#hdr-Build_Constraints

test/integration/array_test.go
easylogic: I added help.TestDocumentKey and help.BenchDocumentKey function. 

help.TestDocumentKey return automatically unique valid document key by t.Name()

internal/validation/validation_manager.go
```diff
@@ -0,0 +1,185 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/go-playground/locales/en"
+	ut "github.com/go-playground/universal-translator"
+	"github.com/go-playground/validator/v10"
+)
+
+const (
+	// slugRegexString regular expression for key validation
```
hackerwins, line 32:
```suggestion
	// slugRegexString regular expression for slug validation.
```

pkg/document/key/key.go
```diff
@@ -17,10 +17,38 @@
...
+
 // Key represents a document key.
 type Key string
 
 // String returns the string representation of the key.
 func (k Key) String() string {
 	return string(k)
 }
+
+// Validate returns true if the key is valid.
```
hackerwins, line 39:
```suggestion
// Validate checks whether the key is valid or not.
```

pkg/document/key/key_test.go
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2020 The Yorkie Authors. All rights reserved.
```
hackerwins, line 2:
```suggestion
 * Copyright 2023 The Yorkie Authors. All rights reserved.
```

test/helper/helper.go
```diff
@@ -164,3 +170,52 @@ func TestServer() *server.Yorkie {
...
+			sb.WriteRune('-')
+		}
+	}
+
+	return sb.String()
+}
+
+// TestDocumentKey returns a new instance of document key for testing.
+func TestDocumentKey(t *testing.T) string {
+	return getDocumentKey(t.Name())
```
hackerwins, line 215:
```suggestion
	return testDocumentKey(t.Name())
```

test/helper/helper.go
```diff
@@ -164,3 +170,52 @@ func TestServer() *server.Yorkie {
...
+}
+
+// BenchDocumentKey returns a new instance of document key for benchmarking.
+func BenchDocumentKey(b *testing.B) string {
+	return getDocumentKey(b.Name())
```
hackerwins, line 220:
```suggestion
	return testDocumentKey(b.Name())
```

test/helper/helper.go
```diff
@@ -164,3 +170,52 @@ func TestServer() *server.Yorkie {
 	}
 	return y
 }
+
+func getDocumentKey(name string) string {
```
hackerwins, line 174: https://go.dev/doc/effective_go#Getters

```suggestion
func testDocumentKey(name string) string {
```

hackerwins (commented):
Thanks for your contribution.

It would be nice to just clean up the minor code styles and then merge them.

internal/validation/validation_manager.go
```diff
@@ -0,0 +1,185 @@
...
+				Err:   e,
+				Trans: e.Translate(trans),
+			}
+		}
+	}
+	return nil
+}
+
+func init() {
+	// register validation for key
```
hackerwins, line 175:
```suggestion
```

hackerwins:
This will merged by #473

Makefile
```diff
@@ -59,10 +59,11 @@ test: ## runs integration tests that require local applications such as MongoDB
 	go test -tags integration -race ./...
 
 bench: ## runs benchmark tests
+	rm -f pipe output.txt mem.prof cpu.prof bench.test
+	make bench-clean
 	mkfifo pipe
 	tee output.txt < pipe &
 	go test -tags bench -benchmem -bench=. ./test/bench -memprofile=mem.prof -cpuprofile=cpu.prof > pipe
```
hackerwins, line 66:
```suggestion
bench: ## runs benchmark tests
	-rm -f pipe output.txt mem.prof cpu.prof bench.test
	mkfifo pipe
	tee output.txt < pipe &
	go test -tags bench -benchmem -bench=. ./test/bench -memprofile=mem.prof -cpuprofile=cpu.prof > pipe
	
```

hackerwins (approved):
Now we have a validation module.

hackerwins:
Users can express data information using keys. The 30-character key limit was too short, so I increased it to 120 characters.

https://firebase.google.com/docs/firestore/quotas

PR #477, opened 2023-02-28
Add ClientDeactivateThreshold field in admin CLI project list

krapie:

1. Add `ClientDeactivateThreshold` field in project list admin CLI.

`ClientDeactivateThreshold` field was omitted in project list in admin CLI.

~~2. Set minimum value limit on `ClientDeactivateThreshold` by restricting string value to must contain hour prefix(`h` prefix). This will ensure `ClientDeactivateThreshold` value to have minimum value limit of `1 hour`.~~
~~Furthermore, there was no minimum value limit on `ClientDeactivateThreshold`, which may result in excessive client deactivation.
For example, if `ClientDeactivateThreshold` is set to `01s`(or any other smaller value than `housekeepingInterval`), client may deactivate every time housekeeping occurs.~~

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/cmd/yorkie/project/list.go
+++ b/cmd/yorkie/project/list.go
@@ -64,6 +64,7 @@ func newListCommand() *cobra.Command {
 				"SECRET KEY",
 				"AUTH WEBHOOK URL",
 				"AUTH WEBHOOK METHODS",
+				"CLIENT DEACTIVATE THRESHOLD",
 				"CREATED AT",
 			})
 			for _, project := range projects {
@@ -73,6 +74,7 @@ func newListCommand() *cobra.Command {
 					project.SecretKey,
 					project.AuthWebhookURL,
 					project.AuthWebhookMethods,
+					project.ClientDeactivateThreshold,
 					units.HumanDuration(time.Now().UTC().Sub(project.CreatedAt)),
 				})
 			}
--- a/internal/validation/validation_test.go
+++ b/internal/validation/validation_test.go
@@ -35,6 +35,12 @@ func TestValidation(t *testing.T) {
 
 		err = ValidateValue("invalid-key-$-wrong-string-value", "required,slug,min=4,max=30")
 		assert.Equal(t, err.(Violation).Tag, "slug")
+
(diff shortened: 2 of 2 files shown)


internal/validation/validation.go
```diff
@@ -37,7 +37,7 @@ const (
 	containNumberRegexString          = `[0-9]`
 	containSpecialCharRegexString     = `[\{\}\[\]\/?.,;:|\)*~!^\-_+<>@\#$%&\\\=\(\'\"\x60]`
 	alphaNumberSpecialCharRegexString = `^[a-zA-Z0-9\{\}\[\]\/?.,;:|\)*~!^\-_+<>@\#$%&\\\=\(\'\"\x60]+$`
-	timeDurationFormatRegexString     = `^(\d{1,2}h\s?)?(\d{1,2}m\s?)?(\d{1,2}s)?$`
+	timeDurationFormatRegexString     = `^(\d{1,2}h\s?)(\d{1,2}m\s?)?(\d{1,2}s)?$`
```
hackerwins, line 40: `timeDurationFormatRegexString` is used to check whether the given string for `duration` is valid, so inserting the `1h` check seems to be out of the scope of the variable.

Responsibility for setting it is up to the user, so how about trusting them to set the appropriate values ​​for now?

hackerwins (changes requested):
Thanks for your contribution.

I left a brief comment.

internal/validation/validation.go
krapie: > `timeDurationFormatRegexString` is used to check whether the given string for `duration` is valid, so inserting the `1h` check seems to be out of the scope of the variable.

Yes, checking minimum limit on duration string format regex is not appropriate. It is tricky where to put limit checks...

> Responsibility for setting it is up to the user, so how about trusting them to set the appropriate values ​​for now?

Project settings will be used for application developers, or administrators. Maybe trusting them is right.
Reverted minimum limit related codes.

hackerwins (approved):
LGTM.

PR #479, opened 2023-03-02
Lock watchDocuments depending on the client

chacha912:

If a client requests to attach to multiple documents simultaneously, 
the existing connection will be disconnected and reconnected. 

Depending on the order of the published events, problems may occur on the client. 
If the `watched` event is published before the `unwatched` event, the test will fail 
since the final state is `watched`, but the `unwatched` event arrives later. 
(please refer to the [yorkie-js-sdk test](https://github.com/yorkie-team/yorkie-js-sdk/pull/464#issuecomment-1447547066)). 

image

This image shows the server log when the `watched` event is published first.

To resolve this, I added a lock for each client in the `watchDocument`.
This ensures that a lock is set when a client establishes a stream and released when the client unwatches.

Fixes https://github.com/yorkie-team/yorkie-js-sdk/pull/464#issuecomment-1447547066

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/server/rpc/yorkie_server.go
+++ b/server/rpc/yorkie_server.go
@@ -18,6 +18,7 @@ package rpc
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/yorkie-team/yorkie/api/converter"
 	"github.com/yorkie-team/yorkie/api/types"
@@ -394,7 +395,7 @@ func (s *yorkieServer) WatchDocuments(
 
 	locker, err := s.backend.Coordinator.NewLocker(
 		stream.Context(),
-		sync.NewKey(cli.ID.String()),
+		sync.NewKey(fmt.Sprintf("watchdocs-%s", cli.ID.String())),
 	)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


server/rpc/yorkie_server.go
```diff
@@ -392,13 +392,27 @@ func (s *yorkieServer) WatchDocuments(
...
 	subscription, peersMap, err := s.watchDocs(stream.Context(), *cli, docKeys)
 	if err != nil {
 		logging.From(stream.Context()).Error(err)
 		return err
 	}
 	defer func() {
 		s.unwatchDocs(docKeys, subscription)
+		if err := locker.Unlock(stream.Context()); err != nil {
+			logging.DefaultLogger().Error(err)
+		}
```
hackerwins, line 415: 1. If an error is returned by `s.watchDocs`, releasing the lock is not executed. So, How about moving `locker.Unlock` to just below `locker.lock` with defer?

2. `locker.Unlock` may not be executed when the stream is terminated and the context is in the cancel state. How about using `context.Background` instead of `stream.Context`?

3. We need to separate the key of the lock with `snapshot-*` or `pushpull-*`. How about creating a key with the prefix `watchdocs-*`?

hackerwins (commented):
Thanks for your contribution.

I left a few comments.

hackerwins (approved):
LGTM. 👍

PR #501, opened 2023-03-24
remove panic & wrapping informational error

### LINKED ISSUE #497: Don't Panic in the Server
**Description**:

We often use panic in development to reveal errors, but codes running in production must avoid panic to prevent stopping the server.

Refer to https://github.com/uber-go/guide/blob/master/style.md#dont-panic

**Why**:
- Prevent stopping the server.

emplam27:

ref. https://github.com/uber-go/guide/blob/master/style.md#dont-panic

Remove `panic()` in server production
- Change function from using panic to return error and caller to decide how to handle it.
- Wrapping error more informationally

Remove panic method & Wrapping errors in CRDT logics
- Currently, if error occurred in CRDT logic, server was shutdown because of `panic()` method
- Removed `panic()` methods & Change all errors able to logging

Fixes #497 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins (changes requested):
Thanks for your contribution.

I think the interface has changed only for the function below:
  - validation.RegisterValidation
  - validation.RegisterTranslation
  - config.configPath
  - time.ParseDuration

Next, it would be nice to remove the panic that can occur in CRDT data structure.

hackerwins (commented):
Thanks for the second work. I leave a few comments.

### We need to check whether an actual error can occur

While removing panic, some interfaces have been changed to return an error. The changes may reduce the usability of interfaces, It is necessary to check whether an actual error can occur.

For example, there is a situation like below:

An error chain `Text.String > RGATreeSplitNode.String > TextValue.String` need not return an actual error.

```go
func (t *TextValue) String() (string, error) {
	return t.value, nil
}
```

PR #503, opened 2023-03-30
Check client Sequence while pushpulled

humdrum:
- Add sequence validation logic.
  - When process push changes, the client sequence has to be prevClientSeq + 1 
  - If the client sends an invalid ServerSeq, Send a snapshot so that the client can try to fix it.
- Preserve sequence numbers for the push-pull response of detach and remove API. 

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Didn't break anything


hackerwins:
I think we need a simple design on how to deal with this issue. For now, I will register a separate issue and close this PR. #805

PR #512, opened 2023-04-12
Add gRPC MaxConnectionAge & MaxConnectionAgeGrace options

krapie:

Add gRPC `MaxConnectionAge` & `MaxConnectionAgeGrace` options.

This is to avoid "split-brain" of long-lived gRPC stream connection when performing hash-based load balancing in Yorkie cluster.

For more information, follow: https://github.com/envoyproxy/envoy/issues/26459

gRPC suggest to use `MAX_CONNECTION_AGE` to avoid long-lived RPC issue on load balancing(https://youtu.be/Naonb2XD_2Q?t=136), but stream is not actually closed even `MAX_CONNECTION_AGE` is configured.

This is because after `MAX_CONNECTION_AGE`, server sends GoAway to client, but GoAway is not a signal to close connection instantly. It's just a signal to tell client not to send additional request to server(https://github.com/grpc/grpc-java/issues/8770)

Therefore, `MAX_CONNECTION_AGE_GRACE` is also introduced to forcibly close stream.

So this is how it works:
1. After stream is established, stream is available for `maxConnectionAge`.
2. After `maxConnectionAge`, GoAway frame is sent to client to gracefully close the stream. But stream is not closed at this point.
3. After `MaxConnectionAgeGrace`, stream is forcibly closed.

Reference: https://pkg.go.dev/google.golang.org/grpc/keepalive

Hence, `maxConnectionAge` and `maxConnectionAgeGrace` options are added as gRPC server's keepalive option parameters.

Fixes #

As I researched some information about long-lived connection, graceful close, etc, I now have good explanation about why `close_connections_on_host_set_change`(gracefully draining connections) are not working for long-lived gRPC connections.

It is because envoy sends HTTP2 GoAway frame on draining sequence, but GoAway is not for instant connection close. Therefore, no connection close are made after draining, and `close_connections_on_host_set_change` option will not work as well.

References: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/operations/draining

Also, I found some interesting issue related to our issue on long-lived connection close: https://github.com/grpc/grpc/issues/26703.

Maybe it will be useful to introduce `force_close_connections_on_host_set_change` on envoy:: https://github.com/envoyproxy/envoy/issues/26459

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/cmd/yorkie/server.go
+++ b/cmd/yorkie/server.go
@@ -191,6 +191,12 @@ func init() {
 		server.DefaultRPCMaxRequestsBytes,
 		"Maximum client request size in bytes the server will accept.",
 	)
+	cmd.Flags().StringVar(
+		&conf.RPC.MaxConnectionAge,
+		"rpc-max-connection-age",
+		server.DefaultRPCMaxConnectionAge.String(),
+		"Maximum duration of connection may exist before it will be closed by sending a GoAway.",
+	)
 	cmd.Flags().IntVar(
 		&conf.Profiling.Port,
 		"profiling-port",
--- a/cmd/yorkie/server.go
+++ b/cmd/yorkie/server.go
@@ -197,6 +197,12 @@ func init() {
 		server.DefaultRPCMaxConnectionAge.String(),
 		"Maximum duration of connection may exist before it will be closed by sending a GoAway.",
 	)
+	cmd.Flags().StringVar(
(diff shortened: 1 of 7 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #522, opened 2023-04-25
Remove panic method in crdt text

emplam27:
Remove panic method in crdt text
- go sdk crdt text has the condition that cause panic
- remove panic prevent shutdown server when crdt text error occurred

Fixes #501 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/api/converter/from_bytes.go
+++ b/api/converter/from_bytes.go
@@ -163,43 +163,43 @@ func fromJSONText(
 ) (*crdt.Text, error) {
 	createdAt, err := fromTimeTicket(pbText.CreatedAt)
 	if err != nil {
-		return nil, fmt.Errorf("from bytes fromJsonText: %w", err)
+		return nil, err
 	}
 	movedAt, err := fromTimeTicket(pbText.MovedAt)
 	if err != nil {
-		return nil, fmt.Errorf("from bytes fromJsonText: %w", err)
+		return nil, err
 	}
 	removedAt, err := fromTimeTicket(pbText.RemovedAt)
 	if err != nil {
-		return nil, fmt.Errorf("from bytes fromJsonText: %w", err)
+		return nil, err
 	}
 
 	rgaTreeSplit, err := crdt.NewRGATreeSplit(crdt.InitialTextNode())
 	if err != nil {
-		return nil, fmt.Errorf("from bytes fromJsonText: %w", err)
+		return nil, err
 	}
 
 	current := rgaTreeSplit.InitialHead()
 	for _, pbNode := range pbText.Nodes {
 		textNode, err := fromTextNode(pbNode)
 		if err != nil {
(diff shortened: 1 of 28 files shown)


hackerwins:
@emplam27 Thanks for your contribution.

`crdt.Text` uses RGATreeSplit and SplayTree. So To remove all panic from `crdt.Text`, we also need to remove panic in `RGATreeSplit` and `SplayTree`.

```
### pkg/document/crdt/rga_tree_split.go

46:// argument is nil, it would panic at runtime.
49:		panic("RGATreeSplitNodeID cannot be null")
351:		panic("the node of the given id should be found: " + s.StructureAsString())
367:		panic("offset should be less than or equal to length: " + s.StructureAsString())

### pkg/splay/splay.go
206:		panic(fmt.Sprintf(
```

https://github.com/yorkie-team/yorkie/blob/main/design/data-structure.md#overview

But since SplayTree is also used in Array, I think it would be better to remove it from another PR(e.g. Remove panic in `crdt.Array`).

hackerwins (changes requested):
@emplam27 Thanks for your contribution.

`crdt.Text` uses RGATreeSplit and SplayTree. So To remove all panic from `crdt.Text`, we also need to remove panic in `RGATreeSplit` and `SplayTree`.

```
### pkg/document/crdt/rga_tree_split.go

46:// argument is nil, it would panic at runtime.
49:		panic("RGATreeSplitNodeID cannot be null")
351:		panic("the node of the given id should be found: " + s.StructureAsString())
367:		panic("offset should be less than or equal to length: " + s.StructureAsString())

### pkg/splay/splay.go
206:		panic(fmt.Sprintf(
```

https://github.com/yorkie-team/yorkie/blob/main/design/data-structure.md#overview

But since SplayTree is also used in Array, I think it would be better to remove it from another PR(e.g. Remove panic in `crdt.Array`).

emplam27:
@hackerwins 
i'm done to remove `rga_tree_split`'s panic.
then, how about `element_rht` and `rga_tree_list`'s panic? is it ok to remove those panics?

```
### pkg/document/crdt/element_rht.go
173:		panic("fail to find: " + elem.CreatedAt().Key())

### pkg/document/crdt/rga_tree_list.go
209:		panic("fail to find the given createdAt: " + createdAt.Key())
241:		panic("fail to find the given prevCreatedAt: " + prevCreatedAt.Key())
246:		panic("fail to find the given createdAt: " + createdAt.Key())
261:		panic("fail to find the given prevCreatedAt: " + createdAt.Key())
278:		panic("fail to find the given createdAt: " + elem.CreatedAt().Key())
290:		panic("fail to find the given createdAt: " + createdAt.Key())
```

hackerwins (approved):

PR #492, opened 2023-03-20
Add user agent metrics

emplam27:
**What this PR does / why we need it**:

- doc : https://codepair.navercorp.com/p9diyg
- Add metric about yorkie-sdk-type/version, projectID/Name, server instance name etc.
- every grpc request success, collect metric about yorkie-sdk information
- `yorkie-js-sdk`, `yorkie-ios-sdk`, `yorkie-android-sdk` need to set header and put yorkie-sdk informations
- User can set `hostname` when the server start (hostname is used in metrics etc.)
  - Use CLI flags : `$ yorkie server ... --hostname {host name value} ...`
  - Use config.yaml : `Backend.Hostname` field value

**Which issue(s) this PR fixes**:

Fixes #393 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Server hostname settings. hostname is used in metrics
- Use CLI flags : `$ yorkie server ... --hostname {host name value} ...`
- Use config.yaml : `Backend.Hostname` field value
```

**Additional documentation**:

```docs
yorkie documentation : https://yorkie.dev/docs/cli#commands
- required add `--hostname` flag
```

**Checklist**:
- [x] Added relevant tests or not required
- [x] Didn't break anything

PR #529, opened 2023-05-03
Add Leader Election for Housekeeping in Sharded Cluster using MongoDB

### LINKED ISSUE #505: Handle housekeeping with sharded cluster

We recently introduced cluster mode based on the shard key. This shard key is used to assign APIs requested by clients to servers in the cluster based on project or document.

Housekeeping operations should also apply this sharded cluster structure. We can think of two options:

1. Only the master server executes housekeeping in the cluster
2. Split housekeeping tasks per project and execute them on servers mapped by project

**Why is this needed**:

- Bugs may occur in housekeeping of sharded cluster.

krapie:

Add leader election for housekeeping in sharded cluster using mongodb.

- `server/backend/election` package is added for leader election, with `database` implementation.
- `database` implementation uses mongoDB and [TTL indexes](https://www.mongodb.com/docs/manual/core/index-ttl/) to ensure that only one node can acquire housekeeping leader lease by preventing other nodes try to update the same document simultaneously on document expiry.
- `yorkie-cluster` Helm Chart is configured to enable `--housekeeping-leader-election` to perform leader-only housekeeping in sharded cluster.
- Also, small error in `hostname` value in `server/backend/backend.go` is fixed. For more information, follow: [Add user agent metrics PR](https://github.com/yorkie-team/yorkie/pull/492/files#r1175507046)

Fixes #505

I have tested on K8s environment with `yorkie-cluster` Helm chart with minikube, and it worked as expected.

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/server/backend/election/database/election.go
+++ b/server/backend/election/database/election.go
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2023 The Yorkie Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package database is a database implementation of election package.
+package database
+
+import (
+	"context"
+	"time"
+
+	"github.com/yorkie-team/yorkie/server/backend/database"
+	"github.com/yorkie-team/yorkie/server/logging"
+)
+
+// Elector is a database-based implementation of election.Elector.
+type Elector struct {
+	database database.Database
+
+	hostname string
+
+	ctx        context.Context
+	cancelFunc context.CancelFunc
+}
+
+// NewElector creates a new elector instance.
+func NewElector(
+	hostname string,
+	database database.Database,
+) *Elector {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+
+	return &Elector{
+		database: database,
+
+		hostname: hostname,
+
+		ctx:        ctx,
+		cancelFunc: cancelFunc,
+	}
+}
+
+// StartElection starts leader election.
+func (e *Elector) StartElection(
+	leaseLockName string,
+	leaseDuration time.Duration,
+	onStartLeading func(ctx context.Context),
+	onStoppedLeading func(),
+) error {
+	if err := e.database.CreateTTLIndex(context.Background(), leaseDuration); err != nil {
+		return err
+	}
+
+	go e.run(leaseLockName, leaseDuration, onStartLeading, onStoppedLeading)
+	return nil
+}
+
+// Stop stops all leader elections.
+func (e *Elector) Stop() error {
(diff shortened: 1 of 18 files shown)


server/backend/election/database/election.go
```diff
@@ -0,0 +1,124 @@
...
+			logging.From(ctx).Infof(
+				"leader lost: %s", e.hostname,
+			)
+		}
+
+		select {
+		case <-time.After(leaseDuration):
+		case <-e.ctx.Done():
+			return
+		}
```
hackerwins, line 122: I have one question.

After a specific server is elected as a leader and starts the housekeeping task, if the task takes a long time and exceeds the lease duration, resulting in the leader being released, is there a risk of other servers executing the housekeeping task redundantly?

hackerwins (commented):
Thanks for your contribution.
I left a simple question. 🙏

server/backend/election/database/election.go
```diff
@@ -0,0 +1,124 @@
...
+		return err
+	}
+
+	go e.run(leaseLockName, leaseDuration, onStartLeading, onStoppedLeading)
+	return nil
+}
+
+// Stop stops all leader elections.
+func (e *Elector) Stop() error {
+	e.cancelFunc()
```
hackerwins, line 72: How are the background routines executed in this module, handled when the server is stopped?

hackerwins (commented):
Thank you for the explanation.

Could you please add the tests for the two situations I asked about?

1. Lease renewal while handling a long task.
2. Handling background routines when shutting down the server.

server/backend/election/database/election.go
krapie: @hackerwins I have updated elector to wait for background routines to finish when shutting down the server (when elector stops). And added the tests that you have mentioned above.

krapie:
I personally think we can also use K8s CronJob Object to run housekeeping in cluster mode without using leader election for housekeeping.

Using K8s CronJob will ease overhead for the server cluster to process leader elections. This might have a big impact when you have a huge cluster of servers.

krapie:
I forgot to re-request review on this PR 😄 Requesting it now.

hackerwins:
> I personally think we can also use K8s CronJob Object to run housekeeping in cluster mode without using leader election for housekeeping.
> 
> Using K8s CronJob will ease overhead for the server cluster to process leader elections. This might have a big impact when you have a huge cluster of servers.

Using CronJob seems to be a good and straightforward approach. If we choose to use it, we can run the cluster without a leader election.

However, we should consider how to provide private APIs that are meant to be exclusively called by internal components and should not be exposed publicly. The proposed idea doesn't seem to depend on k8s, as we only need to expose the private APIs.

https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#concurrency-policy

hackerwins (commented):
Thanks for applying my requests.

> I personally think we can also use K8s CronJob Object to run housekeeping in cluster mode without using leader election for housekeeping.
> 
> Using K8s CronJob will ease overhead for the server cluster to process leader elections. This might have a big impact when you have a huge cluster of servers.

Using CronJob seems to be a good and straightforward approach. If we choose to use it, we can run the cluster without a leader election.

However, we should consider how to provide private APIs that are meant to be exclusively called by internal components and should not be exposed publicly. The proposed idea doesn't seem to depend on k8s, as we only need to expose the private APIs.

https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#concurrency-policy

How would you prefer to proceed?

krapie:
Closing the PR as per https://github.com/yorkie-team/yorkie/pull/1373.

PR #532, opened 2023-05-16
Integrate SDK RPC Server and Admin RPC Server to Single RPC Server

krapie:

Integrate SDK RPC server and Admin RPC server to single RPC server.

There are two reasons for this changes:

- To make both SDK and admin RPC route to same server, and process same data on the same server.
- To simplify API url and provide unified url on both SDK and admin APIs.

Fixes #

```release-note
Because API url will be unified to https://api.yorkie.dev for both SDK and admin,
it will affect all other repositories that uses both SDK and admin API for their service (eg: dashboard).
```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


server/rpc/interceptors/context.go
```diff
@@ -83,33 +87,42 @@ func (i *ContextInterceptor) Stream() grpc.StreamServerInterceptor {
...
+				)
+			}
 		}
 
 		return err
 	}
 }
 
+func isYorkieService(method string) bool {
+	return method[:len("/yorkie.v1.YorkieService")] == "/yorkie.v1.YorkieService"
```
hackerwins, line 123: Can we use the method below?

https://pkg.go.dev/strings#HasPrefix

hackerwins (approved):
Thanks for your contribution.

We should test this PR using the dashboard in minikube.

hackerwins:
@krapie We should replace the remaining 11103 port.
https://github.com/search?q=repo%3Ayorkie-team%2Fyorkie+11103&type=code

hackerwins:
@krapie It would be better to deploy all after merging the tasks that combine the Admin and SDK ports.

PR #545, opened 2023-06-07
Support UTF16 Code Units in yorkie.Tree

krapie:

Support UTF16 code units with rune in yorkie.Tree.

This is to support yorkie.Tree to handle multilingual.

Because previous yorkie.Tree did not support this, there was a bug when splitting text nodes that contains other languages like Korean (eg: splitting text node `한글` with offset of `2` resulted something like `�`, `글`, `��`). 

And because of this, tree structure got corrupted, leading to misbehavior of tree edit operation and failure of snapshot(bytes) conversion.

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/api/converter/converter_test.go
+++ b/api/converter/converter_test.go
@@ -113,6 +113,16 @@ func TestConverter(t *testing.T) {
 				Increase(10).
 				Increase(math.MaxInt64)
 
+			// tree
+			root.SetNewTree("k5").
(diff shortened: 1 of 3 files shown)


hackerwins (approved):
Thanks for your contribution.

Issue #544, opened 2023-06-07
Simplify CLI commands by saving rpc-addr and project to context


hackerwins:
**Description**:

Simplify CLI commands by saving rpc-addr and project to context

Currently, we need to pass the rpc-addr and project for every command.
For example to list documents:
```sh
$ yorkie document ls default --rpc-addr localhost:11101
```

We can simplify commands using the config command:

```sh
// save the context of `localhost:11101` to `~/.yorkie/config.json`
$ yorkie login -u admin -p admin --insecure localhost:11101
$ yorkie document ls codepair

// save the context of api.yorkie.dev to `~/.yorkie/config.json`
$ yorkie login -u admin -p admin api.yorkie.dev:443
$ yorkie document ls whiteboard

// list the contexts from `~/.yorkie/config.json`
$ yorkie context ls

// switch the context to `localhost:11101`
$ yorkie context set localhost:11101

// delete the current context from `~/.yorkie/config.json`
$ yorkie logout
```

Refer to https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-context-and-configuration

**Why**:

Wu22e:
I'm interested in this issue. Can I work on it?

hackerwins:
@Wu22e Of course. If you have any questions, feel free to ask.

PR #548, opened 2023-06-13
Add TLS Option & Insecure Flag in Admin CLI

krapie:

Add TLS option & insecure flag in admin CLI.

This is to set TLS connectivity between hosted server (`api.yorkie.dev`) and admin CLI.
Also, `--insecure` global flag is added to disable the TLS connection of the client when needed.

Fixes https://github.com/yorkie-team/devops/issues/43

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/admin/client.go
+++ b/admin/client.go
@@ -22,8 +22,11 @@ import (
 	"fmt"
 	"strings"
 
+	"crypto/tls"
+
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/metadata"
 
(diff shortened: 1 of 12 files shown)


server/rpc/server_test.go
```diff
@@ -735,7 +735,7 @@ func TestAdminRPCServerBackend(t *testing.T) {
 				Password: invalidSlugName,
 			},
 		)
-		assert.Equal(t, codes.Unauthenticated, status.Convert(err).Code())
+		assert.Error(t, err)
```
hackerwins, line 738: Is there any reason to remove the status code check?

admin/client.go
```diff
@@ -22,8 +22,11 @@ import (
 	"fmt"
 	"strings"
 
+	"crypto/tls"
```
hackerwins, line 25:
```suggestion
	"crypto/tls"
```

hackerwins (approved):
Thanks for your contribution. Overall, it looks good.

After this PR, referring to ArgoCD CLI might be helpful for Yorkie CLI. For instance, it would be beneficial to change the Login command to receive the address as an argument instead of a flag, similar to how it is done in the ArgoCD CLI, and store it in the config.

https://github.com/yorkie-team/yorkie/issues/544

server/rpc/server_test.go
krapie: It was for testing purpose, it should not be included.

I have reverted the code.

hackerwins (approved):
LGTM. 👍

PR #566, opened 2023-06-28
Apply garbage collection for tree

JOOHOJANG:
- Apply garbage collection for tree
- Fix some issues
  - miscalculation in `pathToTreePos`
  - null point exception when `Node.Parent` is `nil`

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/pkg/document/crdt/tree.go
+++ b/pkg/document/crdt/tree.go
@@ -19,6 +19,7 @@ package crdt
 import (
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"unicode/utf16"
 
@@ -248,9 +249,10 @@ func (n *TreeNode) DeepCopy() *TreeNode {
 // Tree represents the tree of CRDT. It has doubly linked list structure and
 // index tree structure.
 type Tree struct {
-	DummyHead    *TreeNode
-	IndexTree    *index.Tree[*TreeNode]
-	NodeMapByPos *llrb.Tree[*TreePos, *TreeNode]
+	DummyHead      *TreeNode
+	IndexTree      *index.Tree[*TreeNode]
+	NodeMapByPos   *llrb.Tree[*TreePos, *TreeNode]
+	removedNodeMap map[string]*TreeNode
 
 	createdAt *time.Ticket
 	movedAt   *time.Ticket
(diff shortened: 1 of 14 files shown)


pkg/document/crdt/tree.go
```diff
@@ -433,7 +498,7 @@ func (t *Tree) Edit(from, to *TreePos, content *TreeNode, editedAt *time.Ticket)
 				}
 			}
 		} else {
-			if fromPos.Node.Parent.Value.IsRemoved() {
+			if fromPos.Node.Parent != nil && fromPos.Node.Parent.Value.IsRemoved() {
```
hackerwins, line 501: May I know in which case the problem occurs?

I guess that `fromPos.node` is the root node, I'm wondering if we can just ignore it or if we need to do something else.

hackerwins (commented):
Thanks for your contribution.
I left a simple question.

pkg/document/crdt/tree.go
JOOHOJANG: As you said, we need this line of code when fromPos.Node is root of the tree. 
I think no other process is needed

pkg/document/crdt/tree.go
```diff
@@ -282,6 +285,64 @@ func (t *Tree) Marshal() string {
 	return builder.String()
 }
 
+// returns the length of removed nodes
```
hackerwins, line 288:
```suggestion
// removedNodesLen returns the length of removed nodes.
```

test/integration/gc_test.go
```diff
@@ -20,6 +20,7 @@ package integration
 
 import (
 	"context"
+	"github.com/yorkie-team/yorkie/pkg/document/time"
```
hackerwins, line 23: Could you put this import to the below group?

https://github.com/uber-go/guide/blob/master/style.md#import-group-ordering

pkg/index/tree.go
```diff
@@ -93,6 +93,18 @@ func postOrderTraversal[V Value](node *Node[V], callback func(node *Node[V], dep
 	callback(node, depth)
 }
 
+// postOrderTraversalAll traverses the whole tree (include tombstones) with postorder traversal.
+func postOrderTraversalAll[V Value](node *Node[V], callback func(node *Node[V], depth int), depth int) {
```
hackerwins, line 97:
```suggestion
func postorderTraversalAll[V Value](node *Node[V], callback func(node *Node[V], depth int), depth int) {
```

Previous codes also did, but I think we could replace `postOrder` with `postorder`.

hackerwins (approved):
Thanks for your reply.

I only left minor style comments.

PR #570, opened 2023-07-03
Remove panic from code

JOOHOJANG:
- Remove panic from crdt/tree, index/tree, element-rht and let them return error when it happens
- let json/... handle error

Related to #497

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins (approved):
Thanks for your contribution.

I did 3 additional minor code cleanups below.

1. We can omit the error checking if the interface of return of caller and method are the same and called it at the end.

AS-IS
```go
func (n *TreeNode) Append(newNodes ...*TreeNode) error {
        // ...
	err := n.IndexTreeNode.Append(indexNodes...)
	if err != nil {
		return err
	}

	return nil
}
```

TO-BE
```go
func (n *TreeNode) Append(newNodes ...*TreeNode) error {
        // ...
	return n.IndexTreeNode.Append(indexNodes...)
}
```

2. We can shorten it by binding the error checking with the method call.

https://www.digitalocean.com/community/tutorials/handling-errors-in-go#handling-errors

3. I removed panic from buildDescendants, because it has an interface with an error return.

PR #581, opened 2023-07-14
Remove unneccesary codes from gc

JOOHOJANG:
- According to https://github.com/yorkie-team/yorkie-js-sdk/pull/568, remove unnecessary code (no need to traverse indexTree during gc) 

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Didn't break anything

### DIFF
--- a/pkg/document/crdt/tree.go
+++ b/pkg/document/crdt/tree.go
@@ -324,27 +324,10 @@ func (t *Tree) purgeRemovedNodesBefore(ticket *time.Ticket) (int, error) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


hackerwins:
In the process of GC in Tree, when using `indexTree.TraverseAll`, there is a situation where the next node is skipped when performing a splice within the loop for (`const node of node._children`). This happens because the indexes are shifted one by one.

This commit fixes the issue.

For example, if the array is `[node A, node B, ...]` and we remove `A` while the index is pointing to `A`, it will skip `B` and move on to the next node.

hackerwins (approved):

PR #542, opened 2023-06-01
Move Presence from Client to Document  

chacha912:
**What this PR does / why we need it**:

Move Presence from Client to Document

### UpdatePresence

The UpdatePresence API has been removed, and the document and presence changes are now sent through the PushPull API. As a result, the presence is no longer passed as part of the Client, but instead included within the change. The changePack now includes the snapshotPresence.

- as-is
image

- to-be
image

### WatchDocument

In the WatchDocument API, the request and response only include the client ID without the presence.

wat

**Which issue(s) this PR fixes**:

Fixes #442

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
- The UpdatePresence API has been removed.
- Presence changes are now sent through the PushPull API, so the Change includes presence information and the ChangePack includes the snapshotPresence.
- In the WatchDocument API's request and response, only the client ID is passed without the presence.
```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Added relevant tests or not required
- [ ] Didn't break anything

Issue #442, opened 2023-01-11
Extract `Room` from `Document` and move `Presence` from `Client` to `Room`


hackerwins:
**Description**:

Extract `Room` from `Document` and move `Presence` from `Client` to `Room`

AS-IS
<img width="387" alt="Screenshot 2023-01-11 at 9 41 54 AM" src="https://user-images.githubusercontent.com/2059311/211691588-c571f5c1-b3e9-4498-aabb-527d3f18e179.png">

TO-BE
<img width="530" alt="Screenshot 2023-01-11 at 9 41 58 AM" src="https://user-images.githubusercontent.com/2059311/211691594-34493a42-89e9-4500-9e8c-48769c9de452.png">

Related to 
- https://github.com/yorkie-team/yorkie-js-sdk/issues/379
- https://github.com/yorkie-team/yorkie-js-sdk/issues/302

**Why**:

- Topics cannot be separated if a client attaches two different documents.
- We can use transactions by combining the document and the presence. This can reduce the workload by representing `Text`'s selection as presence because presence is cheaper than Document(CRDT).

hackerwins:
Move Presence to Document
- [x] Go: https://github.com/yorkie-team/yorkie/pull/582
- [x] JS SDK: https://github.com/yorkie-team/yorkie-js-sdk/pull/574
- [x] iOS SDK: https://github.com/yorkie-team/yorkie-ios-sdk/pull/94
- [x] Android SDK: https://github.com/yorkie-team/yorkie-android-sdk/pull/130

PR #582, opened 2023-07-14
Move Presence to Document

chacha912:

Move Presence from Client to Document

### UpdatePresence

The UpdatePresence API has been removed, and the document and presence changes are now sent through the PushPull API. As a result, the presence is no longer passed as part of the Client but instead included within the change. The changePack now includes the presence.

- as-is
image

- to-be
image

### WatchDocument

In the WatchDocument API, the request and response only include the client ID without the presence.

wat

Address #442

This PR is one of the separations from #542.

```release-note
- The UpdatePresence API has been removed.
- Presence changes are now sent through the PushPull API, so the Change includes presence information and the ChangePack includes the snapshotPresence.
- In the WatchDocument API's request and response, only the client ID is passed without the presence.
```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins (approved):
Thank you for your contribution. 🚀

PR #586, opened 2023-07-22
feat: cache project info

### LINKED ISSUE #568: Cache ProjectInfo to Improve Overall RPC Response Time
**Description**:

Cache `ProjectInfo` to improve overall RPC response time.

**Why**:

Every time RPC is called, `buildContext()` in the RPC interceptor is called before the RPC handler to build context data, which includes project metadata for RPC.

https://github.com/yorkie-team/yorkie/blob/aa2f3618f8fc0a53cfdca6717b37a6dfa1eeadd0/server/rpc/interceptors/context.go#L153

Currently, `ProjectInfo` is queried from DB every time when `buildContext()` is called. And it is inefficient to query the same data every time the RPC is called.

By introducing caching for `ProjectInfo`, we can dramatically improve query time for building context data, resulting faster RPC response time.

blurfx:

Cache ProjectInfo to improve response time.

For caching I used the existing LRU cache implementation because it seems there is no external cache layer like redis.

In cluster mode, each node has its own cache, so there is still a possibility of query requests for same project info.

Fixes #568

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/server/rpc/interceptors/context.go
+++ b/server/rpc/interceptors/context.go
@@ -19,7 +19,12 @@ package interceptors
 
 import (
 	"context"
+	"fmt"
 	"strings"
+	"time"
+
+	"github.com/yorkie-team/yorkie/pkg/cache"
+	"github.com/yorkie-team/yorkie/server/logging"
 
 	grpcmiddleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	"google.golang.org/grpc"
@@ -34,15 +39,24 @@ import (
 	"github.com/yorkie-team/yorkie/server/rpc/metadata"
 )
 
+const projectInfoCacheSize = 1
+const projectInfoCacheTTL = time.Hour
+
 // ContextInterceptor is an interceptor for building additional context.
 type ContextInterceptor struct {
-	backend *backend.Backend
+	backend          *backend.Backend
+	projectInfoCache *cache.LRUExpireCache[string, *types.Project]
 }
 
 // NewContextInterceptor creates a new instance of ContextInterceptor.
 func NewContextInterceptor(be *backend.Backend) *ContextInterceptor {
+	projectInfoCache, err := cache.NewLRUExpireCache[string, *types.Project](projectInfoCacheSize)
+	if err != nil {
+		logging.DefaultLogger().Fatal("Failed to create project info cache: %v", err)
+	}
 	return &ContextInterceptor{
-		backend: be,
+		backend:          be,
+		projectInfoCache: projectInfoCache,
 	}
 }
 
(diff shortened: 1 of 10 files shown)


server/rpc/interceptors/context.go
```diff
@@ -34,15 +39,24 @@ import (
 	"github.com/yorkie-team/yorkie/server/rpc/metadata"
 )
 
+const projectInfoCacheSize = 1
```
hackerwins, line 42: It looks like the size of the cache is 1, so the cache only holds one project in a time. I wonder if it is intended.

hackerwins (commented):
Thanks for your contribution. I left a small comment.

server/rpc/interceptors/context.go
blurfx: Yes it's intended, I understood that a each project can have one ProjectInfo, isn't it?

blurfx: No it's not intended. I had it set to 256, but I think I accidentally changed it. 

and here's something I'd like to discuss. how much CacheSize and TTL are appropriate?

I was going to set the CacheSize to 256 and the TTL to 1 hour, but what do you think?

hackerwins: I also think 256 and 1 hour are suitable for the current situation. This is because yorkie.dev doesn't have many projects yet and project updates don't affect requests from clients.

blurfx:
@hackerwins 

Integration test `auth_webhook_test.go` fails after the change.

It passes when testing only the `Selected method authorization webhook test` test case, but fails when testing the entire test suite and the `authorization webhook test` test is run first.

I tried to fix it, but there's something I don't understand. If using an invalid token when activating the client below, shouldn't we get an Unauthenticated error?

https://github.com/yorkie-team/yorkie/blob/dba636f8af25b9da8e941a3b12330da865d0440c/test/integration/auth_webhook_test.go#L164-L173

hackerwins (commented):
Thank you for applying my requests.

The reason for CI failure

The purpose of the `Selected method authorization webhook test` is to verify that the interceptor only checks authentication and authorization of specific methods. In this test case, the selected methods are `AttachDocument` and `WatchDocuments`. As a result, the `ActivateClient` method passes even with an invalid token.

After implementing a cache for projects, the project information can be cached from the previous test. Consequently, the project info may not be immediately updated in the current test. So the test failed.

We should take into cache eviction options and modify the test to wait for a certain period until it gets evicted by TTL.

blurfx:
in 576e3c1, I moved the `ProjectInfoCacheSize` and `ProjectInfoCacheTTL` to the config. therefore, I added the default values and also added them to the sample config.

now test backend has 5 seconds `ProjectInfoCacheTTL`, and I modified the existing test to do `time.Sleep(5 * time.Seconds)`.

krapie (approved):
> In cluster mode, each node has its own cache, so there is still a possibility of query requests for same project info.

Since we are using sharded cluster mode, I think we will have a meaningful cache hit rate for querying the same project info.

LGTM!

hackerwins (commented):
Thanks for applying my requests. I left one more question about the cache key.

> Since we are using sharded cluster mode, I think we will have a meaningful cache hit rate for querying the same project info.

We have two types of APIs: project-wide APIs and document-wide APIs. For document-wide APIs, requests are shared based on the document key, causing requests to be evenly distributed across nodes for a project. So it seems likely that caches for all projects will exist on each node.

server/rpc/interceptors/context.go
```diff
@@ -146,15 +155,19 @@ func (i *ContextInterceptor) buildContext(ctx context.Context) (context.Context,
 		md.Authorization = authorization[0]
 	}
 	ctx = metadata.With(ctx, md)
+	cacheKey := fmt.Sprintf("%s:%s", md.APIKey, md.Authorization)
```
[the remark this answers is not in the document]
blurfx: I used the authorization header by misunderstanding. so yes, cache key should be just apikey without auth header. I’ll fix it.

PR #587, opened 2023-07-22
Add Pagination to Listing Projects for Housekeeping

### LINKED ISSUE #569: Introduce Pagination for Housekeeping
**Description**:

Introduce pagination for housekeeping.

**Why**:

When performing housekeeping, all `ProjectInfo`s are loaded from DB.

https://github.com/yorkie-team/yorkie/blob/aa2f3618f8fc0a53cfdca6717b37a6dfa1eeadd0/server/backend/database/mongo/client.go#L240

This can cause severe performance issue as the number of projects in DB grows.

To avoid this, we can introduce pagination for projects and perform housekeeping in smaller portions.

tedkimdev:

To avoid all project info being loaded from DB when doing housekeeping, pagination is added for listing projects and performing housekeeping in smaller portions.

Fixes #569 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/server/backend/database/database.go
+++ b/server/backend/database/database.go
@@ -138,6 +138,8 @@ type Database interface {
 	FindDeactivateCandidates(
 		ctx context.Context,
 		candidatesLimitPerProject int,
+		projectFetchSize int,
+		housekeepingLastProjectID *types.ID,
 	) ([]*ClientInfo, error)
 
 	// FindDocInfoByKey finds the document of the given key.
--- a/server/backend/database/database.go
+++ b/server/backend/database/database.go
@@ -139,8 +139,8 @@ type Database interface {
 		ctx context.Context,
 		candidatesLimitPerProject int,
 		projectFetchSize int,
-		housekeepingLastProjectID *types.ID,
-	) ([]*ClientInfo, error)
+		housekeepingLastProjectID types.ID,
+	) (types.ID, []*ClientInfo, error)
 
 	// FindDocInfoByKey finds the document of the given key.
 	FindDocInfoByKey(
--- a/server/backend/database/database.go
+++ b/server/backend/database/database.go
@@ -139,7 +139,7 @@ type Database interface {
 		ctx context.Context,
 		candidatesLimitPerProject int,
 		projectFetchSize int,
-		housekeepingLastProjectID types.ID,
+		lastProjectID types.ID,
 	) (types.ID, []*ClientInfo, error)
 
 	// FindDocInfoByKey finds the document of the given key.
--- a/server/backend/database/memory/database.go
+++ b/server/backend/database/memory/database.go
@@ -18,6 +18,7 @@
 package memory
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	gotime "time"
@@ -224,16 +225,15 @@ func (d *DB) CreateProjectInfo(
 	return info, nil
 }
 
-// ListAllProjectInfos returns all project infos.
-func (d *DB) listAllProjectInfos(
+// listProjectInfos returns all project infos rotationally.
+func (d *DB) listProjectInfos(
 	ctx context.Context,
+	pageSize int,
+	housekeepingLastProjectID *types.ID,
 ) ([]*database.ProjectInfo, error) {
 	txn := d.db.Txn(false)
 	defer txn.Abort()
 
-	// TODO(krapie): txn.Get() loads all projects in memory,
-	// which will cause performance issue as number of projects in DB grows.
-	// Therefore, pagination of projects is needed to avoid this issue.
 	iter, err := txn.Get(
 		tblProjects,
 		"id",
@@ -242,12 +242,35 @@ func (d *DB) listAllProjectInfos(
 		return nil, fmt.Errorf("fetch all projects: %w", err)
 	}
 
-	var infos []*database.ProjectInfo
-	for raw := iter.Next(); raw != nil; raw = iter.Next() {
-		info := raw.(*database.ProjectInfo).DeepCopy()
-		infos = append(infos, info)
+	lastIDBytes, err := housekeepingLastProjectID.Bytes()
+	if err != nil {
+		return nil, fmt.Errorf("decode last project id: %w", err)
 	}
 
+	var infos []*database.ProjectInfo
+	for i := 0; i < pageSize; {
+		raw := iter.Next()
+		if raw == nil {
+			*housekeepingLastProjectID = database.DefaultProjectID
+			break
+		}
+
+		info := raw.(*database.ProjectInfo).DeepCopy()
+
+		idBytes, err := info.ID.Bytes()
+		if err != nil {
+			return nil, fmt.Errorf("decode project id: %w", err)
+		}
+
+		if bytes.Compare(idBytes, lastIDBytes) > 0 {
+			infos = append(infos, info)
+			i++
+		}
+
+		if i == pageSize {
+			*housekeepingLastProjectID = infos[len(infos)-1].ID
+		}
+	}
 	return infos, nil
 }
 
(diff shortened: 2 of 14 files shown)


server/config.sample.yml
```diff
@@ -36,6 +36,9 @@ Housekeeping:
   # CandidatesLimitPerProject is the maximum number of candidates to be returned per project (default: 100).
   CandidatesLimitPerProject: 100
 
+  # HousekeepingProjectFetchSize is the maximum number of projects to be returned to deactivate candidates. (default: 100)
+  HousekeepingProjectFetchSize: 10
```
krapie, line 40: since default value is 100, it will be better to set this value 100.

krapie (changes requested):
Thank you for your contribution!
I left a small request for you.

Since `HousekeepingProjectFetchSize` option is newly introduced, we need to add this option to the CLI flag.

Ref: https://github.com/yorkie-team/yorkie/pull/512

server/backend/database/memory/database.go
```diff
@@ -36,7 +36,8 @@ import (
 
 // DB is an in-memory database for testing or temporarily.
 type DB struct {
-	db *memdb.MemDB
+	db                      *memdb.MemDB
+	housekeepingProjectPage int
```
hackerwins, line 40: DB instance is shared and used by several routines such as RPC handlers. Having a mutable value as a property in the instance seems to create other side effects.

How about moving `housekeepingProjectPage` into the housekeeping function?

hackerwins (changes requested):
Thanks for your contribution.
I left a comment.

server/backend/database/memory/database.go
tedkimdev: Thank you for the comment. I moved the variable into the run function.
Please let me know if I need to correct something!

hackerwins:
No worries at the moment since we have small projects in `https://yorkie.dev`, and the `offset/limit` isn't causing any performance issues.

However, as our services grow, this could become problematic when fetching the first page. To address this potential issue, we could consider changing the interface and implementing it with keyset pagination.

For more information on this approach, you can check out: https://use-the-index-luke.com/no-offset

tedkimdev:
@hackerwins I changed the database interface like you mentioned.
I did not call `SetSort()` when using `Find()` in the Mongodb client because the natural order refers to the default document ordering based on `_id`.
If I am not correct, please let me know.

server/backend/database/memory/database.go
```diff
@@ -242,12 +242,35 @@ func (d *DB) listAllProjectInfos(
...
 		info := raw.(*database.ProjectInfo).DeepCopy()
-		infos = append(infos, info)
-	}
 
+		idBytes, err := info.ID.Bytes()
+		if err != nil {
+			return nil, fmt.Errorf("decode project id: %w", err)
+		}
+
+		if bytes.Compare(idBytes, lastIDBytes) > 0 {
```
krapie, line 265: It seems like go-memdb does not provide an ordered result (iterator) for `txn.Get()`... right?
I could not find any reference to the fact that go-memdb provides ordered results.

tedkimdev: I think you are right. I will add iterator options to sort by id before iterating.

server/backend/housekeeping/housekeeping.go
```diff
@@ -127,9 +108,11 @@ func (h *Housekeeping) Stop() error {
 
 // run is the housekeeping loop.
 func (h *Housekeeping) run() {
+	housekeepingLastProjectID := database.DefaultProjectID
```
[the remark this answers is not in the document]
tedkimdev: It depends. If it needs to share the variable across different parts of an application, it would be better to put it with a mutex in a struct. Also, if we need to share the variable across all instances in a clustered environment, a good approach is to use a distributed key-value store. Using a variable within the function's scope is like it doesn't share the variable across an application.

What do you think is the best in the situation here?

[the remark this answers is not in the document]
krapie: Then I think this variable is best fit in here, since this variable only effects  housekeeping's internal `deactivateCandidates` function and corresponding database `FindDeactivateCandidates` function.

Thank you for sharing this!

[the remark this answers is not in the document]
hackerwins: I believe that when two methods share a variable, they become dependent on each other, so it would be good to avoid such dependencies whenever possible. One alternative approach could be passing the variable as a value and updating it directly from the caller's side. What do you think about this idea?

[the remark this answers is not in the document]
tedkimdev: Could you see the code changes in the last commit? I am not sure I understand you correctly.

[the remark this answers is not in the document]
krapie: @tedkimdev I think your implementation is correct.
Can you review this? @hackerwins

server/backend/database/memory/database.go
```diff
@@ -243,12 +244,47 @@ func (d *DB) listAllProjectInfos(
...
+			break
+		}
+	}
+
+	lastIndex := beginIndex + pageSize - 1
+	if lastIndex >= len(infos) {
+		lastIndex = len(infos) - 1
+	}
+
+	return infos[beginIndex : lastIndex+1], nil
```
hackerwins, line 287: How about using an index to avoid fetching all projects from DB? MemDB also provides indexes like other DB.

You can refer to the below about how to use indexes in MemDB:
- https://github.com/yorkie-team/yorkie/blob/16fd182021231d75562a933cb32d924af16fc7f4/server/backend/database/memory/database.go#L913-L942
- https://github.com/hashicorp/go-memdb#example

hackerwins (changes requested):
Thanks for applying my requests.

I left one more comment about using indexes in MemDB.

server/backend/database/memory/database.go
tedkimdev: Thank you for the feedback. I didn't notice that in MemDB. I added changes to use an index.

server/backend/database/database.go
```diff
@@ -138,7 +138,9 @@ type Database interface {
 	FindDeactivateCandidates(
 		ctx context.Context,
 		candidatesLimitPerProject int,
-	) ([]*ClientInfo, error)
+		projectFetchSize int,
+		housekeepingLastProjectID types.ID,
```
hackerwins, line 142: In a layered architecture, it is recommended that the callee doesn't know about the caller.

`FindDeactivateCandidates` is only called by `housekeeping` package for now but `database` can be referenced by other packages too. And I think `housekeepingLastProjectID` has a semantic dependency on the caller, `housekeeping` package.

How about removing `housekeeping` prefix from the variable in `database` package?

server/backend/database/memory/housekeeping_test.go
```diff
@@ -59,14 +63,71 @@ func TestHousekeeping(t *testing.T) {
...
+			fetchSize,
+			lastProjectID,
+		)
+		assert.NoError(t, err)
+		assert.Equal(t, database.DefaultProjectID, lastProjectID)
+	})
+}
+
+func createDBandProjects(t *testing.T) (*memory.DB, []*database.ProjectInfo) {
+	t.Helper()
```
hackerwins, line 114: This is the first time to see `t.Helper`. When is this useful? Would it be good to add `t.Helper` to other helper functions too?

hackerwins (commented):
Overall looks good.
I left minor style-related comments.

server/backend/housekeeping/housekeeping.go
```diff
@@ -127,11 +108,15 @@ func (h *Housekeeping) Stop() error {
...
+
 	for {
 		ctx := context.Background()
-		if err := h.deactivateCandidates(ctx); err != nil {
+		lastProjectID, err := h.deactivateCandidates(ctx, housekeepingLastProjectID)
+		if err != nil {
```
hackerwins, line 116: It was missed in the previous codes too, but it would be good to log the error.

server/backend/database/memory/housekeeping_test.go
tedkimdev: > `t.Helper()` indicates to the Go test runner that `createDBandProjects()` helper function is a test helper. It makes it easy to report test failures at the call site of the parent function rather than the call site of `t.Error()` and `t.Errorf()`. When t.Error() is called from `createDBandProjects()` function, the Go test runner will report the filename and line number of the code which called the `createDBandProjects()` function in the output.

hackerwins (approved):
LGTM. 👍

Issue #583, opened 2023-07-19
Replace Text.select with Presence


hackerwins:
**Description**:

Recently we introduced batch updates of documents and presences to provide atomic change delivery and events from `document.subscribe`. All operations and presence changes created in `document.update` guarantee atomic delivery, So we can replace `Text.select` with presence and delete the method for now.

```ts
// Before
doc.update((root) => {
  root.text = new yorkie.Text();            // {"text":""}
  root.text.edit(0, 0, 'hello');            // {"text":"hello"}
  root.text.select(0, 1);                   // {"text":"^h^ello"}
});

// After
doc.update((root, presence) => {
  root.text = new yorkie.Text();                          // {"text":""}
  root.text.edit(0, 0, 'hello');                          // {"text":"hello"}
  presence.set('selection', root.text.createRange(0, 1)); // {"text":"^h^ello"}
});
```

We need to do the below:

- [x] Provide `Text.createRange`
- [x] Replace `Text.select` with presence in all Text examples
- [x] Remove `Select` operation from `Text`

Batch updates of documents and presences:
- Server and Go SDK: https://github.com/yorkie-team/yorkie/pull/582
- JS SDK: https://github.com/yorkie-team/yorkie-js-sdk/pull/574

**Why**:
- Clean up unnecessary features to keep the product simple.

joonhyukchoi:
I'm interested in this issue. can i work on it?

krapie:
@joonhyukchoi let's give it a try!

chacha912:
@joonhyukchoi  Yes. When making that changes to protobuf, there can be compatibility issues with previous versions. In such cases, we have been labeling issues and PRs with `protocol changed📝` to indicate the compatibility impact.
(Since there have been no real production yet, we have been working on it with breaking version compatibility.)

joonhyukchoi:
@chacha912 Thank you. I'll follow that!

hackerwins:
Tasks:

- [x] Go SDK: https://github.com/yorkie-team/yorkie/pull/589
- [ ] JS SDK:
- [ ] Android SDK:
- [ ] iOS SDK:

blurfx:
I'll work on the js sdk

PR #589, opened 2023-07-26
Remove Select operation from Text

joonhyukchoi:

Recently we moved presence to document from client and the role of Text.select has become replaceable. So we decided to remove Select operation from Text. 

Address #583 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins (commented):
Thanks for your contribution.

I left a comment.
https://github.com/yorkie-team/yorkie/issues/583#issuecomment-1652792682

hackerwins:
The migration appears to be challenging due to operations being stored in a binary format. To accomplish this task, we would need to retrieve operations from DB, unmarshal them using the previous version of the protobuf message and then marshal them again with the new version before storing them back in DB.

To avoid the migration process, how about retaining the orders in the protobuf message as they are?

hackerwins (approved):
LGTM. 👍

PR #594, opened 2023-08-01
Clean up methods related to presence

chacha912:

- Return presence to user by retrieving only online clients and making a copy
- Trigger unwatched event instead of presence-changed event during detach

Related https://github.com/yorkie-team/yorkie-js-sdk/pull/599

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/pkg/document/internal_document.go
+++ b/pkg/document/internal_document.go
@@ -283,34 +283,56 @@ func (d *InternalDocument) ApplyChanges(changes ...*change.Change) ([]DocEvent, 
 
 // MyPresence returns the presence of the actor currently editing the document.
 func (d *InternalDocument) MyPresence() innerpresence.Presence {
-	p := d.presences.LoadOrStore(d.changeID.ActorID().String(), innerpresence.NewPresence())
+	if d.status != StatusAttached {
+		return innerpresence.NewPresence()
+	}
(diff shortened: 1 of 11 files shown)


pkg/document/internal_document.go
```diff
@@ -283,34 +295,56 @@ func (d *InternalDocument) ApplyChanges(changes ...*change.Change) ([]DocEvent,
...
+		}
+		presences[key.(string)] = p.DeepCopy()
+		return true
+	})
+	return presences
+}
+
+// AllPresences returns the presence map of all clients
+// regardless of whether the client is online or not.
+func (d *InternalDocument) AllPresences() map[string]innerpresence.Presence {
```
hackerwins, line 337: Do we need to change `AllPresences` to `PresencesForTest`?

hackerwins (approved):
Overall looks good. I left a minor question.

pkg/document/internal_document.go
chacha912: I named it `AllPresences` because it is used to retrieve all presences when creating snapshot(for non-testing purposes).

hackerwins: It makes sense. 👍

PR #597, opened 2023-08-03
Fetch latest snapshot metadata to determine snapshot creation need

### LINKED ISSUE #387: Introduce repository method for only fetching SnapshotInfo's metadata
Introduce new repository method which only fetching SnapshotInfo's metadata
The SnapshotInfo's metadata that i'm suggesting contains fields except for 'Snapshot' field

**Why is this needed**:
Snapshot field of SnapshotInfo would be a large field, As document getting bigger and bigger.
large SnapShot field would require many Network bandwidth and I/O time

snapshots.go#storeSnapshot always fetch every field of snapshotInfo

https://github.com/yorkie-team/yorkie/blob/50d8722d50b2e2ab48cdc59d00b296042ddaf66b/server/packs/snapshots.go#L36-L38

But sadly, 'SnapShot' field of SnapshotInfo is necessary, only when a below conditions are satisfied.

https://github.com/yorkie-team/yorkie/blob/50d8722d50b2e2ab48cdc59d00b296042ddaf66b/server/packs/snapshots.go#L42-L47

To sum up, I want to suggest separating SnapShotInfo related repository into two parts. 
one is a method that fetching only metadata of SnapshotInfo 
another is a method that fetching only Snapshot field of SnapshotInfo

hyemmie:
To reduce unnecessary data I/O between DB and server, added repository function which only fetches snapshot's metadata.

Fixes #387 

```release-note
NONE
```

**Additional documentation**:
The timing of fetching metadata and fulldata is different, so it needs to check if system needs to keep atomicity. I'm not sure if the two data will be the same because I haven't checked the cycle of saving snapshots yet.

 
https://github.com/hyemmie/yorkie/blob/fetching-metadata/server/packs/snapshots.go#L60
```Go
        // 01. get the closest snapshot of this docInfo
	snapshotMetadata, err := be.DB.FindClosestSnapshotMetadata(ctx, docInfo.ID, docInfo.ServerSeq)
	if err != nil {
		return err
	}
	if snapshotMetadata.ServerSeq == docInfo.ServerSeq {
		return nil
	}
	if docInfo.ServerSeq-snapshotMetadata.ServerSeq < be.Config.SnapshotInterval {
		return nil
	}

	// 02. retrieve the changes between last snapshot and current docInfo
	changes, err := be.DB.FindChangesBetweenServerSeqs(
		ctx,
		docInfo.ID,
		snapshotMetadata.ServerSeq+1,
		docInfo.ServerSeq,
	)
	if err != nil {
		return err
	}

	// 03. create document instance of the docInfo
	// TODO: check policy to guarantee atomicity
	snapshotInfo, err := be.DB.FindClosestSnapshotFullData(ctx, docInfo.ID, docInfo.ServerSeq)
	if err != nil {
		return err
	}
	doc, err := document.NewInternalDocumentFromSnapshot(
		docInfo.Key,
		snapshotInfo.ServerSeq,
		snapshotInfo.Lamport,
		snapshotInfo.Snapshot,
	)
	if err != nil {
		return err
	}

	pack := change.NewPack(
		docInfo.Key,
		change.InitialCheckpoint.NextServerSeq(docInfo.ServerSeq),
		changes,
		nil,
	)
	pack.MinSyncedTicket = minSyncedTicket

	if err := doc.ApplyChangePack(pack); err != nil {
		return err
	}

...
}


```

**Checklist**:
- [x] Didn't break anything


hackerwins (approved):
Thanks for your contribution.

PR #598, opened 2023-08-04
Remove panic from crdt counter

MoonGyu1:

Remove the panic methods in `crdt/counter.go`
- Before, there were panic methods to reveal errors in development environment, but because it risks stopping the server in production, so those panics were replaced with error methods.
- As the panic methods in counter were replaced, the related test codes also were modified.

Related to [#497](https://github.com/yorkie-team/yorkie/issues/497#issuecomment-1617477643) (`crdt/counter`)

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/pkg/document/crdt/counter.go
+++ b/pkg/document/crdt/counter.go
@@ -33,16 +33,16 @@ const (
 )
 
 // CounterValueFromBytes parses the given bytes into value.
-func CounterValueFromBytes(counterType CounterType, value []byte) interface{} {
+func CounterValueFromBytes(counterType CounterType, value []byte) (interface{}, error) {
 	switch counterType {
 	case IntegerCnt:
 		val := int32(binary.LittleEndian.Uint32(value))
-		return int(val)
+		return int(val), nil
 	case LongCnt:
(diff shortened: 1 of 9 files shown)


pkg/document/crdt/counter.go
```diff
@@ -33,16 +33,16 @@ const (
...
 	switch counterType {
 	case IntegerCnt:
 		val := int32(binary.LittleEndian.Uint32(value))
-		return int(val)
+		return int(val), nil
 	case LongCnt:
-		return int64(binary.LittleEndian.Uint64(value))
+		return int64(binary.LittleEndian.Uint64(value)), nil
+	default:
+		return nil, fmt.Errorf("unsupported type")
```
krapie, line 44: Since `unsupported type` error message is used in several other places, how about constantize this to something like `ErrUnsupportedType`?

krapie (dismissed):
Thank you for your contribution 👍🏼 

I left a small comment below.

krapie (changes requested):
Follow above comment 😅

hackerwins (approved):

PR #599, opened 2023-08-05
Set cobra default output to stdout

### LINKED ISSUE #490: Working in Linux pipe for yorkie CLI

Linux pipe is not working in yorkie CLI

```bash
$ ls -al | wc -l
 30

// pipe is not working. result is 0
$ ./bin/yorkie document ls default | wc -l
 ID                        KEY           ...
 6406074c3d6050a35caa0dbb  example10  ...
 6406074b3d6050a35caa0d8c  example9   ...
 640607493d6050a35caa0d4e  example8   ...
 640607473d6050a35caa0d0f  example7   ...
 640607463d6050a35caa0cd7  example6   ...
 0
```

**Why is this needed**:

User can use linux pipe

blurfx:

Fixes an issue that could not pipe output from the Yorkie CLI

Fixes #490 

cobra prints output to `os.Stderr` by default when calling the Print function internally. See https://github.com/spf13/cobra/blob/v1.7.0/command.go#L1349-L1351

So we can pipe the results if we explicitly tell it to output to Stdout like:
```
$ ./yorkie project ls --rpc-addr "api.yorkie.dev:443" | wc -l
2

$ ./yorkie project ls --rpc-addr "api.yorkie.dev:443" | awk 'NR>1 {print $1}'
myProject
```

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/cmd/yorkie/commands.go
+++ b/cmd/yorkie/commands.go
@@ -18,6 +18,8 @@
 package main
 
 import (
+	"os"
+
 	"github.com/spf13/cobra"
 
 	"github.com/yorkie-team/yorkie/cmd/yorkie/config"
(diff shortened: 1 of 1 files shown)


hackerwins (approved):
Thanks for your contribution.

Issue #36, opened 2020-03-30
Replace ActorID type with UUID or Long instead of String to reduce the size of CRDT meta


hackerwins:
The sizes of the types:

| Name        | Binary Size | String Size    | Features
|-------------|-------------|----------------|----------------
| [UUID]      | 16 bytes    | 36 chars       | configuration free, not sortable
| [shortuuid] | 16 bytes    | 22 chars       | configuration free, not sortable
| [Snowflake] | 8 bytes     | up to 20 chars | needs machine/DC configuration, needs central server, sortable
| [ObjectID]   | 12 bytes    | 24 chars  `<-- currently we use`      | configuration free, sortable
| xid         | 12 bytes    | 20 chars       | configuration free, sortable
| uint64         | 8 bytes    | N/A       | configuration free, needs central server, sortable

computerphilosopher:
Could I try this issue?

hackerwins:
@computerphilosopher Sure. :)

dc7303:
@hackerwins 
The current ActorID is checked by the `_id` value of the `clients` collection in mongoDB.
Are you thinking of adding the actorID field to the `clients` collection to fix this issue?

hackerwins:
@dc7303 

Yes. If we use a type other than `ObjectID`, we will need to add a new field to the `clients` collection. Looking at it again, choosing a `UUID` will increase its size more than it is now.

The sizes of the types listed in this issue are:
- [ObjectID](https://github.com/mongodb/mongo-go-driver/blob/master/bson/primitive/objectid.go#L29): 12 bytes
- [UUID](https://github.com/google/uuid/blob/master/uuid.go#L19): 16 bytes
- [uint64](https://golang.org/pkg/builtin/#uint64): 8 bytes

Even if we use `uint64`, the `actorID` is expected to decrease by only 33%. We need to calculate how much it reduces in the [snapshot](https://github.com/yorkie-team/yorkie/issues/9#issuecomment-605638867).

autumn-n:
@hackerwins Could I try this if still alive?

hackerwins:
@autumn-n Sure. If you have any questions, please feel free to comment.

autumn-n:
@hackerwins 

Thank you :) It's the first time with Golang, MongoDB, but I'll give it a try.

Before getting started, I'd like to hear your thoughts on the below.

1. Is it breaking backward compatibility, right?
[...]

PR #614, opened 2023-08-18
Cleanup proto

### LINKED ISSUE #498: Cleanup protobuf response
**Description**:

Remove unnecessary client information from the protobuf response.

We are currently returning `client_id` or `client_key` as follows:
https://github.com/yorkie-team/yorkie/blob/main/api/yorkie/v1/yorkie.proto#L41-L116

```
message ActivateClientResponse {
  string client_key = 1;
  bytes client_id = 2;
}

message DeactivateClientResponse {
  bytes client_id = 1;
}

message AttachDocumentResponse {
  bytes client_id = 1;
  string document_id = 2;
  ChangePack change_pack = 3;
}

message DetachDocumentResponse {
  string client_key = 1;
  ChangePack change_pack = 2;
}

message RemoveDocumentResponse {
  string client_key = 1;
  ChangePack change_pack = 2;
}

message PushPullChangesResponse {
  bytes client_id = 1;
  ChangePack change_pack = 2;
}
```

We only need to return the `client_id` when activating, and the rest seems unnecessary. 
It would be better to clean up.

**Why**:
to make protobuf concise

chacha912:

- Remove unnecessary client information from the protobuf response
- Change `clientID` type from `bytes` to `string` in the API 
  - Only the `actorID` in `changeID` needs to be kept as bytes: https://github.com/yorkie-team/yorkie/issues/36
  - but, if `clientID` needs to be maintained as bytes, it's possible to revert the commit, so please let me know.

Fixes #498 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


chacha912:
@hackerwins  Yes, your point is accurate. It would also reduce the need for frequent conversions between `bytes` and `string` during usage.

PR #615, opened 2023-08-18
Fix treePos calculating logic for text node

JOOHOJANG:
Fix `toTreePos` to calculate treePos of text node correctly

Related to https://github.com/yorkie-team/yorkie-js-sdk/pull/611

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Didn't break anything

### DIFF
--- a/pkg/document/crdt/tree.go
+++ b/pkg/document/crdt/tree.go
@@ -541,22 +541,6 @@ func (t *Tree) FindPos(offset int) (*TreePos, error) {
 			leftSibling = node.Parent.Value
 		} else {
 			leftSibling = node.Value
-			absOffset := node.Value.ID.Offset
-			split, err := node.Value.Split(offset, absOffset)
-			if err != nil {
-				return nil, err
-			}
-
-			if split != nil {
-				split.InsPrev = node.Value
-				t.NodeMapByID.Put(split.ID, split)
-
-				if node.Value.InsNext != nil {
-					node.Value.InsNext.InsPrev = split
-					split.InsNext = node.Value.InsNext
-				}
-				node.Value.InsNext = split
-			}
 		}
 		node = node.Parent
 	} else {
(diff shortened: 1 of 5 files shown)


pkg/document/crdt/tree.go
```diff
@@ -831,24 +831,22 @@ func (t *Tree) toTreePos(pos *TreePos) *index.TreePos[*TreeNode] {
...
-			offset, err = t.IndexTree.LeftSiblingsSize(parentNode.IndexTreeNode, offset)
+			treePos = &index.TreePos[*TreeNode]{
+				Node:   leftSiblingNode.IndexTreeNode,
+				Offset: leftSiblingNode.IndexTreeNode.PaddedLength(),
+			}
+		} else {
+			leftSiblingOffset, err := parentNode.IndexTreeNode.FindOffset(leftSiblingNode.IndexTreeNode)
 			if err != nil {
-				return nil // NOTE(sejongk): should return error instead?
+				return nil
```
hackerwins, line 842: Could we skip the return of the error?

hackerwins (changes requested):
Thanks for your contribution. I left a question.

pkg/document/crdt/tree.go
JOOHOJANG: I think it's safer to return error rather than skip it

let me fix

hackerwins:
When loading a tree with tombstones from a snapshot loaded from the server, the weight of the index tree is sometimes negative, even though there are active nodes in the tree.

Screenshot 2023-08-18 at 6 53 55 PM

This seems to be caused by duplicate counting, including nodes removed from removed ancestors.

sejongk:
@JOOHOJANG @hackerwins 
I changed toTreePos to consider the case when parentNode is removed.
Would you mind reviewing this logic?

hackerwins:
@sejongk Please add test code.
https://github.com/yorkie-team/yorkie-js-sdk/pull/611/commits/7780fb96f19c0b8c2aebaa0c51dae6abaadfe518#diff-1dc3229f2a08c07529137827aafba58e6a28cc6144bc561da58287baf71fae7eR211-R245

hackerwins (approved):
Thanks for your contribution.

PR #609, opened 2023-08-16
Update presence design document

chacha912:
**What this PR does / why we need it**:

Update design document for presence.
It contains information about the changes made in the following pull request.
- Move Presence to Document: 
  - https://github.com/yorkie-team/yorkie/pull/582
  - https://github.com/yorkie-team/yorkie-js-sdk/pull/574
  - https://github.com/yorkie-team/yorkie-js-sdk/pull/578

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

While working on the document, I think it is better to make the following changes. (I will address these changes through a separate pull request):
- Change "documents" to "document" in DocEvent (since the `WatchDocument` API has been updated to operate on a single document: https://github.com/yorkie-team/yorkie-js-sdk/pull/480).
- Rename `doc.getPresences()` to `doc.getUsers()` (as the returned value aligns more closely with user information, making this name easier to comprehend).

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Added relevant tests or not required
- [x] Didn't break anything

PR #616, opened 2023-08-18
Rename Presences to Users

chacha912:

Rename `doc.getPresences()` to `doc.getUsers()` as the returned value aligns more closely with user information, making this name easier to comprehend.

The reason `users` is used in `getUsers` is that using `clients` (as yorkie Client) might imply that it should hold all information about the yorkie client. However, users only contain clientID and presence information, so I used the term `users` instead of `clients.`

Related https://github.com/yorkie-team/yorkie/pull/609
https://github.com/yorkie-team/yorkie-js-sdk/pull/618

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/design/presence.md
+++ b/design/presence.md
@@ -55,7 +55,7 @@ doc.getMyPresence(); // { color: 'blue', cursor: { x: 1, y: 1 } }
 doc.getPresence(clientID);
 
 // Get all users currently participating in the document
-const users = doc.getPresences();
+const users = doc.getUsers();
 for (const { clientID, presence } of users) {
   // Do something...
 }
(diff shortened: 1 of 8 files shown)


hackerwins:
I think the term "getUsers" isn't the best choice, considering that a single user might be using multiple clients. Additionally, the SDK hasn't introduced the concept of a user just yet.

Will Presence be confused?

chacha912:
@hackerwins  I think that presence`'s'` wasn't clear. In the js-sdk, `getPresences()` returns an array of `{ clientID: ActorID; presence: P }`, which feels closer to a user. However, I agree that this needs more discussion since it's the first time expressing the concept of a user.
image

hackerwins:
I will close this PR for now. Please let me know when you proceed.

PR #622, opened 2023-08-21
Change TreeNode to have IDs instead of insPrev, insNext 

JOOHOJANG:
- Update protocol and change converter
  - pass `insPrev`, `insNext`
- Let `CRDTTreeNode` to have `insPrevID`, `insNextID` instead of `insPrev`,`insNext`
  - it's related to protocol change. 
  - during conversion of pb to tree nodes, it's hard to find insPrev/insNext node from id cause we can't use LLRB at that moment. Therefore, I changed it to keep id of insPrev and insNext and find actual node when we need it.
- Change `Prepend` to call `updateAncestorSize` only when node is not removed during conversion of pb to tree nodes
- Add test cases
  - to check split link is transmitted
  - to check calculation of indextree's size is correct

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


hackerwins (approved):
Thanks for your contribution.

PR #624, opened 2023-08-22
Support multi-level and partial element selection

### LINKED ISSUE #610: Enhancements for Element Selection and Multi-level Node Traversal in Tree.Edit
**Description**:
Following the implementation of #607, there are two crucial tasks that require attention:

1. Selection of Elements with Closing Tags:
Currently, the capability to select an element is limited to its opening tag. However, to address both editing and styling scenarios comprehensively, it's necessary to extend this functionality to encompass the selection of an element along with its closing tag. 

2. Multi-level Node Traversal for Element Selection:
Presently, within the Tree.Edit module, the scope of node editing is restricted to a single level. In scenarios involving multiple levels, it becomes necessary to devise a method that traverses nodes across different levels within a specified range.

**Why**:
Support various editing cases.

sejongk:
Support the below features by introducing tree traversal to Tree.Edit.
- Select multiple nodes in a multi-level range.
- Select a part of nodes (ex. selecting only the closing tag of the node).

It also removes every skip in test cases for Tree.
However, after introducing the later features like Tree.Move, those test cases should be changed to apply new features.

Fixes #610 

I used the index-based tree traversal, because the implementation is more concise and clear than the node-based traversal.
This method requires converting a crdt.TreePos into a index in a user's tree for tree traversal.
If there is any room for performance degradation because of this conversion, we can replace this with the node-based traversal.

```release-note

```

**Additional documentation**:
```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/pkg/index/tree.go
+++ b/pkg/index/tree.go
@@ -119,27 +119,27 @@ func postorderTraversal[V Value](node *Node[V], callback func(node *Node[V], dep
 	callback(node, depth)
 }
 
-// TagContain represents whether the opening or closing tag of a element is selected.
-type TagContain int
+// TagContained represents whether the opening or closing tag of a element is selected.
+type TagContained int
 
 const (
-	// ContainsAll represents that both opening and closing tag of a element are selected.
-	ContainsAll TagContain = 1 + iota
-	// ContainsOpening represents that only the opening tag is selected.
-	ContainsOpening
-	// ContainsClosing represents that only the closing tag is selected.
-	ContainsClosing
+	// AllContained represents that both opening and closing tag of a element are selected.
+	AllContained TagContained = 1 + iota
+	// OpeningContained represents that only the opening tag is selected.
+	OpeningContained
+	// ClosingContained represents that only the closing tag is selected.
+	ClosingContained
 )
 
 // ToString returns the string of TagContain.
-func (c TagContain) ToString() string {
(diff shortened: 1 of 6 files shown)


hackerwins:
@sejongk Could you also apply this task to JS SDK? It's crucial to thoroughly test the integration of the Tree feature with ProseMirror to ensure everything is functioning correctly.

sejongk:
> @sejongk Could you also apply this task to JS SDK? It's crucial to completely test the integration of the Tree feature with ProseMirror to ensure everything is functioning correctly.

Sure, I'll implement this for JS SDK.

pkg/index/tree.go
```diff
@@ -150,6 +176,10 @@ func nodesBetween[V Value](root *Node[V], from, to int, callback func(node V)) e
 			if child.IsText() {
 				toChild = to - pos
 			}
+			// pos can be affected by children traversal and callback call.
+			// so, update pos before these to keep it safe.
+			pos += child.PaddedLength()
```
hackerwins, line 181: It is better to revert this because we can't be sure this approach protects the problem.

hackerwins (approved):
Thanks for your contribution. Overall looks good.

I left minor comments.

hackerwins (approved):
Thanks for your contribution.

Issue #584, opened 2023-07-19
Move Client.Watch inside Client.Attach and hide it from external interface


hackerwins:
**Description**:

Move `Client.Watch` inside `Client.Attach` and hide it from the external interface.

Go SDK is just used in integration tests of servers without other SDK installations. So it was OK to expose `Client.Watch` to the external interface. But by adding more and more features to the SDK, it is quite difficult to keep simple tests.

Let's move Client.Watch inside Client.Attach and hide it from the external interface to maintain consistency with other SDKs and simplify testing.

**Why**:

Keep the product simple

karockai:
It looks interesting. May I work on this issue?

krapie:
@karockai sure! give it a try!

krapie:
Do we need to change the entire interface to perform `runWatchLoop()` and provide `subscribe()` interface just like other SDKs?

chacha912:
@krapie  Yes, that seems necessary.
Here's the discussion we had about what tasks need to be carried out regarding this issue, for your reference.
https://codepair.yorkie.dev/community/65e02d3166a65e76f6703d15/share?token=2ij9re

hackerwins:
Following #803, we need to apply the below features:
- [ ] Reconnection logic for watch connection failure.
- [ ] Apply SyncMode introduced in the JS SDK.

https://github.com/yorkie-team/yorkie-js-sdk/pull/772

hackerwins:
I'll close this for now and reopen it later when we have more Go SDK users.

PR #631, opened 2023-08-28
Introduce Broadcast API

### LINKED ISSUE #628: Introduce broadcast API for event sharing
Yorkie presently relies on the Publish-Subscribe model for sharing document and presence events (refer to: [pub-sub.md](https://github.com/yorkie-team/yorkie/blob/main/design/pub-sub.md)). 
However, this lacks the capability to extend its scope to encompass additional event types, notably notifications for end users  concerning new document updates or comments.

To address this limitation, the introduction of a "broadcast" feature is recommended. 
This feature would enable users to define and share a wider range of general events beyond the existing document and presence events.

It's also related to #442, which extracts `Room` from `Document` and moves `Presence` from `Client` to `Room`.

**Why is this needed**:
Provide a more comprehensive event-sharing mechanism that satisfies various use cases.

sejongk:
This pull request introduces the new "broadcast" feature, allowing for the sharing of general events.

- [x] Add Document.Broadcast API
  - [x] Update protocol: Add broadcast-related types
  - [x] SDK: Implement `Document.broadcast(event)`
  - [x] Server: Implement Broadcast API
- [x] SDK: Implement Document.Subscribe

Address #628

The following are the current implementations that need to be discussed.
I added test cases for the below specs.
- Clients who don't subscribe to an event type can broadcast for that event type.
- The payload in a broadcast event should be serializable to JSON.
- Clients can keep watching the document although an error occurs during the handling of a broadcast event.

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


sejongk:
@hackerwins 
There are some points to note regarding this implementation:

1. `DocumentBroadcast` is one of the `DocEvent` types, meaning that broadcast events are delivered as `DocEvent` objects through the watch stream.
2. Given the frequency of broadcasts and the complexity of the implementation, it appears to be a good practice to always send broadcasts to all clients, allowing clients to filter them out locally. If necessary, we can consider implementing server-side event filtering at a later stage.
3. The `PubSub` module can remain mostly unchanged since we won't be implementing server-side event filtering as mentioned above.

hackerwins (approved):
Thanks for your contribution.

It would be great to refine `SubscribeBroadcastEvent` along with integrating the Attach and Watch interfaces(#584).

PR #636, opened 2023-09-05
Remove panic from crdt primitive

### LINKED ISSUE #497: Don't Panic in the Server

fourjae:

### Action to remove panic method from crdt/primitive.go

I am leaving a question because I encountered a problem while uninstalling crdt/primitive.go Panic.

In the case of the func (p *Primitive) Marshal() method inside the crdt/primitive.go package, Marshal() of the type Element interface is implemented and used.

However, if you modify the interface to Marshal() (string, error), the related implementations must be modified.

At this time, Marshal() is called when calling other String() methods, and I am asking because I think it will be difficult to change this part of String().

https://github.com/yorkie-team/yorkie/blob/1b557be51471337d3c6949d5f9ed5634b3a677e2/pkg/document/crdt/rga_tree_list.go#L89

Please let me know if I'm missing something or if there's a better way. thank you

Fixes #497 ```crdt/primitive```

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


krapie:
@fourjae I think we need to change `Marshal() string` interface to `Marshal() (string, error)`, just like when we removed `panic()` in `crdt/text`.

In #522, we changed `DeepCopy() Element` signature to `DeepCopy() (Element, error)`, and updated all related codes.

Changing `Marshal()` will affect many other related codes including `String()`, so adding error handling will take some time.

fourjae:
@hackerwins

As we progressed, we encountered this problem. What do you think?

I'm trying to fix crdt/primitive marshal(), but I have the following problem.

1. Marshal() implements Marshal() of type Element interface. Therefore, the return format has been modified to (String, error) and is in progress.
2. Among the implemented Marshal()s, Marshal(), which receives the *Object structure as an argument, is used when calling from String() in crdt/rga_tree_list.go.
3. The String() is used in pkg/llrb.go and pkg/splay.go and implements the type Value interface. String() is implemented and used in approximately 12,000 places.
 
As a result, all String() methods must be changed to handle errors in marshal().

Do you have any good troubleshooting methods?

krapie:
> As we progressed, we encountered this problem. What do you think?
> 
> I'm trying to fix crdt/primitive marshal(), but I have the following problem.
> 
> 1. Marshal() implements Marshal() of type Element interface. Therefore, the return format has been modified to (String, error) and is in progress.
> 2. Among the implemented Marshal()s, Marshal(), which receives the *Object structure as an argument, is used when calling from String() in crdt/rga_tree_list.go.
> 3. The String() is used in pkg/llrb.go and pkg/splay.go and implements the type Value interface. String() is implemented and used in approximately 12,000 places.
> 
> As a result, all String() methods must be changed to handle errors in marshal().
> 
> Do you have any good troubleshooting methods?

@hackerwins Any ideas on this issue?

krapie (commented):
For discussion record: It will be nice to move `panic()` to the `json` layer so that we can catch errors at the `json` level and change types in the `crdt` layer to never emit errors on the `crdt` layer. This will prevent interface changes on `crdt` layer.

krapie:
After you resolve CI error and other issues, please open your PR.

fourjae:
There were things I didn't know about, and after analyzing them, I came to the following conclusion and changed it to PR.

1. (p *Primitive) As a result of checking the primitive type of the ```Marshal()``` function, there are 8 types (Null, Boolean, Integer, Long, Double, String, Bytes, Date) already declared as constants. Only the above 8 types could be executed.
Therefore, the panic itself could not occur.
https://github.com/yorkie-team/yorkie/blob/940941a0428ef8969df62c70568421aabd48b02b/pkg/document/crdt/primitive.go#L31

2. In the ```NewPrimitive``` function, the exception type (float) is also set to Double and values ​​other than those declared in the constant cannot be entered.

3. (p *Primitive) ```Marshal()``` function is Since we are not calling from ```json``` layer, we don't need to use any method to catch error at ```json``` level.

Therefore, the (p *Primitive) ```Marshal()``` function only erased the panic from the ```crdt``` layer.

krapie (changes requested):
I left some comments below.

fourjae:
Thanks for your comment!

Fixed code conventions that did not fit.

The hard-coded err message has been changed to return an err variable.

Please let me know if you need any additional changes.

pkg/document/json/array.go
```diff
@@ -194,7 +227,6 @@ func (p *Array) addInternal(
 ) crdt.Element {
 	return p.insertAfterInternal(p.Array.LastCreatedAt(), creator)
 }
-
```
krapie, line 197: It will be nice to have space between functions

pkg/document/json/object.go
```diff
@@ -120,34 +124,47 @@ func (p *Object) SetNewTree(k string, initialRoot ...*TreeNode) *Tree {
 // SetNull sets the null for the given key.
 func (p *Object) SetNull(k string) *Object {
 	p.setInternal(k, func(ticket *time.Ticket) crdt.Element {
-		return crdt.NewPrimitive(nil, ticket)
```
krapie, line 123: It will be nice to keep space convention of spacing before `return` in this file. I will leave only one comment for this, so check all the `return` codes in this file.

pkg/document/crdt/primitive_test.go
```diff
@@ -34,7 +34,7 @@ func TestPrimitive(t *testing.T) {
 		valueType crdt.ValueType
 		marshal   string
 	}{
-		{nil, crdt.Null, "null"},
+		{nil, crdt.Null, ""},
```
krapie, line 37: > Is there any reason for this change? I will be better to keep the previous code.

It will be better to revert this code too.

krapie (changes requested):
Thank you for reflecting my review.
I left just few more comments 😄

krapie (commented):
Looks good to me. 

Could you review this PR @hackerwins?

hackerwins (approved):

PR #639, opened 2023-09-06
Support concurrent formatting of Text

### LINKED ISSUE #638: Concurrent formatting do not work
**Description**:
In a situation where two clients are working simultaneously, if one applies a `Text.Style` operation for formatting while the other performs a `Text.edit` operation to insert, the results of these two operations do not converge into one. 

For example:
> before: `The fox jumped`
> client A: **`The fox jumped`**
> client B: `The brown fox jumped`
> expected: **`The`** `brown` **`fox jumped`** or **`The brown fox jumped`**

If we start with the sentence 'The fox jumped', client A makes the entire sentence bold, and client B inserts the word 'brown' between 'The' and 'fox,' the synchronized results are not the same, as shown below:
image

The result should converge into a single status, whether it's the entire sentence 'The brown fox jumped' with bold formatting or some other form that includes bold formatting except for 'brown.'

**Why**:
This is because we are unable to check for concurrent cases when applying the `Text.Style` operation. By introducing a `map` called `latestCreatedAtMapByActor`, which is used in `Text.edit` or `Tree.edit`, we can determine the causality between the operations of the two clients.

MoonGyu1:

Currently, we are unable to check for concurrent cases when applying the `Text.Style` operation. This pull request introduces a `map` called `latestCreatedAtMapByActor` to track the causality between the operations of the two clients and ensures that the results converge into one.
* Add `latestCreatedAtMapByActor` to `Text.Style` operation
* Add related testcases

Address #638

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


api/yorkie/v1/resources.proto
```diff
@@ -109,8 +109,9 @@ message Operation {
     TimeTicket parent_created_at = 1;
     TextNodePos from = 2;
     TextNodePos to = 3;
-    map<string, string> attributes = 4;
-    TimeTicket executed_at = 5;
+    map<string, TimeTicket> created_at_map_by_actor = 4;
+    map<string, string> attributes = 5;
+    TimeTicket executed_at = 6;
```
chacha912, line 114: It would be good to add `created_at_map_by_actor` at the end  to preserve protocol's compatibility.

```suggestion
    map<string, string> attributes = 4;
    TimeTicket executed_at = 5;
    map<string, TimeTicket> created_at_map_by_actor = 6; // add
```

pkg/document/crdt/text.go
```diff
@@ -250,28 +250,58 @@ func (t *Text) Edit(
...
 	// 02. style nodes between from and to
 	nodes := t.rgaTreeSplit.findBetween(fromRight, toRight)
+	createdAtMapByActor := make(map[string]*time.Ticket)
+	var toBeStyled []*RGATreeSplitNode[*TextValue]
+
 	for _, node := range nodes {
+		actorIDHex := node.id.createdAt.ActorIDHex()
+
+		var latestCreatedAt *time.Ticket
+		if latestCreatedAtMapByActor == nil {
```
chacha912, line 276: Style operations generated in the previous protocol version were setting `createdAtMap` as an empty map, which led to an issue where the `latestCreateAt` was set to `InitialTicket`, causing the style not to be applied.  To resolve this issue, we can set the `latestCreateAt` to `MaxTicket` when the map is empty. 

```suggestion
		if len(latestCreatedAtMapByActor) == 0 {
```

chacha912 (commented):
Thanks for contributing! :) 
The issue of the document not converging has been resolved!

Due to the protocol changes, it's needed to verify that the previous version of style operations also functions without issues. I've left comments related to this, so please check them.

chacha912 (approved):
Thanks. It looks good :)

hackerwins (approved):
Thanks for your contribution.

Issue #472, opened 2023-02-20
Introducing Lookup Based Cluster Mode to Yorkie Servers


krapie:
**What would you like to be added**:

Add lookup based cluster mode to Yorkie servers.

**Why is this needed**:

To overcome current Yorkie cluster mode's limitations and make Yorkie server capable of handling large amount of workloads in production environment.

For more information, follow below:

# Introducing Lookup Based Cluster Mode to Yorkie Servers

> Latest update: 2023.03.02

## Table of Contents

- [Introduction to Lookup Based Cluster Mode](#introduction-to-lookup-based-cluster-mode)
- [LookUp Based Cluster Mode Design](#lookup-based-cluster-mode-design)
  * [1. Workload Unit & Server Mapping Strategy & Hash Function](#1-workload-unit---server-mapping-strategy---hash-function)
  * [2. API Compatiability & LookUp Strategy](#2-api-compatiability---lookup-strategy)
  * [3. Server Addition/Removal Strategy](#3-server-addition-removal-strategy)
  * [4. System Design & Architecture](#4-system-design---architecture)
- [LookUp Based Cluster Mode Implementation](#lookup-based-cluster-mode-implementation)
  * [1. Tasks Needed for Lookup Cluster Mode Implementation](#1-tasks-needed-for-lookup-cluster-mode-implementation)
  * [2. Lookup Cluster Mode Implementation (docker-compose)](#2-lookup-cluster-mode-implementation-docker-compose)
  * [3. Lookup Cluster Mode Implementation (K8s & Istio)](#2-lookup-cluster-mode-implementation-k8s-istio)
  * [4. Yorkie Cluster PoC/Prototype Repository](#3-yorkie-cluster-poc-prototype-repository)
  * [5. Actual Infrastuctural Implementation](#5-actual-infrastuctural-implementation)

## Introduction to Lookup Based Cluster Mode

In production environment, server clusters are necessary to handle large amount of workloads with ensuring high availability, reliability, and scalability.

Current Yorkie server supports [server cluster mode](https://github.com/yorkie-team/yorkie/blob/main/design/cluster-mode.md), which is based on broadcasting, pub/sub, and distributed lock between servers. This architecture can handle certain amount of workloads, but there are limitations.

- **Broadcast and pub/sub overheads**: Because all servers have to communicate to each other with broadcast and pub/sub to share **workload states**, complexity, latency, and performance overheads will grow as more servers are created to handle workloads.
[...]

PR #642, opened 2023-09-13
Introduce sharding rules to MongoDB collections

### LINKED ISSUE #530: Introduce sharding rules to MongoDB collections
**Description**:

This description is written by @sejongk

To address the load concentrated on the database, we need to introduce a MongoDB Sharded Cluster and apply appropriate sharding rules to the collections for a solution.

- Indexes: https://github.com/yorkie-team/yorkie/blob/main/server/backend/database/mongo/indexes.go
- Note: https://codepair.yorkie.dev/khfbyv

**Why**:

The Yorkie cluster stores almost all data, including documents and changes representing edits, in MongoDB. It has been observed that in real-world service scenarios, the majority of the load is concentrated on the database rather than the application servers, as was the case with Yorkie's predecessor, Kaleido.

sejongk:
This PR introduces sharding rules to MongoDB collections to distribute loads on the database cluster.
It takes reference from #472.

Addresses #673  

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


build/docker/sharding/min/scripts/init-config1.js
```diff
@@ -0,0 +1,10 @@
+rs.initiate(
+    {
+      _id: "config-rs",
+      configsvr: true,
+      members: [
+        { _id : 0, host : "config1:27017" },
+      ]
+    }
+  )
+
```
hackerwins, line 10:
```suggestion

```

build/docker/sharding/min/scripts/init-config1.js
```diff
@@ -0,0 +1,9 @@
+rs.initiate(
+    {
+      _id: "config-rs",
+      configsvr: true,
+      members: [
+        { _id : 0, host : "config1:27017" },
+      ]
+    }
+  )
```
hackerwins, line 9:
```suggestion
  )

```

krapie (commented):
I left some small comments below.

> The first thing is how to guarantee the uniqueness of fields. In the Yorkie, there is a requirement to force unique constraints on several field combinations (e.g. owner and name in the project collection).

I personally think locking with distributed keys is the lowest priority option that we have considering our current sharded cluster mode, since we are avoiding communications between server instances in sharded cluster mode.

Proxy collection seems more attractive to me for now, since it is MongoDB's official suggestion and we already have some DB actions that require atomic execution.

But benchmarking will clearly show what is most suitable for our situation.

> The second thing is about the shard key. I choose the following keys for good performance under the current query patterns. Note that the `server_seq` ranged-sharding is used for changes, because range query is frequently used for the changes.

I think we have to constantly benchmark and tune collection shard keys based on your shard key selection.

PR #650, opened 2023-10-18
Revise Prometheus PVC Spec Syntax Error

krapie:

Revise Prometheus PVC spec syntax error.

I found out that the `yorkie-monitoring` helm chart was not storing prometheus metrics, which leads to data loss of metrics every time a node goes down (e.g., EC2 spot instance goes down).

This was because the helm chart syntax of `volumeClaimTemplate` was wrong (it was not linked to `storageSpec`). Therefore, I corrected the syntax to make Prometheus successfully claim PVC.

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/build/charts/yorkie-monitoring/Chart.yaml
+++ b/build/charts/yorkie-monitoring/Chart.yaml
@@ -13,8 +13,8 @@ maintainers:
 
 sources:
   - https://github.com/yorkie-team/yorkie
-version: 0.4.1
-appVersion: "0.4.1"
+version: 0.4.7
+appVersion: "0.4.7"
 kubeVersion: ">=1.24.0-0"
 
 keywords:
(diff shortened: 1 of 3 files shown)


build/charts/yorkie-monitoring/Chart.yaml
```diff
@@ -13,8 +13,8 @@ maintainers:
 
 sources:
   - https://github.com/yorkie-team/yorkie
-version: 0.4.0
-appVersion: "0.4.0"
+version: 0.4.1
+appVersion: "0.4.1"
```
hackerwins, line 17: How about matching this with Yorkie version like SDK version?

hackerwins (commented):
Thanks for your contribution. I left a comment.

build/charts/yorkie-monitoring/Chart.yaml
krapie: Seems like a good idea. I will reflect it.

krapie:
@hackerwins Can you wait before merging this PR?

PR #678, opened 2023-11-20
Modify converter to allow setting values for object and array

chacha912:
**What this PR does / why we need it**:

This PR is associated with https://github.com/yorkie-team/yorkie-js-sdk/pull/691. Since modifications have been made to allow setting values in set and add operations, it is necessary to update the converter accordingly. This involves enabling the setting of values for objects and arrays in the converter.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Added relevant tests or not required
- [x] Didn't break anything

PR #687, opened 2023-11-23
Modify converter to allow setting values for object and array

chacha912:

> This PR is associated with https://github.com/yorkie-team/yorkie-js-sdk/pull/691. Since modifications have been made to allow setting values in set and add operations, it is necessary to update the converter accordingly. This involves enabling the setting of values for objects and arrays in the converter.

[PR#678 triggers the regression error](https://github.com/yorkie-team/yorkie/pull/678#issuecomment-1823708123), so I made corrections and re-submitted. I've added handling for cases where values are not set when converting to `PbJSONElementSimple`. In the `bytesToObject` function, It was previously creating a new object with initialTicket when the bytes were nil. However, this logic became inappropriate after separating it from `bytesToSnapshot`, so I removed it.

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/api/converter/converter_test.go
+++ b/api/converter/converter_test.go
@@ -36,13 +36,8 @@ import (
 
 func TestConverter(t *testing.T) {
 	t.Run("snapshot simple test", func(t *testing.T) {
-		obj, err := converter.BytesToObject(nil)
-		assert.NoError(t, err)
-		assert.Equal(t, "{}", obj.Marshal())
-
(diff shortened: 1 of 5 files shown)


hackerwins (approved):
Thanks for your contribution.

hackerwins:
CC) @humdrum, @7hong13

PR #703, opened 2023-11-28
Migrate RPC to ConnectRPC

krapie:
**What this PR does / why we need it**:

Migrate RPC to [ConnecRPC](https://connectrpc.com/).

ConnectRPC supports following features:
- Supports interoperation and multiplexing of `grpc`, `grpc-web`, and `Connect` protocols.
- Supports language-familiar primitives like `net/http` and `fetch`.
- Supports simple and no-boilerplate codes.

With this support, you can resolve several issues:
- Removes dependency with Envoy proxy to communicate with JS SDK web client by [grpc-web](https://github.com/grpc/grpc-web).
- Removes dependency with deprecated [gogo/protobuf](https://github.com/gogo/protobuf).
- Supports JSON based REST API which will satisfy [these needs](https://github.com/yorkie-team/yorkie/issues/677).

**Which issue(s) this PR fixes**:

Related to #668

**Special notes for your reviewer**:

> [!NOTE]
> After merging this PR, I will remove envoy proxy on docker configuration on other repositories too.

> [!NOTE]
> JS SDK compatibility: Since `connect-go` does not support `grpc-web-text` which current JS SDK gRPC-web client uses, we need to migrate JS SDK's RPC to ConnectRPC as well. We can temporary change JS SDK's docker-compose settings (keep envoy proxy and change server rpc port to `11101`) to support compatibility with ConnectRPC server and JS SDK gRPC-web client before finishing migration of ConnectRPC on JS SDK.

> [!NOTE]
> Server configuration: We need to find corresponding `net/http` server configurations that matches with `gRPC` server options such as `MaxRequestBytes`, `MaxConnectAge` and `MaxConnectionAgeGrace`. Because of this issue, `watch document test` is temporary commented.

> [!NOTE]
> gRPC Metrics: Since we cannot use `go-grpc-prometheus` that measured gRPC server metrics, I have implemented RPC metrics by referencing `go-grpc-prometheus`. Because the syntax of newly implemented metrics are different from `go-grpc-prometheus`, we also need to update Grafana query (ex: `grpc_server_handled_total` to `yorkie_rpc_server_handled_total`, `grpc_type` to `rpc_type`).

> [!NOTE]
> Since we no longer have Envoy proxy to handle CORS, I have implemented CORS on RPC server. But for cluster mode, we have ingressgateway to handle CORS. So we need to implement disable option for CORS on server.

**Does this PR introduce a user-facing change?**:

```release-note
[...]

PR #741, opened 2024-01-01
Fix Issue with Non-ConnectError

chacha912:

Previously, there was an issue in [Dashboard](https://github.com/yorkie-team/dashboard/pull/135) where errors were not being properly distinguished as `ConnectError`. To address this problem, I replaced the usage of gRPC status error with `connect.NewError`.

- as-is:
image
- to-be:
image

Related #703 

I'm unsure about how to validate this change in the test code, as it is currently passing test using the existing approach.

https://github.com/yorkie-team/yorkie/blob/d932b59ae4f68203d953b91a31ace24c56c8ad41/test/integration/auth_webhook_test.go#L92-L141

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/server/rpc/interceptors/admin_auth.go
+++ b/server/rpc/interceptors/admin_auth.go
@@ -31,6 +31,9 @@ import (
 	"github.com/yorkie-team/yorkie/server/users"
 )
 
+// ErrUnauthenticated is returned when authentication is failed.
+var ErrUnauthenticated = errors.New("authorization is not provided")
+
 // AdminAuthInterceptor is an interceptor for authentication.
 type AdminAuthInterceptor struct {
 	backend      *backend.Backend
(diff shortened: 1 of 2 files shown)


krapie (approved):
Seems like I have forgot to update some of error handling codes on interceptors. Thanks for the updates 👍🏼

> I'm unsure about how to validate this change in the test code, as it is currently passing test using the existing approach.

`auth_webhook_test.go` does not validates admin interceptor. Since this code is for admin, you need to add new test on `admin_test.go`.

it will look something like this:

```go
t.Run("authorization test", func(t *testing.T) {
	// try to do admin operation without authorization.
	adminCli2, err := admin.Dial(defaultServer.RPCAddr(), admin.WithInsecure(true))
	assert.NoError(t, err)
	defer func() {
		adminCli2.Close()
	}()
	_, err = adminCli2.GetProject(ctx, "default")
	assert.Equal(t, connect.CodeUnauthenticated, connect.CodeOf(err))
})
```

hackerwins (approved):

Issue #626, opened 2023-08-23
Add a feature to Tree that allows the removal of attributes from elements


hackerwins:
**What would you like to be added**:

Add a feature to Tree that allows the removal of attributes from elements

Our objective in developing Tree is to create a data structure capable of representing tree-based models for WYSIWYG Editors.

```ts
root.tree = new Tree({
  type: 'doc',
  children: [{
    type: 'p',
    attributes: { bold: true },
    children: [{ type: 'text', value: 'hello' }],
  }],
});
assert.equal(root.tree.toXML(), `<doc><p bold="true">hello</p></doc>`);
```

Currently, users can modify attributes of elements using `Tree.style`:

```ts
root.t.style(0, 1, { italic: true });
assert.equal(root.tree.toXML(), `<doc><p bold="true" italic="true">hello</p></doc>`);
```

However, the current interface lacks a method for removing attributes from elements. As a workaround, users resort to setting attributes to false values like null or undefined. This approach can lead to confusion, especially in projects where multiple versions are maintained using Yorkie.

To address this issue, we could introduce a dedicated interface for removing attributes. This would enhance clarity and user experience across different implementations.

```ts
root.t.removeStyle(0, 1, 'italic');
```

**Why is this needed**:

hackerwins:
This issue has been resolved.

PR #748, opened 2024-01-05
Implement Tree.RemoveStyle

justiceHui:

* Implemented `Tree.RemoveStyle`.
* Added test codes to test `RHT` behavior in a concurrency scenario:
  * Edit style by 2 users (4 * 6 * 6 = 144 cases):
    * Ranges: A = B, A contains B, A and B are intersecting, not intersecting.
    * Operations: `RemoveStyle({'bold'})`, `Style({'bold': 'aa'})`, `Style({'bold': 'bb'})`, `RemoveStyle({'italic'})`, `Style({'italic': 'aa'})`, `Style({'italic': 'bb'})`.
  * Edit tree and Edit style (6 * 5 * 2 = 60 cases):
    * Ranges: A = B, A contains B, B contains A, A and B are intersecting, B is next to A, A is next to B.
    * Operations for A: Insert front of range, Insert back of range, Insert middle of range, Delete, Change.
    * Operations for B: `RemoveStyle({'bold'})`, `Style({'bold': 'aa'})`.

Addresses #626

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


test/integration/tree_test.go
```diff
@@ -734,6 +734,41 @@ func TestTree(t *testing.T) {
...
+				},
+			})
+			return nil
+		}))
+		assert.NoError(t, c1.Sync(ctx))
+		assert.Equal(t, `<root><p>ab</p><p italic="true">cd</p></root>`, d1.Root().GetTree("t").ToXML())
+
+		assert.NoError(t, d1.Update(func(root *json.Object, p *presence.Presence) error {
+			// NOTE(sejongk): 0, 4 -> 0,1 / 3,4
+			root.GetTree("t").RemoveStyle(0, 4, []string{"italic"})
```
hackerwins, line 757: If this deletes the italic in the second paragraph, it should be modified as shown below.

```suggestion
			root.GetTree("t").RemoveStyle(4, 8, []string{"italic"})
```

For more details about indexes in Tree:
https://github.com/yorkie-team/yorkie/blob/main/design/tree.md#proposal-details

pkg/document/crdt/tree.go
```diff
@@ -876,6 +876,34 @@ func (t *Tree) Style(from, to *TreePos, attributes map[string]string, editedAt *
...
+	if err != nil {
+		return err
+	}
+
+	err = t.traverseInPosRange(fromParent, fromLeft, toParent, toLeft,
+		func(token index.TreeToken[*TreeNode], _ bool) {
+			node := token.Node
+			if !node.IsRemoved() && !node.IsText() && len(attributesToRemove) > 0 && node.Attrs != nil {
+				for _, value := range attributesToRemove {
+					node.Attrs.Remove(value, editedAt)
```
hackerwins, line 896: `RHT.Remove` has never been used and seems buggy.

The problem I found is that when the code below is called, a panic occurs because `Node.removedAt` is nil.

https://github.com/yorkie-team/yorkie/blob/3b712df6c3d03029a1cf565070f0cfc2bf960abe/pkg/document/crdt/rht.go#L116-L124

I think we need to add tests for concurrent value setting and deletion in RHT.
https://github.com/yorkie-team/yorkie/blob/main/pkg/document/crdt/rht_test.go

pkg/document/json/tree.go
```diff
@@ -223,6 +223,37 @@ func (t *Tree) Style(fromIdx, toIdx int, attributes map[string]string) bool {
 	return true
 }
 
+// RemoveStyle sets the attributes to the elements of the given range.
+func (t *Tree) RemoveStyle(fromIdx, toIdx int, attributesToRemove []string) bool {
```
hackerwins, line 227: How about returning early if attributesToRemove is empty? I think this can prevent an empty operation from being created.

hackerwins (commented):
Thanks for your contribution. I left a few comments.

test/integration/tree_test.go
justiceHui: I fixed it.

pkg/document/crdt/tree.go
justiceHui: I fixed it. I found that after remove element, return value of `RHT.Len` is incorrect, and I have also fixed this issue.

pkg/document/json/tree.go
justiceHui: I fixed `Tree.Style` and `Tree.RemoveStyle` not to make empty operation.

pkg/document/crdt/tree.go
```diff
@@ -876,6 +876,34 @@ func (t *Tree) Style(from, to *TreePos, attributes map[string]string, editedAt *
...
+	}
+	toParent, toLeft, err := t.FindTreeNodesWithSplitText(to, editedAt)
+	if err != nil {
+		return err
+	}
+
+	err = t.traverseInPosRange(fromParent, fromLeft, toParent, toLeft,
+		func(token index.TreeToken[*TreeNode], _ bool) {
+			node := token.Node
+			if !node.IsRemoved() && !node.IsText() && len(attributesToRemove) > 0 && node.Attrs != nil {
```
hackerwins, line 894: How about removing `len(attributesToRemove) > 0 && node.Attrs != nil` conditions in this line? I think empty `attributesToRemove` should not be passed here and `node.Attrs` should be created before calling this method.

```suggestion
			if !node.IsRemoved() && !node.IsText() {
```

hackerwins (commented):
Thank you for your reflection. 👍

Next, we need to be sure that concurrent adding attributes and removing them is correct by adding test cases.

pkg/document/crdt/tree.go
justiceHui: The condition `len(attributesToRemove) > 0` can be removed, but `node.Attrs != nil` is necessary because a user might call `Tree.RemoveStyle` without having called `Tree.Style` previously.

justiceHui: I modified the code to create a new `RHT` if `node.Attrs` is `nil` to support concurrency.

hackerwins (approved):

PR #752, opened 2024-01-08
Fix RHT.Remove and Add test code

justiceHui:

* Fixed nil pointer dereference error in `RHT.Remove`.
* Fixed `RHT.Len` to exclude the removed elements.
* Fixed `RHT.Nodes` to exclude the removed elements.
* Fixed an error in `RHT.Set` when modifying an already deleted key.

Related to #748

Fixes #

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/pkg/document/crdt/rht.go
+++ b/pkg/document/crdt/rht.go
@@ -29,7 +29,6 @@ type RHTNode struct {
 	key       string
 	val       string
 	updatedAt *time.Ticket
-	removedAt *time.Ticket
 	isRemoved bool
 }
 
@@ -38,20 +37,19 @@ func newRHTNode(key, val string, updatedAt *time.Ticket) *RHTNode {
 		key:       key,
 		val:       val,
 		updatedAt: updatedAt,
+		isRemoved: false,
 	}
 }
 
 // Remove removes this node. It only marks the deleted time (tombstone).
 func (n *RHTNode) Remove(removedAt *time.Ticket) {
-	if n.removedAt == nil || removedAt.After(n.removedAt) {
-		n.removedAt = removedAt
-		n.isRemoved = true
-	}
+	n.updatedAt = removedAt
+	n.isRemoved = true
 }
 
 // IsRemoved returns the node has been removed.
 func (n *RHTNode) IsRemoved() bool {
-	return n.removedAt != nil && n.isRemoved
+	return n.isRemoved
 }
 
 // Key returns the key of this node.
@@ -67,11 +65,6 @@ func (n *RHTNode) Value() string {
 // UpdatedAt returns the last update time.
 func (n *RHTNode) UpdatedAt() *time.Ticket {
 	return n.updatedAt
-}
-
-// RemovedAt returns the deletion time of this node.
-func (n *RHTNode) RemovedAt() *time.Ticket {
-	return n.removedAt
 }
 
 // RHT is a hashtable with logical clock(Replicated hashtable).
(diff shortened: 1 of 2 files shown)


pkg/document/crdt/rht.go
```diff
@@ -48,7 +48,7 @@ func (n *RHTNode) Remove(removedAt *time.Ticket) {
 }
 
 func (n *RHTNode) isRemoved() bool {
-	return n.removedAt != nil
+	return n.removedAt != nil && n.updatedAt != nil && n.removedAt.Compare(n.updatedAt) == 0
```
hackerwins, line 51: Why is this comparison(`&& n.updatedAt != nil && n.removedAt.Compare(n.updatedAt) == 0`) necessary?

hackerwins (approved):
Thanks for your contribution.
Overall looks good. I left a simple question.

pkg/document/crdt/rht.go
justiceHui: Since the removed elements always guarantee the invariant `updatedAt = removedAt`, it is essential to check this condition in the `isRemoved`. Although I considered modifying only `removedAt` in the `Remove`, but this approach would be inconvenient, because we have to compare both `updatedAt` and `removedAt` with `executedAt` in the `RHT.Set`.

hackerwins: Then, is it possible to express whether to remove a node with just a flag(`isRemoved`)?

justiceHui: I think it's a good idea. I'll add the `isRemoved` flag and reimplement it.

pkg/document/crdt/rht.go
```diff
@@ -99,24 +98,38 @@ func (rht *RHT) Get(key string) string {
...
+		return node != nil && !node.IsRemoved()
 	}
 
 	return false
 }
 
 // Set sets the value of the given key.
 func (rht *RHT) Set(k, v string, executedAt *time.Ticket) {
-	if node, ok := rht.nodeMapByKey[k]; !ok || executedAt.After(node.updatedAt) {
+	if node, ok := rht.nodeMapByKey[k]; !ok || node.updatedAt == nil || executedAt.After(node.updatedAt) {
```
hackerwins, line 109:
```suggestion
	if node, ok := rht.nodeMapByKey[k]; !ok || executedAt.After(node.updatedAt) {
```

I think it would be better to remove `nil` check. `node.updatedAt` should always be set and if it is nil unintentionally, it's good to let the system crash rather than prevent it, because we should fix it.

justiceHui: I added code to check if `updatedAt` is `nil` because the test code calls the `Set` function with a `nil` value as `excutedAt` when testing `Marshal` and `ToXML`. Should I modify the test code to avoid passing `nil` values?

pkg/document/crdt/rht.go
```diff
@@ -108,15 +96,28 @@ func (rht *RHT) Has(key string) bool {
...
 // Remove removes the Element of the given key.
 func (rht *RHT) Remove(k string, executedAt *time.Ticket) string {
-	if node, ok := rht.nodeMapByKey[k]; ok && executedAt.After(node.removedAt) {
-		node.Remove(executedAt)
+	if node, ok := rht.nodeMapByKey[k]; ok && executedAt.After(node.updatedAt) {
+		alreadyRemoved := node.isRemoved
+		if !alreadyRemoved {
+			rht.numberOfRemovedElement++
+		}
+		// node is removed if and only if updatedAt = removedAt
```
hackerwins, line 114:
```suggestion
```

hackerwins (approved):
LGTM. 👍

Issue #677, opened 2023-11-15
Add API for Retrieving All Documents


devleejb:
# Motivation

During the operation of our service, the need for data analysis, such as inspecting accumulated data in Yorkie Documents, has arisen.

# Description

Proposing the addition of a REST API to retrieve the contents of Yorkie Documents, enabling developers using Yorkie to utilize the data.

krapie:
Currently, there is an `GetDocumentSummary()` admin gRPC API for getting a snapshot of a document.
Are you suggesting an HTTP REST API in JSON format to get document snapshots?

devleejb:
Exactly. I think I need an HTTP REST API. The service data is needed more for data analysis rather than in real-time, and gRPC has significant technical overhead.

hackerwins:
@krapie I remember connect-rpc also exposing the APIs in HTTP as well as gRPC. Am I right?
https://connectrpc.com/

krapie:
@hackerwins Yes, connect-rpc also exposes plain HTTP along with gRPC. I think we can also resolve this issue when we switching current server rpc to connect-rpc.

hackerwins:
Great. We will soon be using `Secret Key` for the Admin API too.
<img width="1126" alt="Screenshot 2023-11-17 at 2 25 53 PM" src="https://github.com/yorkie-team/yorkie/assets/2059311/36bfc93f-2c74-4049-9de8-21159c72c31a">

hackerwins:
After migrating to Connect RPC, we can use REST API.

I think this issue could be solved by using the Private Key created in the Dashboard during the authentication step.

devleejb:
Then, is it possible to provide API for other features? (Document Editing, ...)

hackerwins:
> Then, is it possible to provide API for other features? (Document Editing, ...)

It would be better to deal with this as a separate issue.

PR #755, opened 2024-01-09
Add API for Retrieving All Documents

raararaara:
This PR introduces additional functionality required for the creation and retrieval of documents through REST API. Specifically, it adds the following enhancements:

* Authentication via Project Secret Key: Previously, authentication for API access was limited to user JWT tokens only. With this update, we have extended the authentication mechanism to also allow authentication using a Project Secret Key. This change provides more flexibility in how users and systems can authenticate and access our API.

Related #677 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/server/backend/database/database.go
+++ b/server/backend/database/database.go
@@ -124,7 +124,7 @@ type Database interface {
 	// FindUserInfo returns a user by the given username.
 	FindUserInfo(ctx context.Context, username string) (*UserInfo, error)
 
-	// FindUserInfoById finds a user by the given id.
+	// FindUserInfoByID finds a user by the given id.
 	FindUserInfoByID(ctx context.Context, id string) (*UserInfo, error)
 
 	// ListUserInfos returns all users.
--- a/server/backend/database/database.go
+++ b/server/backend/database/database.go
@@ -125,7 +125,7 @@ type Database interface {
 	FindUserInfo(ctx context.Context, username string) (*UserInfo, error)
 
 	// FindUserInfoById finds a user by the given id.
-	FindUserInfoById(ctx context.Context, id string) (*UserInfo, error)
+	FindUserInfoByID(ctx context.Context, id string) (*UserInfo, error)
 
 	// ListUserInfos returns all users.
 	ListUserInfos(ctx context.Context) ([]*UserInfo, error)
--- a/server/backend/database/database.go
+++ b/server/backend/database/database.go
@@ -68,6 +68,12 @@ type Database interface {
 		publicKey string,
 	) (*ProjectInfo, error)
 
+	// FindProjectInfoBySecretKey returns a project by secret key.
+	FindProjectInfoBySecretKey(
+		ctx context.Context,
+		secretKey string,
+	) (*ProjectInfo, error)
+
 	// FindProjectInfoByName returns a project by the given name.
 	FindProjectInfoByName(
 		ctx context.Context,
@@ -117,6 +123,9 @@ type Database interface {
 
 	// FindUserInfo returns a user by the given username.
 	FindUserInfo(ctx context.Context, username string) (*UserInfo, error)
+
+	// FindUserInfoById finds a user by the given id.
+	FindUserInfoById(ctx context.Context, id string) (*UserInfo, error)
 
 	// ListUserInfos returns all users.
 	ListUserInfos(ctx context.Context) ([]*UserInfo, error)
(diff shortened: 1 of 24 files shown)


server/backend/database/database.go
```diff
@@ -118,6 +124,9 @@ type Database interface {
 	// FindUserInfo returns a user by the given username.
 	FindUserInfo(ctx context.Context, username string) (*UserInfo, error)
 
+	// FindUserInfoById finds a user by the given id.
```
hackerwins, line 127:
```suggestion
	// FindUserInfoByID finds a user by the given id.
```

server/projects/projects.go
```diff
@@ -107,3 +107,21 @@ func GetProjectFromAPIKey(ctx context.Context, be *backend.Backend, apiKey strin
...
+
+// GetProjectFromSecretKey returns a project from a secret key.
+func GetProjectFromSecretKey(ctx context.Context, be *backend.Backend, secretKey string) (*types.Project, error) {
+	if secretKey == "" {
+		info, err := be.DB.FindProjectInfoByID(ctx, database.DefaultProjectID)
+		if err != nil {
+			return nil, err
+		}
+		return info.ToProject(), nil
+	}
```
hackerwins, line 119: I think this logic returns the default project if we don't pass a secretKey. Why is this needed?

server/backend/database/database.go
```diff
@@ -118,6 +124,9 @@ type Database interface {
...
+	FindUserInfoByID(ctx context.Context, id string) (*UserInfo, error)
```
hackerwins, line 128: We use `types.ID` to identify the resource in DB. Is it possible to change this?

```suggestion
	FindUserInfoByID(ctx context.Context, id types.ID) (*UserInfo, error)
```

server/rpc/interceptors/admin_auth.go
```diff
@@ -162,14 +163,20 @@ func (i *AdminAuthInterceptor) authenticate(
...
+		user, err := users.GetUser(ctx, i.backend, claims.Username)
+		if err == nil {
+			return user, nil
+		}
 	}
 
-	user, err := users.GetUser(ctx, i.backend, claims.Username)
-	if err != nil {
-		return nil, connect.NewError(connect.CodeUnauthenticated, ErrUnauthenticated)
+	project, err := projects.GetProjectFromSecretKey(ctx, i.backend, authorization)
```
hackerwins, line 173:
```suggestion
	// NOTE(hackerwins): If the token is secret key, return the owner of the project.
	project, err := projects.GetProjectFromSecretKey(ctx, i.backend, authorization)
```

server/rpc/interceptors/admin_auth.go
```diff
@@ -162,14 +163,20 @@ func (i *AdminAuthInterceptor) authenticate(
 	}
 
 	claims, err := i.tokenManager.Verify(authorization)
```
hackerwins, line 165:
```suggestion
	// NOTE(hackerwins): If the token is access token, return the user of the token.
	claims, err := i.tokenManager.Verify(authorization)
```

hackerwins (changes requested):
Thanks for your contribution. Overall looks good.

I left a few comments.

server/projects/projects.go
raararaara: I was also curious about the necessity of this logic, and since `GetProjectFromAPIKey` was returning the default project in the same way, I thought there must be a reason, so I did the same.
I don't particularly think of any side effects that occur when the default project is not returned.

server/backend/database/database.go
raararaara: Initially, it was implemented based on `types.ID`, but when testing in memoryDB, finding user by clientID with type `types.ID` did not work correctly. 

This was solved by converting `types.ID` to `string` when querying memoryDB.

server/projects/projects.go
hackerwins: To get started with Yorkie, users can easily install it using a package manager like Homebrew and they can run Yokie in standalone mode. We allows users to test the functionality without the need to create a user or project explicitly to get the apiKey.

But I don't think this feature(return default project without apiKey) needs for secretKey.

hackerwins: https://github.com/yorkie-team/yorkie/pull/755/commits/b3a93275abf0d17c6ad65ff3f6f7bee7e5b93d7c

hackerwins (approved):
LGTM. 🚀

PR #766, opened 2024-01-19
Improve object creation with an initial value of specific types

highcloud100:
---

Current users can only set the array initial value to []interface{} without any specific type, this problem also occurs when set an object with the json literal. 

--- 

Fixes https://github.com/yorkie-team/yorkie/pull/756 Limitations

--- 
[Specification of generating crdt.Object with literal](https://hackmd.io/XcQYDee0QP6jCTpLFC8HCA)

---

### 1. SetNewArray with the specific type

Supported Types
```shell
interface{ }, int, float, string, bool, []byte, time.Time, json.Counter, json.Text, json.Tree, map[string]interface{ }, user defined struct, array, slice
```

> SetNewArray

- As-is 
```go
root.SetNewArray("k2", []interface{0, 1, 2})
```

- To-be
```go
root.SetNewArray("k2", []int{0, 1, 2})
```

> SetNewObject with json literal

- As-is
```go
root.SetNewObject("obj", map[string]interface{}{
     "array": []interface{}{1, 2, 3, []interface{}{7, 8}},
})
```

- To-be
```go
root.SetNewObject("obj", map[string]interface{}{
     "array": []interface{}{1, 2, 3, []int{7, 8}},
})
```

### 2. SetNewObject with user defined struct
```go
text := "foo"
root.SetNewObject("obj", struct{ M string }{M: text})
root.SetNewObject("obj", &struct{ M *string }{M: &text}) // It also supports pointers

// `{"obj":{"M":"foo"}}`
```

### 3. Support struct tag
- when generating NewObject or NewArray with user defined struct, user can use struct tag
- Currently only support "-", "omitEmpty", "user defined name"

```
`yorkie:(user defined name || - ),options`
```

```go
T2 struct {
	Skip  string `yorkie:"-"`
	M     string `yorkie:"m"`
	Empty string `yorkie:"e,omitEmpty"`
}
```

**Checklist**:
- [ ] Didn't break anything

### DIFF
--- a/pkg/document/crdt/counter.go
+++ b/pkg/document/crdt/counter.go
@@ -60,9 +60,6 @@ type Counter struct {
 
 // NewCounter creates a new instance of Counter.
 func NewCounter(valueType CounterType, value interface{}, createdAt *time.Ticket) (*Counter, error) {
-	if value == nil {
-		value = 0
-	}
 	switch valueType {
 	case IntegerCnt:
 		intValue, err := castToInt(value)
--- a/pkg/document/crdt/counter.go
+++ b/pkg/document/crdt/counter.go
@@ -60,6 +60,9 @@ type Counter struct {
 
 // NewCounter creates a new instance of Counter.
 func NewCounter(valueType CounterType, value interface{}, createdAt *time.Ticket) (*Counter, error) {
+	if value == nil {
+		value = 0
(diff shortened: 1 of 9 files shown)


hackerwins (approved):
LGTM. 👍

Issue #673, opened 2023-11-09
Introduce sharding rules to MongoDB collections


sejongk:
> Last updated at November 09, 2023
> Last edited by @sejongk 
> If you find something deprecated, please feel free to add to or correct this issue.

**What would you like to be added**:
Support MongoDB Sharding feature in Yorkie. The previous issue about this is already at #530, which contains specification and initial discussion about it.

This issue is more about sharing the sharding related discussions and implements up to date, with the goal of getting more contributors involved in the work.

**Why is this needed**:
The Yorkie cluster stores most of data in MongoDB. Therefore, the most of loads are concentrated on DB clusters rather than application servers. For the case of real-world service scenarios, it is necessary to support sharding for distributing data and query loads across horizontally-scalable db clusters.

# 1. Introduction of Sharding
## 1.1. What is the Sharding?
When large data should be stored or high query throughput is required, data can be divided into several partitions and distributed to several nodes. This method is called as ”Sharding” or “Partitioning”.  Sharding enables horizontal scaling since each shard only has to manage a subset of the data for a sharded collection. As a result, the data and the query loads can be distributed evenly to nodes.

## 1.2. Sharded Cluster Components
A MongoDB sharded cluster consists of the following components.
1. shard: Each shard contains a subset of the sharded data.
2. mongos: The mongos acts as a query router, providing an interface between client applications and the sharded cluster.
3. config servers: Config servers store metadata and configuration settings for the cluster.

For a production deployment, consider the following to ensure data redundancy and system availability.
* Config Server (3 member replica set): `config1`,`config2`,`config3`
* 3 Shards (each a 3 member replica set):
	* `shard1-1`,`shard1-2`, `shard1-3`
	* `shard2-1`,`shard2-2`, `shard2-3`
	* `shard3-1`,`shard3-2`, `shard3-3`
* 2 Mongos: `mongos1`, `mongos2`

## 1.3. Shard Key Selection
It is important to choose a good shard key to distribute data evenly across shards and achieve high efficiency and performance of operations. 

### 1.3.1. Shard Key Selection Criteria
It is notable that only one of the following criteria does not, on its own, guarantee even distribution of data across the sharded cluter. Therefore, it is necessary to consider overall criteria to decide appropriate shard keys.
1. High Shard Key Cardinality
[...]

PR #769, opened 2024-01-19
Add Helm chart for MongoDB sharded cluster

sejongk:
This PR adds a Helm chart for MongoDB, supporting both standalone mode and sharded cluster mode.
Shard rules, including shard keys, shard methods, and unique constraints, are defined and managed in the `values.yaml`.

This chart depends on the Bitnami's `mongodb-sharded` chart (https://github.com/bitnami/charts/tree/main/bitnami/mongodb-sharded).
There are a few known issues with it:
- A problem with livenessProbe and readinessProbe
  - Issues
    - https://github.com/bitnami/charts/issues/21957
  - Solution: use custom livenessProbe and readinessProbe instead
- Lack of ARM64 support in the Bitnami `mongodb-sharded` **container**
  - Issues 
    - https://github.com/bitnami/charts/issues/7305#issuecomment-1443187343
    - https://github.com/bitnami/containers/issues/40947
  - Solution: use the official mongo 6.0 image instead and set up the cluster via a Job. 

### How to install and uninstall
1. Create `mongodb` namespace
```
kubectl create namespace mongodb
```
2. Install helm chart with the set flag and the namespace flag (The reason why the namespace flag is needed to be set is due to the misguided value in the Bitnami chart. In [here](https://github.com/bitnami/charts/blob/947458bc7c22d489a506306ba48782d38b906450/bitnami/mongodb-sharded/templates/_helpers.tpl#L36), it uses `.Release.Namespace` without offering an option to override it. 
```
helm install mongodb . --namespace mongodb --set=sharded.enabled=true
```
3. Uninstall helm chart with the namespace flag
```
helm install mongodb . --namespace mongodb
```

Addresses #673 

We need to do following tasks in the future:

- [ ] Enable livenessProbe, readinessProbe
- [ ] Connect to Prometheus via `PodMonitor`s

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


build/charts/yorkie-mongodb/templates/sharded/configmap.yaml
```diff
@@ -0,0 +1,70 @@
...
+          {{- range .shardKeys }}
+            {{ printf "%s: %s," .name .method }}
+          {{- end }}
+        }, 
+        {{ .unique }}
+      );
+      {{- end }}
+    EOF
+    
+
```
hackerwins, line 69:
```suggestion
    
```

build/charts/yorkie-mongodb/values.yaml
```diff
@@ -0,0 +1,189 @@
+namespace: &namespace mongodb
+name: &name mongodb
+port: &port 27017
+clusterDomain: &clusterDomain cluster.local
+
+image:
```
hackerwins, line 6: Could you add descriptions for each section(image, persistence, sharded, and mongodb-sharded)?

hackerwins (changes requested):
Thanks for your contribution.
I left a few comments.

krapie (commented):
Thank you for your contribution 😄
First off, I have left some small conventional suggestions along with comprehensive questions.

#### Question 1

It seems like `yorkie-cluster` Helm chart no longer has mongoDB subshart. This means that we now need to install both `yorkie-cluster` and `yorkie-mongodb` Helm chart to fully use yorkie cluster. Is there any reason for this separation?

#### Question 2

It seems like `yorkie-mongodb` Helm chart performs sharded mongoDB setup by using Job object. Is there any reason for this approach?

I can see that mongodb-sharded Helm chart provides `common.initScriptsCM` value to initialize mongoDB instances (Nevertheless [I also think using Job is the best approach for this kind of operation](https://github.com/docker-library/mongo/issues/339#issuecomment-1276780427), I just want to know the reasons for this approach).

> Lack of ARM64 support in the Bitnami mongodb-sharded container
> Solution: use the official mongo 6.0 image instead and set up the cluster via a Job.

Also, is there any relations between ARM64 support and setting up cluster via Job?

#### Suggestion 3

How about adding [`NOTES.txt`](https://github.com/yorkie-team/yorkie/blob/main/build/charts/yorkie-cluster/templates/NOTES.txt) to keep consistency with other yorkie Helm charts?

#### Suggestion 4

We might need to add [ArgoCD application manifest](https://github.com/yorkie-team/yorkie/blob/main/build/charts/yorkie-argocd/templates/applications.yaml) for `yorkie-mongodb` in `yorkie-argocd` Helm chart.

krapie (approved):
LGTM 👍🏼

hackerwins (approved):
LGTM.

sejongk:
> #### Question 1
> It seems like `yorkie-cluster` Helm chart no longer has mongoDB subshart. This means that we now need to install both `yorkie-cluster` and `yorkie-mongodb` Helm chart to fully use yorkie cluster. Is there any reason for this separation?
> 
> #### Question 2
> It seems like `yorkie-mongodb` Helm chart performs sharded mongoDB setup by using Job object. Is there any reason for this approach?
> 
> I can see that mongodb-sharded Helm chart provides `common.initScriptsCM` value to initialize mongoDB instances (Nevertheless [I also think using Job is the best approach for this kind of operation](https://github.com/docker-library/mongo/issues/339#issuecomment-1276780427), I just want to know the reasons for this approach).
> 
> > Lack of ARM64 support in the Bitnami mongodb-sharded container
> > Solution: use the official mongo 6.0 image instead and set up the cluster via a Job.
> 
> Also, is there any relations between ARM64 support and setting up cluster via Job?

Thanks for your careful review.
Here are my answers for your questions.

#### Answer 1
After separating MongoDB from yorkie-cluster, there are now two options for users: self-hosted or managed databases (e.g., MongoDB Atlas, AWS DynamoDB). I believe we can focus on supporting self-hosted clusters by solely maintaining yorkie-mongodb.

#### Answer 2
I initially explored the approach using Bitnami options and tools. Bitnami charts are designed to be used with Bitnami containers together. However, issues arose, including the lack of ARM64 images in the Bitnami `mongodb-sharded` container. Instead of incurring the current and potential future engineering costs associated with external tools, I opted to implement and maintain the code ourselves. 
I believe the current requirements for this are relatively simple and manageable. If future requirements become more complex and there is a good alternative, that would be the time to consider such an option.

PR #786, opened 2024-02-06
Introduce Hybrid vector clock

highcloud100:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Address #723

**Special notes for your reviewer**:

https://hackmd.io/kfaV_ZAgSX-REVyKRnoU1Q?view
https://hackmd.io/ogc6dOGkRW-At6O8SmiDrg?both#3-vector-clock%EC%9D%98-%ED%95%84%EC%9A%94%EC%84%B1

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Added relevant tests or not required
- [ ] Didn't break anything

PR #793, opened 2024-02-13
Prevent RunTestConcurrency from creating garbage clients

justiceHui:

Change `RunTestTreeConcurrency` function in `tree_concurrency_test.go` not to create and close clients in each test.

This PR can solve the performance degradation problem that occurs when vector clocks are used (#786).

**Checklist**:
- [x] Didn't break anything


hackerwins (changes requested):
Thanks for your contribution.

I think the changes in `tree.go` are not related to the current PR. Could you exclude the changes?

hackerwins (approved):
LGTM. 👍

PR #593, opened 2023-07-31
hide Watch method from Client.go and move into Attach method

karockai:
**What this PR does / why we need it**:
- make `Watch` method of `Client` to private.
- make `Attach` method to execute the `watch` method atomically after attaching.
- modify some related tests.

**Which issue(s) this PR fixes**:
Fixes #584 

**Special notes for your reviewer**:
There is one problem in this PR. After a client watching is changed to execute continuously after attaching a document, some tests in `document_test` have had an issue. Please see the scenario below.

1. the client tries to attach 'doc1'
2. the client tries to detach the doc
3. the client tries to attach another doc

There is no problem Until step 2 but step3 has. 
The err is `closing transport due to: connection error: desc = "error reading from server: EOF", received prior goaway: code: NO_ERROR` 
and server log is `2023-07-30T23:53:36.646+0900	WARN	r7	RPC : stream "/yorkie.v1.YorkieService/WatchDocument" 9.632547583s => "rpc error: code = Unknown desc = connection error: desc = \"transport is closing\""`

I said above that it wouldn't be a problem in step 2, but it really isn't. Even though error checking tells there is no err after step 2, to pass tests a time buffer should be set between step 1 and step 2, and also step 2 and step 3. The time buffer should be more than 4 sec. The test couldn't be passed when there is no time buffer between step 2 and step 3, 

After tried to solve this problem several days, I found that it is better to share this problem rather than try only myself.

further detail:
client side err is coming from 
```
# Client.watch L:482
	pbResp, err := stream.Recv()
	if err != nil {
		return nil, err
	}
``` 

server side err is coming from 
```
#yorkie_server.WatchDocument L:411
	if err := stream.Send(&api.WatchDocumentResponse{
		Body: &api.WatchDocumentResponse_Initialization_{
			Initialization: &api.WatchDocumentResponse_Initialization{
				ClientIds: pbClientIDs,
			},
		},
```
 

**Does this PR introduce a user-facing change?**:
```release-note
- Client provides `Watch` no more. `Attach` method will execute watching continuously. so remove your `Watch` from where to use.
```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Added relevant tests or not required
- [x] Didn't break anything

PR #803, opened 2024-02-28
Move Client.Watch inside Client.Attach

krapie:

Move `Client.Watch()` inside `Client.Attach()` to hide it from external interface.
This PR also introduces `Client.Subscribe()` which performs similar functionality with other SDKs. 

This PR is recreated from https://github.com/yorkie-team/yorkie/pull/593 

Address #584

I prefer keeping this PR size as small as possible to wrap up this task on time.

```release-note
Go SDK user will use `Client.Subscribe()` instead of `Client.Watch()`, just like other SDKs.
```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -78,6 +78,10 @@ var (
 type Attachment struct {
 	doc   *document.Document
 	docID types.ID
+
+	rch        <-chan WatchResponse
+	ctx        context.Context
+	cancelFunc context.CancelFunc
 }
 
 // Client is a normal client that can communicate with the server.
@@ -322,6 +326,15 @@ func (c *Client) Attach(ctx context.Context, doc *document.Document, options ...
 		docID: types.ID(res.Msg.DocumentId),
 	}
 
+	watchCtx, cancelFunc := context.WithCancel(ctx)
+	c.attachments[doc.Key()].ctx = watchCtx
+	c.attachments[doc.Key()].cancelFunc = cancelFunc
+
+	err = c.runWatchLoop(watchCtx, doc)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -384,6 +397,9 @@ func (c *Client) Detach(ctx context.Context, doc *document.Document, options ...
 	}
 	delete(c.attachments, doc.Key())
 
+	attachment.cancelFunc()
+	attachment.rch = nil
+
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


hackerwins:
Thank you for your contribution. In JS SDK, there is also a setting for SyncMode as shown in the link below. How about applying them together?

https://github.com/yorkie-team/yorkie-js-sdk/pull/772#pullrequestreview-1988728945

krapie:
Is the syncMode feature big? Like I mentioned, I want to keep this PR small.

client/client.go
```diff
@@ -406,21 +428,37 @@ func (c *Client) Sync(ctx context.Context, options ...SyncOptions) error {
...
+) (<-chan WatchResponse, context.CancelFunc, error) {
+	attachment, ok := c.attachments[doc.Key()]
+	if !ok {
+		return nil, nil, ErrDocumentNotAttached
+	}
+
+	if attachment.isSubscribed.Load() {
+		return nil, nil, ErrAlreadySubscribed
+	}
+	attachment.isSubscribed.Store(true)
```
hackerwins, line 443: I think this employs the "check-then-act" pattern, where it first checks the state and then changes it. I suspect that there could be a race condition.

How about introducing CompareAndSet, which can perform state comparison and setting atomically?

https://pkg.go.dev/sync/atomic#Bool.CompareAndSwap

client/client.go
```diff
@@ -72,12 +73,22 @@ var (
...
+	ErrAlreadySubscribed = errors.New("already subscribed")
 )
 
 // Attachment represents the document attached.
 type Attachment struct {
 	doc   *document.Document
 	docID types.ID
+
+	// TODO(krapie): assuming that a client do not open multiple subscriptions for the same document.
+	isSubscribed atomic.Bool
```
hackerwins, line 87: This comment is `TODO`, so how about writing about what we think will happen in the future?

hackerwins (changes requested):
Thanks for your contribution. 🙏
I left simple questions.

client/client.go
krapie: Oops, this will definitely create race condition. Thank you for your feedback :)

hackerwins (approved):
LGTM. 👍

hackerwins:
@krapie, @karockai Thank you for your contribution.

PR #855, opened 2024-05-10
Fix inconsistent garbage collection for multiple nodes in text and tree type

chacha912:

> There was an issue in the garbage collection process when dealing with multiple nodes in text(tree) types. When some nodes in text were collected, the corresponding text was removed from `elementHasRemovedNodesSetByCreatedAt` in the root, preventing the GC from correctly processing the remaining nodes.
>
> This PR addresses the inconsistency in the GC process by modifying the condition under which text is removed from `elementHasRemovedNodesSetByCreatedAt`. Instead of removing text when _**some**_ nodes are garbage collected, text will now only be removed when _**all**_ nodes have been garbage collected. This change ensures that GC behaves consistently even when multiple nodes are involved.

Related https://github.com/yorkie-team/yorkie-js-sdk/pull/806 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/pkg/document/crdt/root.go
+++ b/pkg/document/crdt/root.go
@@ -157,7 +157,7 @@ func (r *Root) GarbageCollect(ticket *time.Ticket) (int, error) {
 			return 0, err
 		}
 
-		if purgedNodes > 0 {
+		if node.removedNodesLen() == 0 {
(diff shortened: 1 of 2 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #907, opened 2024-06-27
Fix missing document detachments when client is deactivated

### LINKED ISSUE #758: ErrDocumentIfNotAttached occurs when trying to remove documents in Dashboard
**What happened**:

`ErrDocumentIfNotAttached` occurs when trying to remove documents in Dashboard.

스크린샷 2024-01-11 오전 11 00 03

A document should be removed if it is not attached to clients.

Try to remove any documents in Dashboard.

It is good to check the reason for the problem is `documentIfNotAttached` flag: https://github.com/yorkie-team/yorkie/pull/560

**Environment**:
- Operating system: MacOS
- Browser and version: Chrome latest
- Yorkie version (use `yorkie version`): v0.4.12
- Yorkie JS SDK version: v0.4.12

raararaara:

This PR addresses the issue where documents attached to a client are not being properly detached during deactivation. Specifically, information about document detachment is not being reflected in the database, leading to situations where documents remain attached to a deactivated client.

By implementing this fix, attempting to delete a document from the dashboard will now correctly handle the scenario where the document is attached, resolving the problem outlined in issue #758.

Address #758

The DB resources below should be updated when executing Client.Deactivate:
- Update ClientInfo.Status to 'Deactivate'
- Update ClientInfo.Documents to set ClientDocInfo.Status to 'Detached'
- Remove entry from SyncedSeq

Updating ClientInfo involves updating a single MongoDB document, ensuring ACID compliance. However, as SyncedSeq is spread across multiple documents, ACID compliance cannot be guaranteed.

In addition, I have considered the order of updates for ClientInfo and syncedSeq, along with the potential issues or advantages that may arise.

A. ClientInfo -> SyncedSeq
- If the ClientInfo update succeeds but the SyncedSeq update fails, the SyncedSeq may remain undeleted, leading to decreased GC efficiency. However, this issue can be addressed by adding correction logic to remove the SyncedSeq.

B. SyncedSeq -> ClientInfo
- If the SyncedSeq is successfully deleted but the ClientInfo update fails, the tombstone might be deleted during the next GC run. This could cause sync errors when the client tries to edit again. While this scenario is unlikely, it could lead to document corruption if it occurs.

In conclusion, for this PR, we have decided to update in the order of `A. ClientInfo -> SyncedSeq`. We plan to add correction logic later to handle cases where the ClientInfo update succeeds but the SyncedSeq update fails.

```release-note

```

**Additional documentation**:

```docs

```

### Checklist
- [x] Added relevant tests
- [x] Didn't break anything

## Summary by CodeRabbit

- **New Features**
  - Enhanced client deactivation process to ensure documents are detached before deactivation.

- **Bug Fixes**
  - Fixed issues related to document detachment when deactivating a client.

- **Tests**
  - Added tests to verify the detachment of documents when a client is deactivated.


server/backend/database/testcases/testcases.go
```diff
@@ -391,6 +391,44 @@ func RunActivateClientDeactivateClientTest(t *testing.T, db database.Database, p
...
+		assert.NoError(t, clientInfo.RemoveDocument(docInfo.ID))
+		assert.NoError(t, db.UpdateClientInfoAfterPushPull(ctx, clientInfo, docInfo))
+
+		// 04. Deactivate the client
+		result, err := db.DeactivateClient(ctx, types.ClientRefKey{
+			ProjectID: projectID,
+			ClientID:  clientInfo.ID,
+		})
+		assert.NoError(t, err)
+		assert.NoError(t, result.EnsureDocumentsNotAttachedWhenDeactivated())
```
hackerwins, line 430: This test can check whether attached documents remain, but cannot check whether a removed document remains. How about removing `EnsureDocumentsNotAttachedWhenDeactivated` and test whether the status for each document has changed as expected?

server/clients/clients.go
```diff
@@ -49,29 +49,17 @@ func Deactivate(
...
-		if err := clientInfo.DetachDocument(docID); err != nil {
-			return nil, err
-		}
-
-		// TODO(hackerwins): We need to remove the presence of the client from the document.
-		// Be careful that housekeeping is executed by the leader. And documents are sharded
-		// by the servers in the cluster. So, we need to consider the case where the leader is
-		// not the same as the server that handles the document.
-
 		if err := db.UpdateSyncedSeq(
```
hackerwins, line 63: It would be good to add two comments to this logic.

A. When deactivating a client, we need to update three properties(`ClientInfo.Status`, `ClientInfo.Documents` and `SyncedSeq`) in DB. Updating the sub-properties of `ClientInfo` guarantees atomicity as it involves a single MongoDB document. However, `SyncedSeq`s are stored in separate documents, so we can't ensure atomic updates for both. Currently, if `SyncedSeq`s update fails, it mainly impacts GC efficiency without causing major issues. We need to consider implementing a correction logic to remove SyncedSeq in the future.

B. We're currently updating `SyncedSeq` one by one. This approach is similar to n+1 query problem. We need to investigate if we can optimize this process by using a single query in the future.

hackerwins (commented):
Thanks for your contribution. I left a few comments.

server/backend/database/testcases/testcases.go
raararaara: I agree with what you suggested. It looks like I also need to check the removed case.

server/clients/clients.go
raararaara: I'll add a comment.

hackerwins (approved):
LGTM. 👍

PR #931, opened 2024-07-16
Enhance GetDocuments API by adding bulk retrieval

### LINKED ISSUE #921: Implement DB Query for GetDocuments API to improve performance

Recently, GetDocuments API was added as a bulk version of the GetDocument API to retrieve multiple documents at once. However, the necessary DB Query for the GetDocuments API has not been implemented yet. It is crucial to implement a database query that can retrieve multiple documents at once to improve performance.

https://github.com/yorkie-team/yorkie/blob/28ef053a5af5a53c2af1d0e043240adc62be1701/server/documents/documents.go#L130-L148

- https://github.com/yorkie-team/yorkie/pull/909
- https://github.com/yorkie-team/codepair/issues/173

**Why is this needed**:

To enhance performance by efficiently retrieving multiple documents with a single query from the database.

kokodak:

This PR implements a bulk retrieval operation for the GetDocuments API to enhance performance.

The specific tasks accomplished include:
- Addition of the `FindDocInfosByKeys()` method to the Database interface
- Implementation of the `FindDocInfosByKeys()` logic in `mongo.Client` and `memory.DB` respectively
- Addition of test code for mongo and in-memory implementations
- Replacement of DB queries used in GetDocument API and GetDocuments API

While the query to retrieve DocInfos has been reduced from N times to once when calling the GetDocuments API, there still remains an issue where `packs.BuildDocumentForServerSeq()` is called N times.

However, this logic seems to be related to CRDT or logical clock functionalities, which I do not fully understand yet, so I could not work on it. Therefore, I did not remove the TODO comment regarding the N+1 issue.

Fixes #921 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

## Summary by CodeRabbit

- **New Features**
    - Enhanced document retrieval by adding a method to find document information based on given keys.
    - Introduced a new field to specify whether to include snapshots in document requests.
  
- **Bug Fixes**
    - Improved efficiency of document information retrieval, addressing the N+1 query problem.

- **Documentation**
    - Updated OpenAPI specifications for improved readability and consistency.


sejongk:
Although the logic you mentioned is related to CRDT, it seems okay to understand it intuitively. 
Do you think it is possible to implement the bulk operation of `BuildDocumentForServerSeq` as well, perhaps using `$or`?

server/backend/database/testcases/testcases.go
```diff
@@ -93,6 +93,41 @@ func RunFindDocInfoTest(
...
+		assert.NoError(t, err)
+
+		actualKeys := make([]key.Key, len(infos))
+		for i, info := range infos {
+			actualKeys[i] = info.Key
+		}
+
+		assert.ElementsMatch(t, docKeys, actualKeys)
+		assert.Len(t, infos, len(docKeys))
+	})
```
hackerwins, line 128: It seems that MemDB and MongoDB implementations behave differently for non-existent keys. How about adding the corresponding test code?

hackerwins (changes requested):

server/backend/database/testcases/testcases.go
kokodak: Oh, I see. Thanks for the review. In the mongo implementation of `FindDocInfoByKey()`, it returns an `ErrDocumentNotFound` error if there is no DocInfo for the given key. This suggests that the business logic requires a DocInfo to always be present for the requested key.

Should this business rule be followed for `FindDocInfosByKeys()` as well? That is, if any key does not have a corresponding DocInfo, should it return an `ErrDocumentNotFound` error?

If the above approach is appropriate, I will write test code with this scenario.

kokodak:
> Although the logic you mentioned is related to CRDT, it seems okay to understand it roughly and focus on query optimization. Do you think it is possible to implement the bulk operation of `BuildDocumentForServerSeq` as well, perhaps using `$or`?

@sejongk 
I'll give it a try. Is it okay if I ask questions if I encounter any issues during the implementation?

server/backend/database/testcases/testcases.go
hackerwins: I think there are two approaches for handling missing documents in GetDocuments API and each have their own advantages and disadvantages.

A. Returning the remaining documents even if some are missing
- Pros: More flexible and user-friendly as it returns information for available documents even if some are missing
- Cons: Behaves differently from the `GetDocument API`

B. Returning an error for missing documents
- Pros: Keeps consistency with `GetDocument API` by returning an error for non-existent keys
- Cons: Lacks flexibility as the entire request fails if even one document is missing

Personally, I prefer option A approach.

sejongk:
> @sejongk I'll give it a try. Is it okay if I ask questions if I encounter any issues during the implementation?
Sure. If you have any suggestions about this, please let me know.

server/backend/database/testcases/testcases.go
kokodak: Listing the pros and cons has helped me consider the trade-offs.

It seems like a choice between maintaining consistency and ensuring flexibility. This decision will affect the API results, which in turn impacts the user experience.

From the user's perspective, they probably care less about whether the `GetDocuments API` is consistent with the `GetDocument API` and more about whether the document displays correctly on their screen.

With this in mind, I agree with your opinion that `approach A` is appropriate. I will make the logic of both the in-memory and mongo implementations follow the same flow as `approach A` and add tests accordingly.

hackerwins:
> Although the logic you mentioned is related to CRDT, it seems okay to understand it roughly and focus on query optimization. Do you think it is possible to implement the bulk operation of BuildDocumentForServerSeq as well, perhaps using $or?
> 
> @sejongk
> I'll give it a try. Is it okay if I ask questions if I encounter any issues during the implementation?

DocumentSummaries, which is in the response from `GetDocuments API`, contains both `time-related metadata` about Document and its content, `snapshot`. Unlike metadata, retrieving `snapshot` requires loading the document into memory, which can be relatively resource-intensive.

Document List Page in CodePair, which uses this API, only uses the `time-related metadata` and not `snapshot`.

Therefore, how about adding an option(`include snapshot`) in the API request to specify whether snapshot should be included.

sejongk:
> > Although the logic you mentioned is related to CRDT, it seems okay to understand it roughly and focus on query optimization. Do you think it is possible to implement the bulk operation of BuildDocumentForServerSeq as well, perhaps using $or?
> > @sejongk
> > I'll give it a try. Is it okay if I ask questions if I encounter any issues during the implementation?
> 
> @kokodak @sejongk
> 
> DocumentSummaries, which is in the response from `GetDocuments API`, contains both `time-related metadata` about Document and its content, `snapshot`. Unlike metadata, retrieving `snapshot` requires loading the document into memory, which can be relatively resource-intensive.
> 
> Document List Page in CodePair, which uses this API, only uses the `time-related metadata` and not `snapshot`.
> 
> https://www.figma.com/design/OYc1Cr0nvFuBnWZxhscfDk/Code-Pair?node-id=42-101&t=lCXENp1HuDnFAkwq-0
> 
> Therefore, how about adding an option(`include snapshot`) in the API request to specify whether snapshot should be included.

Thanks for your suggestion. I believe this suggested method is related to https://github.com/yorkie-team/yorkie/pull/597.

kokodak:
I have reviewed all the comments provided.

Currently, I have completed the implementation of bulk query methods for `DB.FindClosestSnapshotInfo()` and `DB.FindChangesBetweenServerSeqs()`, which are used in `BuildDocumentForServerSeq()`.

However, I am facing some issues and need help with the following:

1. Although the bulk query operations are implemented, I am having difficulty writing test cases. Creating good test scenarios is challenging. Could I get some help with this?

2. I generally understand the context of @hackerwins comment, but I am a bit unclear about the exact meaning of "snapshot" since the term is used in several places in the code.
If the request value for `include snapshot` is `false`, does it mean that `DB.FindClosestSnapshotInfo()` should be called with `includeSnapshot` set to `false`, or does it mean that `packs.BuildDocumentForServerSeq()` should not be executed at all? (I am inclined to believe it's the latter.)
- 2-a. If the latter is correct, should we still keep the bulk query code mentioned in `point 1`, to handle cases where `include snapshot` is `true`?

3. I also agree that passing only the minimal information needed to render the screen is a good idea. However, if it turns out that snapshots will never be used in the `GetDocuments API`, we might consider configuring the code to exclude snapshots without adding an option to the API request. What are your thoughts on this? Should we still include the option in the request for flexibility?
- 3-a. If we decide to include the option, we will need to coordinate with the front-end regarding the changes in the API structure. How should we approach this discussion?

hackerwins (approved):

kokodak:
> > Although the logic you mentioned is related to CRDT, it seems okay to understand it roughly and focus on query optimization. Do you think it is possible to implement the bulk operation of BuildDocumentForServerSeq as well, perhaps using $or?
> > @sejongk
> > I'll give it a try. Is it okay if I ask questions if I encounter any issues during the implementation?
> 
> @kokodak @sejongk
> 
> DocumentSummaries, which is in the response from `GetDocuments API`, contains both `time-related metadata` about Document and its content, `snapshot`. Unlike metadata, retrieving `snapshot` requires loading the document into memory, which can be relatively resource-intensive.
> 
> Document List Page in CodePair, which uses this API, only uses the `time-related metadata` and not `snapshot`.
> 
> https://www.figma.com/design/OYc1Cr0nvFuBnWZxhscfDk/Code-Pair?node-id=42-101&t=lCXENp1HuDnFAkwq-0
> 
> Therefore, how about adding an option(`include snapshot`) in the API request to specify whether snapshot should be included.

Based on the discussions with @hackerwins and @sejongk regarding the comment ideas above, we have decided to implement the option to include or exclude snapshots in the API request.

As a result, the GetDocuments API request specification has changed, which can be reviewed in [this commit](https://github.com/yorkie-team/yorkie/pull/931/commits/327e91dfebb316308229be8476e5229541885b1f).

https://github.com/yorkie-team/codepair/blob/e1cebe0ec3304b2ad8cdad94926a25fadef39da3/backend/src/workspace-documents/workspace-documents.service.ts#L165-L176

Consequently, by adding the `include_snapshot` field with a value of `false` in the CodePair code, we can expect performance improvements in the GetDocuments API.

PR #934, opened 2024-07-20
Support basic account action for admin

### LINKED ISSUE #849: Support Basic Account Action for Admin

The AdminService currently only includes the SignUp and LogIn API endpoints. To enhance the functionality and provide basic account actions for administrators, additional actions such as SignOut and ChangePassword need to be implemented.

```yaml
service AdminService {
  rpc SignUp(SignUpRequest) returns (SignUpResponse) {}
  rpc LogIn(LogInRequest) returns (LogInResponse) {}

  # Needs to be implemented.
  rpc SignOut(SignOutRequest) returns (SignOutResponse) {}
  rpc ChangePassword(SignOutRequest) returns (SignOutResponse) {}
```

**Why is this needed**:

Expanding the basic account actions available through the AdminService will improve the user experience and provide necessary functionality for administrative tasks.

gusah009:

Added the admin APIs ChangePassword and DeleteAccount.

As their names suggest, these are APIs for changing passwords and deleting accounts.

Fixes #849 

- I've verified that it works with mongoDB and in-memory DBs using postman.

- As mentioned below, we will need to open an endpoint to access this API from your dashboard or CLI. client development is not included in this PR.

- I generated the files with the `make proto` command, which broke the formatting of the `api/docs/yorkie/v1/*.openapi.yaml` files.

- I wanted the validation of `username` or `password` on `ChangePassword` to be the same as `SignUp`, but I didn't know how to separate them into a common string, so I added a warning as a comment. If there's a good way to do it, I'd love to hear about it :)

```release-note
NONE
```
There is no user impact yet. we need to open an endpoint that the client can access from the dashboard or CLI.

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Didn't break anything

## Summary by CodeRabbit

- **New Features**
	- Added functionalities for users to change their passwords, create new accounts, and delete accounts via new API endpoints.
	- Implemented corresponding methods on the server side for secure account management operations.
	- Marked the previous signup endpoint as deprecated to facilitate the transition to the new account creation process.

- **Bug Fixes**
	- Improved error handling for account management operations to ensure accurate feedback during deletions and password changes.

- **Documentation**
	- Updated API documentation to reflect the new endpoints and their corresponding request/response schemas.


api/types/user_fields.go
```diff
@@ -22,6 +22,7 @@ import (
 )
 
 // SignupFields is a set of fields that use to sign up to yorkie server.
+// // must be same with [ChangePasswordFields].
```
krapie, line 25: Comment mark duplicated in this line!

api/types/user_fields.go
```diff
@@ -22,6 +22,7 @@ import (
...
 type SignupFields struct {
```
krapie, line 26: How about changing `SignupFields` to `UserFields` just like file name you changed?

api/yorkie/v1/admin.proto
```diff
@@ -28,6 +28,8 @@ option java_package = "dev.yorkie.api.v1";
 service AdminService {
   rpc SignUp(SignUpRequest) returns (SignUpResponse) {}
   rpc LogIn(LogInRequest) returns (LogInResponse) {}
+  rpc DeleteAccount(DeleteAccountRequest) returns (DeleteAccountResponse) {}
```
krapie, line 31: I'm thinking of changing `SignUp` to `CreateAccount` or other ways to keep consistency of RPC names. How do you think about this?

server/backend/database/database.go
```diff
@@ -143,6 +143,12 @@ type Database interface {
 	// after handling PushPull.
 	UpdateClientInfoAfterPushPull(ctx context.Context, clientInfo *ClientInfo, docInfo *DocInfo) error
 
+	// DeleteUserInfoByName deletes a user by name.
```
krapie, line 146: How about moving this function near User related functions like `CreateUserInfo()`?

server/backend/database/database.go
```diff
@@ -143,6 +143,12 @@ type Database interface {
...
+	DeleteUserInfoByName(ctx context.Context, username string) error
+
+	// ChangePassword changes to new password.
+	ChangePassword(ctx context.Context, username, hashedNewPassword string) error
```
krapie, line 150: How about adding the word `User` for better understanding of the function?

server/backend/database/mongo/client.go
```diff
@@ -643,6 +643,35 @@ func (c *Client) UpdateClientInfoAfterPushPull(
...
+// DeleteUserInfoByName deletes a user by name.
+func (c *Client) DeleteUserInfoByName(ctx context.Context, username string) error {
+	deleteResult, err := c.collection(ColUsers).DeleteOne(ctx, bson.M{
+		"username": username,
+	})
+	if err != nil {
+		return err
+	}
+	if deleteResult.DeletedCount == 0 {
+		return fmt.Errorf("nothing has been deleted")
```
krapie, line 655: How about enhancing error message like `fmt.Errorf("no user found with username %s", username)`?
We want to keep consistency for all the error messages in this file. Refer to other error messages for information.

server/backend/database/mongo/client.go
```diff
@@ -643,6 +643,35 @@ func (c *Client) UpdateClientInfoAfterPushPull(
...
+func (c *Client) ChangePassword(ctx context.Context, username, hashedNewPassword string) error {
+	updateResult, err := c.collection(ColUsers).UpdateOne(ctx,
+		bson.M{"username": username},
+		bson.M{"$set": bson.M{"hashed_password": hashedNewPassword}},
+	)
+	if err != nil {
+		return err
+	}
+	if updateResult.ModifiedCount == 0 {
+		return fmt.Errorf("nothing has been updated")
```
krapie, line 670: > How about enhancing error message like `fmt.Errorf("no user found with username %s", username)`?
We want to keep consistency for all the error messages in this file. Refer to other error messages for information.

Same here.

krapie (changes requested):
I have two things to suggest:

1. User vs. Account
As you can see in the database, we are using the word `User`, so we might need to keep consistency for the word describing "user".

2. Test codes for Admin RPC
How about adding test codes for newly introduced APIs? `TestAdminRPCServerBackend()` test function in `server/rpc/server_test.go` will be the best place to put test codes.

api/types/user_fields.go
gusah009: That's a really good idea :)

api/yorkie/v1/admin.proto
gusah009: I think it would be better to use User instead of Account as @krapie suggested.

So I'm going to use `CreateUser` instead of `SignUp`. thank you for comment! 👍👍

server/backend/database/mongo/client.go
gusah009: oh that's good idea. I miss that thank you! :)

krapie:
@gusah009 There seems to be lint error causing CI to fail. Could you please lint it again?

gusah009:
@krapie hello. I agree with all of your suggestions in the comments, so I've made the following changes.

1. In response to `User vs Account`, I have unified to use `User`. ex) `DeleteAccount` -> `DeleteUser`

2. I also agree that the Admin RPC needs test code, so I added a test case that is simple but covers a few cases.

In addition, I also included the review. Thank you for your careful review.

server/backend/database/database.go
krapie: Well it seems like the lines are still in the same position. Not planning on changing the lines?

krapie (commented):
Now I'm a bit confused by the word `RunAdminCreateUserTest`. Does this mean admin can create other user or create itself(admin)? `RunAdminSignUpTest` makes more sense in terms of creating itself(admin).

Sorry, but I think we should reconsider which word to use. Using the word `User` may confuse others between admin creating user(not admin), or creating itself(admin).

Any ideas? @hackerwins

krapie:
Also, I did not get what below comment means:

> I wanted the validation of username or password on ChangePassword to be the same as SignUp, but I didn't know how to separate them into a common string, so I added a warning as a comment. If there's a good way to do it, I'd love to hear about it :)

server/backend/database/database.go
gusah009: oh my mistake. I only moved methods related to the memory DB. Here is the commit: https://github.com/yorkie-team/yorkie/pull/934/commits/9ab114271c7ba2262b579e91cd04e5bc315e5cb8

I will fix it right now.

krapie:
@gusah009 How about changing the word `user` to `account`? If we use the word `user` in the context of `admin`, it sounds like the action is for `normal user`, not `admin user`. So if we use the word `account` instead, we can clarify the meaning; that these RPCs performs action on the admin itself.

It will look something like below:

```proto
service AdminService {
  rpc CreateAccount(CreateAccountRequest) returns (CreateAccountResponse) {}
  rpc LogIn(LogInRequest) returns (LogInResponse) {}
  rpc DeleteAccount(DeleteAccountRequest) returns (DeleteAccountResponse) {}
  rpc ChangePassword(ChangePasswordRequest) returns (ChangePasswordResponse) {}
}
```

gusah009:
@krapie Oh, that's a good idea. I'll do it right away.

krapie:
@gusah009 Hmm. I'm not sure about changing `username` to `accountName`. I think we can organize like below:

- `Account` is an object which contains `Username`, `Password` fields.
- `Account` parent of `Admin` and `User` (Normal User, TBD).

gusah009:
> Account parent of Admin and User (Normal User, TBD).

@krapie Looking at it like this, I don’t think I need to change the username. Thank you for your good advice.

krapie:
After discussing with @gusah009, we have concluded what we revert our changes to use RPC name as `DeleteAccount` and `ChangePassword`. This is because the scope of changing keyword `User` to `Account` takes more resources and changes that we thought. Considering our scope of this PR of just adding new actions for admin, it will be better to keep our changes small as possible.

I apologize for the frequent change request to @gusah009. Thank you for your understanding.
Next time we will decide our design prior to implementation.

krapie (approved):
LGTM 👍🏼

krapie:
@hackerwins We did not consider cascading delete on user related objects including projects and documents. I think we need to consider to also delete related resources.

hackerwins:
@krapie Please create an issue regarding this. 🙏

PR #935, opened 2024-07-21
Update 'updated_at' only if there are more than one operation in changes

### LINKED ISSUE #916: Update Document updatedAt to only include Document.Root modifications
**Description**:

Currently, modification to `Document.Presence` updates the Document's updatedAt, even if there are no changes to `Document.Root` content. This behavior can lead to unnecessary `updatedAt` updates and may not accurately reflect the actual content changes.

**Why**:

window9u:

The current implementation updates the 'updated_at' field of a document even if the content itself remains unchanged. This happens because changes reported by the client may contain both "presence" and "operations." With the changes introduced in this commit, the 'updated_at' timestamp will only be updated when there is more than one operation in the change list.

Initially, I attempted to address this issue by avoiding database operations when the changes were limited to presence updates. However, this approach was unsuccessful for the following reasons:

- Changes that include only presence still carry a ClientSeq. If these are not recorded, it can lead to inconsistencies between the client and the server states.

Consequently, it is necessary to store all changes in the database, even if they do not modify the core content, to maintain consistency.

In the context of applications like a whiteboard, where numerous operations can occur simply from moving a mouse pointer, this issue becomes more pronounced. Such frequent minor updates lead to a high volume of changes that primarily involve presence, inflating the ClientSeq and triggering server snapshots, which add to the server's load.

#### Proposed Solution

Initially, I considered separating presence data from document updates and managing it through an in-memory database like Redis. This would involve removing presence data from the push-pull synchronization mechanism and introducing a new API dedicated to presence management. However, this approach would require extensive modifications to the existing codebase and protocols, affecting not just the current system but other SDKs as well.

Given the complexities involved, I recommend keeping the current push-pull mechanism unchanged for stability and introducing a new API dedicated to broadcasting user presence. This strategy is ideal for applications where excessive updates are driven primarily by presence changes. Developers can shift presence management from the standard update method to this new broadcast API.

This broadcast API is specifically designed to manage presence updates independently. It broadcasts changes to all document subscribers efficiently without modifying the sequence number (ClientSeq) or permanently storing the updates. This design ensures optimal performance by minimizing unnecessary data load and sequence adjustments, making it highly suitable for managing frequent, low-impact changes.

Fixes #916

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

## Summary by CodeRabbit

- **New Features**
	- Enhanced logic for updating the `UpdatedAt` timestamp, now only updating when there are actual operations present.

- **Tests**
	- Added a new test case to verify the behavior of the `UpdatedAt` timestamp based on document updates and operations.

- **Bug Fixes**
	- Corrected the conditions under which the `UpdatedAt` field is updated, ensuring more accurate timestamp management.

### DIFF
--- a/server/backend/database/memory/database.go
+++ b/server/backend/database/memory/database.go
@@ -900,7 +900,14 @@ func (d *DB) CreateChangeInfos(
 
 	now := gotime.Now()
 	loadedDocInfo.ServerSeq = docInfo.ServerSeq
-	loadedDocInfo.UpdatedAt = now
+
+	for _, cn := range changes {
+		if len(cn.Operations()) > 0 {
+			loadedDocInfo.UpdatedAt = now
+			break
+		}
+	}
+
 	if isRemoved {
 		loadedDocInfo.RemovedAt = now
 	}
--- a/server/backend/database/mongo/client.go
+++ b/server/backend/database/mongo/client.go
@@ -891,8 +891,15 @@ func (c *Client) CreateChangeInfos(
 	now := gotime.Now()
 	updateFields := bson.M{
 		"server_seq": docInfo.ServerSeq,
-		"updated_at": now,
 	}
+
+	for _, cn := range changes {
+		if len(cn.Operations()) > 0 {
+			updateFields["updated_at"] = now
+			break
+		}
+	}
+
 	if isRemoved {
 		updateFields["removed_at"] = now
 	}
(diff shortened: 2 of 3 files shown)


hackerwins (approved):
Thanks for your contribution. I'll apply CodeRabbit's comments after merging this PR.

PR #938, opened 2024-07-22
Show Server Version in Yorkie CLI

### LINKED ISSUE #870: Show Server Version in Yorkie CLI
**Description**:

Add the functionality to display the server version in Yorkie CLI alongside the existing information about the Yorkie client version and so on.

```bash
$ yorkie version
Yorkie: 0.4.0
Commit: 782f93d
Go: go1.20.4
Build date: 2023-06-05
```
[kubectl](https://kubernetes.io/docs/reference/kubectl/) shows bellow information including server version which kubectl is currently targeting.

```bash
$ kubectl version
Client Version: v1.28.0
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.29.4-eks-036c24b
```

**Why**:

Having the ability to see the server version directly in the Yorkie CLI helps users easily track the server version they are interacting with, providing crucial information for debugging and maintenance purposes.

hyun98:

Add the functionality to display the version of connected Yorkie server in the `version` command

Fixes #870 

- The version information can be viewed by converting it into `yaml` or `json` format using the `--output`, `-o` flag.
  This flag accepts `json` and `yaml` strings and converts them to the appropriate format for display.
- In order to solely check the version of the client CLI without considering the server's version, the --client flag has been included.
- The development of this feature was inspired by examining the [`kubectl`](https://github.com/kubernetes/kubectl) code.

```release-note
NONE
```

**Additional documentation**:

```docs
$ yorkie version -h

Print the version number of Yorkie

Usage:
  yorkie version [flags]

Flags:
      --client          Shows client version only. (no server required)
  -h, --help            help for version
  -o, --output string   One of 'yaml' or 'json'.

Global Flags:
      --rpc-addr string   Address of the rpc server (default "localhost:8080")
```

**Checklist**:
- [ ] Didn't break anything

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method to retrieve the server version, providing users with version details including Yorkie version, Go version, and build date.
	- Enhanced the version command functionality to support output format validation and retrieval of the server version.
	- Improved output formatting options for the version command, allowing users to choose between standard, YAML, or JSON outputs.

- **Bug Fixes**
	- Improved error handling during server version retrieval, ensuring clearer feedback for users.

- **Tests**
	- Added new test cases to validate the functionality of the server version retrieval method.


cmd/yorkie/version.go
```diff
@@ -17,27 +17,142 @@
...
 
 	"github.com/spf13/cobra"
-
+	"github.com/spf13/viper"
+	"github.com/yorkie-team/yorkie/admin"
+	"github.com/yorkie-team/yorkie/api/types"
+	"github.com/yorkie-team/yorkie/cmd/yorkie/config"
 	"github.com/yorkie-team/yorkie/internal/version"
+	"gopkg.in/yaml.v3"
+)
```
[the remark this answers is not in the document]
hackerwins: @hyun98 
The CI build seems to be broken because of this issue. Please confirm.

hackerwins (commented):
The code in the PR generally looks good, but there are a few edge cases that need to be addressed. Currently, the behavior varies in different scenarios:

A. When the CLI has an authenticated context:

The user can log in, view contexts, and see both client and server version information.

```sh
$ yorkie login -u admin -p admin --insecure
$ yorkie context ls
 CURRENT  RPC ADDR        INSECURE  TOKEN
 *        localhost:8080  true      eyJhbGciOi...hxtu4u7_jE
$ yorkie version
Yorkie Client: 0.4.27
Go: go1.22.1
Build Date: 2024-07-23
Yorkie Server: 0.4.27
Go: go1.22.1
Build Date: 2024-07-23
```

B. When the CLI has no context:

The user can see client version information, but receives an error when trying to fetch the server version.

```
$ yorkie logout
$ yorkie version
Yorkie Client: 0.4.27
Go: go1.22.1
Build Date: 2024-07-23
Error fetching server version: auth for localhost:8080 does not exist
```

C. When the CLI requests version information from an old server:

The user sees client version information but gets an authentication error for the server version.

```
$ yorkie version --rpc-addr api.yorkie.dev:443
Yorkie Client: 0.4.27
Go: go1.22.1
Build Date: 2024-07-23
Error fetching server version: auth for api.yorkie.dev:443 does not exist
```

D. Questions and suggestions for improvement:

Q) Should we allow users to view the server version without authentication? This could be beneficial for troubleshooting and version compatibility checks.

Q) How should we handle different scenarios?

- When the server endpoint is incorrect or the server is not running: We should provide a clear error message indicating connection issues.
- When the server is running an older version without the version API: We should gracefully handle this case, perhaps by showing the client version and indicating that the server version is unavailable or outdated.

hyun98:
First of all, thank you very much for your review.

> D. Questions and suggestions for improvement:
> 
> Q) Should we allow users to view the server version without authentication? This could be beneficial for troubleshooting and version compatibility checks.

A) I agree that checking the server version without authentication has several advantages. So, I would like to add a flag that can check the version of a specific addr without authentication. For example, options like `--no-auth`. Would this be a good option name? I always worry about naming.
Additionally, I'm asking just in case, but I understand that you don't mean to remove the authentication itself for version checking, right?

> Q) How should we handle different scenarios?
> 
> * When the server endpoint is incorrect or the server is not running: We should provide a clear error message indicating connection issues.

A) Thank you. If the server version is not loaded properly, I need to modify it to display a more appropriate error message.

> * When the server is running an older version without the version API: We should gracefully handle this case, perhaps by showing the client version and indicating that the server version is unavailable or outdated.

A) Likewise, we need to show an appropriate message regarding the case you mentioned. For example, Phrases like: 'server version checking is only supported if the server version is 0.4.27 or higher.'


Lastly, there was an error in my code. I tried to use auth, but there was a problem with only loading auth that does not fit the current context because `config.Preload` was not performed, so I will fix it together.

Thank you :)

hackerwins:
> A) I agree that checking the server version without authentication has several advantages. So, I would like to add a flag that can check the version of a specific addr without authentication. For example, options like --no-auth. Would this be a good option name? I always worry about naming.
Additionally, I'm asking just in case, but I understand that you don't mean to remove the authentication itself for version checking, right?

It would be better to remove authentication entirely for checking the server version. Version is often considered public information and it improves the experience for developers who need to quickly check version. I am also curious about whether kubectl's version command checks authentication. 🤔

hyun98:
> It would be better to remove authentication entirely for checking the server version. Version is often considered public information and it improves the experience for developers who need to quickly check version. I am also curious about whether kubectl's version command checks authentication. 🤔

That's a relief, I asked just in case. Now I understood your intentions correctly. 
I will try to remove authentication with the version check command. I think removing authentication would be a better experience during the development process.

However, there is one thing that concerns me. I don't know much about security, but I wonder if there will be any security issues due to server version exposure. 🤔

Additionally, `kubectl` always goes through an authentication process to obtain the server's version information (only when there is no `--client` option). To obtain version information, the context and auth existing in `$HOME/.kube/config` are used. This is because the k8s api-server is accessed through information.

krapie:
I also think there can be a potential threat when

krapie:
When thinking about who will be using this feature, I think the internal developers who combined Yorkie with their system will use this feature to check or troubleshoot something. So I'm not sure about the necessity of exposing this feature to public.

> However, there is one thing that concerns me. I don't know much about security, but I wonder if there will be any security issues due to server version exposure. 🤔

I think there can be a malicious attack by penetrating the weakness of the certain server version exposed by this feature.

cmd/yorkie/version.go
```diff
@@ -17,27 +17,152 @@
...
+
+			if serverErr != nil {
+				cmd.Printf("Error fetching server version: ")
+				if strings.Contains(serverErr.Error(), "unimplemented") {
+					cmd.Printf("The server does not support this operation. You might need to check your server version.\n")
+				} else {
+					cmd.Print(serverErr)
+				}
+			}
+
```
hyun98, line 128: When calling the version command, the grpc response code received from the server was unknown, which led to difficulties in error filtering.
So, I added it because I thought a detailed message would be most needed if the server does not support the version command.
:confused: This part also looks bad, judging by the `unimplemented` keyword in the error message. Isn't there a better way?

krapie:
I will look into this PR too.

cmd/yorkie/version.go
krapie: @hyun98 How about following the process or error handling in other CLI commands like `login.go`.
Since this PR is about adding server version check features to CLI, lets follow the conventions for now.
After this PR is merged, we can talk about the current error handling of CLI commands and improve that.

cmd/yorkie/version.go
```diff
@@ -17,27 +17,162 @@
...
+			errorChan := make(chan error)
+
+			if !clientOnly {
+				go func() {
+					rpcAddr := viper.GetString("rpcAddr")
+					auth, err := config.LoadAuth(rpcAddr)
+					if err != nil {
+						errorChan <- err
+						return
+					}
```
hyun98, line 65: @krapie I kept the go routine code. Even if an error occurs while retrieving the server version, I think it is good that it does not affect the existing behavior of showing the client version.

cmd/yorkie/version.go
```diff
@@ -17,27 +17,162 @@
...
+					// for example, rpc error handling reference: https://connectrpc.com/docs/go/errors/
+					switch connectErr.Code() {
+					case connect.CodeUnimplemented:
+						cmd.Printf("The server does not support this operation. You might need to check your server version.\n")
+					}
+				} else {
+					cmd.Print(serverErr)
+				}
+			}
+
```
hyun98, line 138: I referred to the connect-rpc documentation as suggested by @krapie and displayed a user-friendly message when the yorkie server could not response the updated version command.

In my opinion, if this pr is released and someone's yorkie client version is updated, the user will not know the cause if the version command only shows `unimplemented: 404 Not Found`(ambiguous default error message).
Therefore, only `unimplemented` errors that were likely to have a significant impact on users were handled as separate messages.

Finally, I left an additional TODO comment, and it would be good to work on improving the remaining error cases and rpc exception handling module in another issue.

cmd/yorkie/version.go
```diff
@@ -17,27 +17,162 @@
...
 		RunE: func(cmd *cobra.Command, args []string) error {
-			fmt.Printf("Yorkie: %s\n", version.Version)
-			fmt.Printf("Go: %s\n", runtime.Version())
-			fmt.Printf("Build date: %s\n", version.BuildDate)
+			if err := Validate(); err != nil {
+				return err
+			}
+
+			var versionInfo types.VersionInfo
+			versionInfo.ClientVersion = getYorkieClientVersion()
```
krapie, line 53: How about extracting the code that retrieves server version to the method with the name `getYorkieServerVersion()`, just like client to keep consistency and code readability?

krapie (changes requested):
Thank you for your contribution 😄
I have left some comments regarding the code style and etc.

FYI: After this PR merges, you might have to update corresponding [Yorkie CLI Docs](https://yorkie.dev/docs/cli) for this CLI feature.

cmd/yorkie/version.go
hackerwins: The current implementation using goroutines doesn't seem to provide significant benefits:

1. Lack of asynchronous advantage: The code waits for the server response before outputting version info, so there's no real benefit from asynchronous processing.

2. Timeout handling: We can handle timeouts using context, eliminating the need for a separate goroutine.

3. Unnecessary complexity: The use of goroutines and channels adds complexity without clear benefits.

4. Error handling: The current structure complicates error handling, which could be simpler with a synchronous approach.

I suggest refactoring to a synchronous implementation, which would make the code simpler and more straightforward.

hackerwins (commented):
Thank you for your work on this PR. I have a suggestion regarding the use of goroutines.

admin/client.go
```diff
@@ -354,6 +354,21 @@ func (c *Client) ListChangeSummaries(
 	return summaries, nil
 }
 
+// GetServerVersion gets the server version.
+func (c *Client) GetServerVersion(ctx context.Context,
+) (*types.VersionDetail, error) {
```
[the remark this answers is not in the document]
hyun98: Oh, It seems that the indentation format is used when there are three or more parameters. I'll fix it!

hyun98:
Thanks for your comment. 
I agree that the current code lacks the advantage of goroutines. 
I'll remove it as it increases complexity such as error handling... without any significant advantage.
Thanks to this, it gave me an opportunity to think more about using goroutines.

hackerwins (approved):
Looks good overall. Thanks for your contribution.

Issue #805, opened 2024-03-04
Validate Checkpoint in ChangePack for PushPull API requests


hackerwins:
**Description**:

When the Client calls the `PushPull API` and sends the request `ChangePack`, `Checkpoint` in the request may be tampered with due to various reasons such as network delays causing duplicate requests, new SDK bugs, or intentional tampering by malicious clients.

Therefore, it is beneficial for the stability and security of the system to validate `Checkpoint`.

Consider the following validation checks:

- `Change`s in the request `ChangePack` passed to `PushPull API` are created by a single Client, so `Change.ID.Checkpoint.ClientSeq` should increment sequentially by one.
- `Checkpoint.ServerSeq` in the request `ChangePack` for `PushPull API` cannot be greater than the server's `Checkpoint.ServerSeq` since it is set when the server saves the Change to the database.

If the Checkpoint is invalid, consider the following exception handling:

- If duplicate requests caused by network delays treat them as `OK`
- For other reasons, return `INVALID_ARGUMENT`

**Why**:

This validation will help ensure the integrity and security of the system.

binary-ho:
I'm interested in this issue. may I try this?

sejongk:
> I'm interested in this issue. may I try this?

@binary-ho Sure. If you have any questions, feel free to ask.

binary-ho:
Hello, I have some questions while trying to resolve the issue.

- [Issue: Validate Checkpoint in ChangePack for PushPull API requests](https://github.com/yorkie-team/yorkie/issues/805)
- [Reference: PushPull function in yorkie_server.go file](https://github.com/yorkie-team/yorkie/blob/main/server/rpc/yorkie_server.go)

## 1. Where do we obtain the server's ClientSeq and ServerSeq?
To resolve this issue, I believe comparing the ClientSeq and ServerSeq included in the client request with the ClientSeq and ServerSeq stored by the server in the Document of the project would be appropriate.

I think we can use the ClientSeq and ServerSeq of the `Checkpoint` in the `ClientInfo` struct, which can be obtained through the `ClientId` of the `Request Message`. Is my understanding correct?

<br> <br>

## 2. Where do we obtain the client's ClientSeq and ServerSeq?
The `ChangePack` of the Request Msg contains a `Checkpoint` and an array of `Changes`.
- The `Checkpoint` has its own ClientSeq and ServerSeq.
- Each `Change` in `Changes` also has its own ClientSeq and ServerSeq in its `ChangeID`.

[...]

PR #947, opened 2024-07-31
Add Option for Initializing Document

### LINKED ISSUE #669: Provide an interface to set the initial value when creating a document
**Description**:

Currently, to create a document and set its initial value, we have to use `doc.update()` as follows:

https://github.com/yorkie-team/yorkie-js-sdk/blob/3b313a54eee2b20e0185f93813643479714a8244/public/index.html#L152-L160

It would be convenient to provide an interface to set the initial value when creating the document, as shown below:

```js
const doc = new yorkie.Document('codemirror');
await client.attach(doc, {
  initialRoot: { content: new yorkie.Text() }
});
```

Related to https://github.com/yorkie-team/yorkie/issues/663

**Why**:

window9u:

- Provides an option for initializing a document during creation.
- When a client attempts to attach to a document created with this option, and there is no existing document, the changes will be applied.
- If there is an existing document, the operations in the changes will be removed, and the server will always return a snapshot.
- This approach eliminates the need for the client to rollback; instead, they can simply apply the snapshot.
-  This change is not applied to every attaching document with changes. it doesn't conflict with this test code [Code reference](https://github.com/yorkie-team/yorkie/blob/main/test/integration/document_test.go#L46)
- It is valid only when the first change has the "initialDoc" message and an existing document already exists.

Fixes #669

#### Handling Client Rollback with Snapshot

To handle failures, we need to implement Client's Rollback.

In the current implementation, there is no rollback for invalid changes. If the initial document fails during attachment, we can roll back the changes by resetting the root, as there was no document initially.

However, we need to consider the following situation:

- Resetting the client's root directly could cause errors and be difficult to manage.
- If a client detaches and then re-attaches a document after normal interaction, resetting the root alone is insufficient for rollback.

**Solution: Always applying a Snapshot from the server.** By applying a snapshot, we can handle rollbacks without changing the Client code, allowing the existing logic to proceed as intended.

#### Handling Failure `Change`

**Possible Solutions:**

1. **Marking Invalid Changes:** Add a special marker to the change message, such as `invalid_init`.
    - **Advantage:** Makes debugging easier.
    - **Disadvantage:** Every component that applies and evaluates changes must handle `invalid_init` on both the server and client sides.

2. **Server Deletion of Invalid Changes:** The server can remove operations related to invalid changes.
    - **Advantage:** Eliminates the need for additional logic to check for invalid changes.
    - **Disadvantage:** Causes the server to mutate the changes.

I chose option 2 for simpler implementation. We could mitigate this by adding a `bool valid` field to the Change, but I'm not certain this approach is correct. Additionally, if there are no plans to check for invalid changes, adding a field for this feature may be excessive.

```release-note
Users can use this option to initialize a document, but changes will not be applied if a document already exists.
```

**Additional documentation**:

```docs

```

**Checklist**:
- [ ] Didn't break anything

## Summary by CodeRabbit

- **New Features**
	- Introduced a method to clear operations from the Change struct, enhancing its functionality.
	- Added an InitialDoc field to the Options struct for more flexible document initialization.

- **Bug Fixes**
	- Enhanced error handling during document instantiation across various tests and functions, improving robustness and reliability.

- **Documentation**
	- Improved clarity by capturing errors during document creation in multiple test cases.

- **Refactor**
	- Updated test cases to ensure document creation errors are handled before proceeding with assertions.

### DIFF
--- a/api/converter/converter_test.go
+++ b/api/converter/converter_test.go
@@ -36,8 +36,9 @@ import (
 
 func TestConverter(t *testing.T) {
 	t.Run("snapshot simple test", func(t *testing.T) {
-		doc := document.New("d1")
-		err := doc.Update(func(root *json.Object, p *presence.Presence) error {
+		doc, err := document.New("d1")
+		assert.NoError(t, err)
+		err = doc.Update(func(root *json.Object, p *presence.Presence) error {
 			root.SetNewText("k1").Edit(0, 0, "A")
 			return nil
 		})
@@ -60,9 +61,10 @@ func TestConverter(t *testing.T) {
 	})
 
 	t.Run("root snapshot test", func(t *testing.T) {
-		doc := document.New("d1")
+		doc, err := document.New("d1")
+		assert.NoError(t, err)
 
-		err := doc.Update(func(root *json.Object, p *presence.Presence) error {
+		err = doc.Update(func(root *json.Object, p *presence.Presence) error {
 			// an object and primitive types
 			root.SetNewObject("k1").
 				SetNull("k1.0").
@@ -132,9 +134,10 @@ func TestConverter(t *testing.T) {
 	})
 
 	t.Run("change pack test", func(t *testing.T) {
-		d1 := document.New("d1")
+		d1, err := document.New("d1")
+		assert.NoError(t, err)
 
-		err := d1.Update(func(root *json.Object, p *presence.Presence) error {
+		err = d1.Update(func(root *json.Object, p *presence.Presence) error {
 			// an object and primitive types
 			root.SetNewObject("k1").
 				SetBool("k1.1", true).
@@ -189,7 +192,8 @@ func TestConverter(t *testing.T) {
 		assert.NoError(t, err)
 		pack.MinSyncedTicket = time.MaxTicket
 
-		d2 := document.New("d1")
+		d2, err := document.New("d1")
+		assert.NoError(t, err)
 		err = d2.ApplyChangePack(pack)
 		assert.NoError(t, err)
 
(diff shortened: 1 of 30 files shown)


window9u:
This PR addresses a simple issue of adding an initializing interface. However, considering scenarios where a document needs to be initialized and attached while an existing document is already present, this PR has expanded in scope.

1. **Add Interface for Initializing Documents**
    - [Commit 07f2658](https://github.com/yorkie-team/yorkie/pull/947/commits/07f265807ba98e8b83bccbb0a6b0c6a070b84909)
    - [Commit ea43263](https://github.com/yorkie-team/yorkie/pull/947/commits/ea4326330bb513703e6cffb5b2e28bd13fbf5448)
2. **Server Handling for Failed `initialDoc`**
    - [Commit 1ab2040](https://github.com/yorkie-team/yorkie/pull/947/commits/1ab2040eb97111559ebe8117e9806b9bcbbe3aba)
    - [Commit d3e077a](https://github.com/yorkie-team/yorkie/pull/947/commits/d3e077a07c8730b43c0bcd4573c2f3526c2cb682)
    - 2.1. Handling Failure `Change`
    - 2.2. Handling Client Rollback with Snapshot

Therefore, I suggest applying only the first part of this issue (#669) in this PR and addressing the second part after resolving issue [#805](https://github.com/yorkie-team/yorkie/issues/805). It is better to consider this approach when dealing with this issue.

window9u:
Closing this PR in favor of a new approach.

This work will be continued in [PR #986](https://github.com/yorkie-team/yorkie/pull/986), which implements a more refined strategy for applying the InitialRoot option during Document attachment.

PR #951, opened 2024-08-04
Handle panic during conversion to connectCode

### LINKED ISSUE #944: Server panics due to convert unhashable struct to connnect.Error
**What happened**:
During the process of handling errors by converting them to `connect.Error`, a panic occurs when a non-hashable error, such as `mongo.CommandError`, is passed.

The issue arises at the following line:

https://github.com/yorkie-team/yorkie/blob/b1f9e0053ab2af8e4f0f9b8b699f80cc1e821ff0/server/rpc/connecthelper/status.go#L154

The server should not panic and handle non-hashable errors gracefully.

**How to reproduce it**:

This issue can be reproduced in the same way as yorkie-team/yorkie#943

1. Pass a non-hashable error, such as `mongo.CommandError`, in the error handling process.
2. Observe the panic occurring when trying to handle the non-hashable error.

**Anything else we need to know**:

For example, `mongo.CommandError` struct is defined like below, and type `bson.Raw(alias of []byte)` and `[]string` is unhashable. So `mongo.CommandError` is unhashable struct.

```go
type CommandError struct {
	Code    int32
	Message string
	Labels  []string // Categories to which the error belongs
	Name    string   // A human-readable name corresponding to the error code
	Wrapped error    // The underlying error, if one exists.
	Raw     bson.Raw // The original server response containing the error.
}
```

**Environment**:
- Operating system: 
- Browser and version: 
- Yorkie version: v0.4.28
- Yorkie JS SDK version:

blurfx:

If an unhashable error is given, the error cannot be converted to a connecteError with a map, causing a panic and shutting down the server.

Fixes #944 

Would it be a good idea to log errors in recover?

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling in the connection status function to enhance resilience against runtime errors.
	- Implemented panic recovery to prevent unexpected failures during execution.
- **Tests**
	- Introduced a test suite for the connection helper functions, ensuring robust error handling through various scenarios.

### DIFF
--- a/server/rpc/connecthelper/status.go
+++ b/server/rpc/connecthelper/status.go
@@ -151,9 +151,10 @@ func errorToConnectError(err error) (*connect.Error, bool) {
 		cause = errors.Unwrap(cause)
 	}
 
+	// NOTE(hackerwins): This prevents panic when the cause is an unhashable
+	// error.
 	var connectCode connect.Code
 	var ok bool
-
 	defer func() {
 		if r := recover(); r != nil {
 			ok = false
@@ -161,7 +162,6 @@ func errorToConnectError(err error) (*connect.Error, bool) {
 	}()
 
 	connectCode, ok = errorToConnectCode[cause]
-
 	if !ok {
 		return nil, false
 	}
--- a/server/rpc/connecthelper/status.go
+++ b/server/rpc/connecthelper/status.go
@@ -151,7 +151,17 @@ func errorToConnectError(err error) (*connect.Error, bool) {
 		cause = errors.Unwrap(cause)
 	}
 
-	connectCode, ok := errorToConnectCode[cause]
+	var connectCode connect.Code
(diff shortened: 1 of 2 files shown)


blurfx:
@krapie No, it doesn't. Should we print or log errors in this case?

krapie:
@blurfx I think printing out error details may help us to take actions upon that specific error like `mongo.CommandError`.

hackerwins:
@krapie @blurfx It seems that an error will propagate to the callers and be output by the interceptor. 🤔

https://github.com/yorkie-team/yorkie/blob/54910c172e5d648235777c49a532ccc8eefdf222/server/rpc/interceptors/default.go#L52-L55

PR #957, opened 2024-08-09
Add MaxHeightSplay for the splay tree to improve its performance by reducing its skewness.

### LINKED ISSUE #941: Improve performance of Splay Tree in crdt.Text data structure to prevent skewness

Currently, Splay trees are used in the `crdt.Text` data structure. While Splay trees are efficient when performing operations sequentially in a contiguous range, they suffer from the drawback that sequentially inserted elements are linearly arranged on the tree, leading to it becoming a skew tree.
In a tree with N vertices, performing M operations guarantees performance in O((N+M) log N) time. However, if the tree becomes skewed, each operation may take O(N) time initially before returning to faster performance.
This issue may lead to performance degradation in the `crdt.Text` data structure due to the skewness of the Splay tree.
It is essential to explore methods to prevent skewness that could affect the performance and efficiency of the Text data structure.

**Why is this needed**:
To maintain and enhance the performance and efficiency of the `crdt.Text` data structure in the face of potential skewness issues in the Splay trees.

m4ushold:

Why this is needed:
Currently, the crdt.Text data structure uses a Splay tree. While Splay trees are efficient for performing operations in a continuous range, they have a downside where consecutively inserted elements may become linearly arranged in the tree, leading to a skew tree. When performing M operations on a tree with N nodes, the performance is generally guaranteed to be O((N+M) log N). However, if the tree becomes skewed, each operation might initially take O(N) time, though it could eventually improve. This skewness can degrade the performance of the crdt.Text data structure, making it crucial to explore ways to prevent this skewness and maintain the tree's performance and efficiency. This PR aims to address the performance and efficiency degradation in the crdt.Text data structure caused by skewness in the Splay tree.

What this PR does:
In this Pull Request, we introduce a method called max_height_splay to reduce the skewness. This method involves finding the deepest leaf node and performing a splay operation on it every √n operations, where n is the number of splay operations performed. The time complexity of max_height_splay is also amortized O(log n), similar to other operations. A POC for this approach was implemented in C++, and the code can be found [here](https://gist.github.com/m4ushold/eb796aa394cf919fc8bd8fd2d37cf124).

Fixes #941 

If operations are primarily occurring at the end of the document, performance degradation might still occur. Although this may require further research, it seems worthwhile to implement as it can provide good performance in most cases. 
I have added test code but did not include benchmark code. If you think it's needed, please mention me! And while adding the test code, I needed to check the height of the Splay tree nodes, which led to some modifications in other test code as well.

```release-note
NONE
```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

## Summary by CodeRabbit

- **New Features**
  - Enhanced Splay tree functionality with improved balancing mechanisms.
  - Added height management for nodes to optimize performance on tree operations.

- **Bug Fixes**
  - Updated expected output formats in tests to reflect accurate representations of arrays and tree structures.

- **Tests**
  - Expanded test cases to validate splay tree behavior with new height management features and updated node representations.
  - Improved fidelity of benchmark tests for the Document structure’s array operations.


m4ushold:
I’ve decided to close this PR as further research is required. The code applying max height splay caused a stack overflow during benchmarking, and I need to analyze the cause in more detail. 

In addition, I found a test case that works inefficiently because the new implementation method calls the max height display function first and then the display.

Given these issues, I see a strong need for benchmarking code that can compare the new implementation with the existing one. I will prioritize this task first, and once all concerns are fully resolved, I will submit a new PR.

PR #958, opened 2024-08-10
Modify health check endpoint and add HEAD method

### LINKED ISSUE #832: Add Plain HTTP Health Check Endpoint

Currently, our system only has the ConnectRPC health check endpoint `grpc.health.v1.Health/Check`, which requires an HTTP POST request with an application/json header and an empty body. This type of health checking is not convenient for common uptime checkers, especially free 3rd-party uptime checkers.

To address this limitation, we should implement a simple Plain HTTP health check endpoint that can be accessed with an HTTP HEAD request for easier and more versatile health checking.

**Why is this needed**:

The current health check method is not user-friendly for common uptime checkers. Introducing a Plain HTTP health check endpoint will make it easier for users to monitor the health status of our system.

taeng0204:
In PR #952 , There were issues with existing health checks needing to support HEAD methods for UptimeRobot and overlapping with envoy's health check path.
So I added the HEAD method and test code, and modified the health check endpoint.

Fixes #832 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

## Summary by CodeRabbit

- **New Features**
	- Introduced a new HTTP health check handler, allowing services to respond to health check requests via GET and HEAD methods.
	- Enhanced server functionality with the registration of an additional health check endpoint.
	
- **Bug Fixes**
	- Improved clarity and structure of the health check response handling.

- **Tests**
	- Expanded health check tests to include both gRPC and HTTP checks for multiple services, improving coverage and robustness.


krapie:
@taeng0204 Could you please resolve the conflicts?

server/rpc/httphealth/httphealth.go
```diff
@@ -24,46 +24,41 @@ import (
...
-const serviceName = "/healthz/"
-
 // CheckResponse represents the response structure for health checks.
 type CheckResponse struct {
 	Status string `json:"status"`
 }
 
 // NewHandler creates a new HTTP handler for health checks.
 func NewHandler(checker grpchealth.Checker) (string, http.Handler) {
+	const serviceName = "/yorkie.v1/healthz/"
```
krapie, line 34: How about changing the service name to `/yorkie.v1.Health/Check` to correspond with `/grpc.health.v1.Health/Check` and `/yorkie.v1.YorkieService/XXX` to keep consistency with other RPCs?

Also, I think we don't need the ending `/` slash, since we are using `?` query parameter after.

krapie (changes requested):

I think we need to change the service endpoint for the readability and consistency with the other RPCs.

taeng0204:
@krapie If it's a consistency concern, can I just unify the "grpchealth" library and endpoints?

krapie:
@taeng0204 You mean `/grpc.health.v1.Health/Check`? If that is applicable, let's combine them.

krapie:
@taeng0204 I think we should include that the health check is within Yorkie service to state that this health check is related to Yorkie, so `/yorkie.v1.YorkieService/health` or other names that includes `/yorkie.v1.YorkieService` will be better.

taeng0204:
@krapie In the existing grpc health check, there is a `grpc.health.v1.Health` that supports health checks for the yorkie service, admin service, and health check service, so why not `http.health.v1.Health` instead?

krapie:
@taeng0204 That is also a good approach. When we proceeding with `http.health.v1.Health`, we need reconfigure [yorkie cluster's routing conditions](https://github.com/yorkie-team/yorkie/blob/80c6ea0f493aedfd7f7d6512ee52de86f8265add/build/charts/yorkie-cluster/templates/istio/virtual-service.yaml#L16).

Note that we are using `/yorkie.v1` for yorkie service route, `/grafana` for Grafana dashbaord, and `/argocd` for ArgoCD dashboard.

taeng0204:
@krapie I tried modifying it like `grpchealth`'s code, but I'm not sure if it's okay to use the variable names and comments as is.
```go
// HealthV1ServiceName is the fully-qualified name of the v1 version of the health service.
const HealthV1ServiceName = "/yorkie.v1.YorkieService/health"
```

server/rpc/httphealth/httphealth.go
```diff
@@ -34,36 +34,34 @@ type CheckResponse struct {
 
 // NewHandler creates a new HTTP handler for health checks.
 func NewHandler(checker grpchealth.Checker) (string, http.Handler) {
+	const serviceName = "/yorkie.v1.YorkieService/health"
```
krapie, line 37: I think now we have introduced `HealthV1ServiceName` constant, we can remove this variable.

krapie (changes requested):
> I'm not sure if it's okay to use the variable names and comments as is.

It is fully okay 👍🏼 
We now only have one last change before the merge :)

server/rpc/httphealth/httphealth.go
taeng0204: It preserves the structure of the existing grpchealth, which we don't need! Thanks:)

krapie (approved):
LGTM 👍🏼

PR #963, opened 2024-08-13
Add metric collection for the number of goroutines attached in the Background task

### LINKED ISSUE #386: Collect Worker Count Metric that Runs Through the Background Package
https://github.com/yorkie-team/yorkie/blob/d0165307ad35dd0a11655841b3628c6b6b94dba4/server/backend/background/background.go#L45

I think you can try to collect the count of workers running in the background package.

**Why is this needed**:
You can know ahead of time when to process many workers. This can prevent problems that may arise from this.

kokodak:

This PR adds a metric to collect the number of goroutines attached to a specific Background task.

Prometheus provides its own thread-safe Gauge metric, which can be used to accurately collect the number of concurrent goroutines.

Also, although the only task running in the background so far is PushPull, I implemented the `AttachGoroutine()` method to take a string called `taskType` as a parameter, since it seems to be focused on reusability.

In my local Grafana, I observed the metrics with a simple promQL query like the one below.

image</br>

The expected effect is that it will be easy to identify which background tasks are bottlenecking and when. For example, during the logic of saving snapshots, the number of observed goroutines may temporarily increase when the DB bottleneck causes each goroutine to have a longer lifecycle. This kind of analysis would help us prevent such issues.

Fixes #386 

If you think there's a better metric or promQL to get more meaningful data, please feel free to comment.

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

## Summary by CodeRabbit

- **New Features**
	- Enhanced background goroutine management with improved metrics tracking.
	- Added functionality to monitor and categorize background tasks through Prometheus.

- **Improvements**
	- Updated function signatures to incorporate metrics tracking for `AttachGoroutine`, `PushPull`, and other related functions.
	- Introduced better observability for background processing tasks, aiding in performance tracking.

These changes significantly enhance the ability to manage and monitor background tasks within the application.

### DIFF
--- a/server/backend/background/background.go
+++ b/server/backend/background/background.go
@@ -25,6 +25,7 @@ import (
 	"sync/atomic"
 
 	"github.com/yorkie-team/yorkie/server/logging"
+	"github.com/yorkie-team/yorkie/server/profiling/prometheus"
 )
 
 type routineID int32
@@ -60,7 +61,11 @@ func New() *Background {
 
 // AttachGoroutine creates a goroutine on a given function and tracks it using
 // the background's WaitGroup.
-func (b *Background) AttachGoroutine(f func(ctx context.Context)) {
+func (b *Background) AttachGoroutine(
+	f func(ctx context.Context),
+	metrics *prometheus.Metrics,
+	taskType string,
+) {
 	b.wgMu.RLock() // this blocks with ongoing close(b.closing)
 	defer b.wgMu.RUnlock()
 	select {
@@ -73,8 +78,12 @@ func (b *Background) AttachGoroutine(f func(ctx context.Context)) {
 	// now safe to add since WaitGroup wait has not started yet
 	b.wg.Add(1)
 	routineLogger := logging.New(b.routineID.next())
+	metrics.AddBackgroundGoroutines(taskType)
 	go func() {
-		defer b.wg.Done()
+		defer func() {
+			b.wg.Done()
+			metrics.RemoveBackgroundGoroutines(taskType)
+		}()
 		f(logging.With(context.Background(), routineLogger))
 	}()
 }
--- a/server/backend/background/background.go
+++ b/server/backend/background/background.go
@@ -50,12 +50,16 @@ type Background struct {
 
 	// routineID is used to generate routine ID.
 	routineID routineID
+
+	// metrics is used to collect metrics with prometheus.
+	metrics *prometheus.Metrics
 }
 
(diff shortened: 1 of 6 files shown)


server/backend/background/background.go
```diff
@@ -60,7 +61,11 @@ func New() *Background {
 
 // AttachGoroutine creates a goroutine on a given function and tracks it using
 // the background's WaitGroup.
-func (b *Background) AttachGoroutine(f func(ctx context.Context)) {
+func (b *Background) AttachGoroutine(
+	f func(ctx context.Context),
+	metrics *prometheus.Metrics,
```
krapie, line 66: I think `metrics` parameter will not change every time we call `AttachGoroutine()`. How about putting `metrics` parameter to `Background` struct?

As you can see in the other use cases of metrics, we pull out `metrics` from struct, just like `Backend`.

krapie (changes requested):
I have left some comments below :)

PS. After we merge this PR, let's add new panel in [Yorkie Grafana Dashbaord](https://grafana.com/grafana/dashboards/18560-yorkie-dashboard/) using the PromQL you mentioned in your description.

server/backend/background/background.go
kokodak: Great idea!
The arguments that go into the parameters are easy things that change, but metrics probably won't. I will be making the changes you suggested.

krapie (approved):
LGTM :)

hackerwins (approved):
Thanks for your contribution.

krapie:
@kokodak We have deployed this feature with Grafana panel added.
But it seems like the metrics is not collected. Can you verify it again?

image

kokodak:
@krapie 
There is no problem with the metric collection itself.
However, in most scenarios, the lifecycle of the goroutines running in the background is extremely short, and the interval between collecting metrics is long, so it doesn't seem to be observed.

This was happening to me locally, but when I forcibly increased the lifecycle of the goroutine, I was able to observe it normally.
On the yorkie main server, I expected to see more traffic than on my local, so I expected to see some observations, but it's not working.

What about changing the metric from Gauge to Counter, and changing it to observe the gradient of the number of goroutines increasing in time?

krapie:
@kokodak Good point! Looks like the lifespan of the goroutine is too short to collect. 
Well, I can see some changes when I change the query to rate_interval.

image

cc. @hackerwins

Issue #943, opened 2024-07-25
Error occurs when executing `FindDocInfosByKeys` query with empty docKeys argument


raararaara:
**What happened**:
When performing the `FindDocInfosByKeys` query with an empty array for docKeys, the server encounters a critical issue. Specifically, the server crashes with the following error:
```
2024-07-26T01:27:13.710+0900    WARN    a16     RPC : "/yorkie.v1.AdminService/GetDocuments" 
1.773583ms: &{"project_name:\"default\"" {'\x00' "/yorkie.v1.AdminService/GetDocuments" 
%!q(bool=false) '\x00'} {"[::1]:49484" "connect" map[]} map["Authorization":
["cqh6cnh97i624rnkavk0"] "Content-Length":["70"] "Content-Type":["application/json"] "Host":
["localhost:8080"]] "POST"} => "find documents: (BadValue) $in needs an array"
```

**How to reproduce it (as minimally and precisely as possible)**:
1. Execute the `FindDocInfosByKeys` query with an empty array for docKeys.
2. Observe the server crashing and generating the mentioned error.

**Anything else we need to know?**:

**Environment**:
- Operating system:
- Browser and version:
- Yorkie version (use `yorkie version`): v0.4.28
- Yorkie JS SDK version:

raararaara:
cc @blurfx

blurfx:
Unhashable errors are not going to have a common interface, so it's probably better to handle panic in that context with `recover`.

Considering just `mongo` for example, `mongo.CommandError` and `mongo.WriteError`, `mongo.WriteException`, and `mongo.BulkWriteException` do not have a common field, so interface checking is not possible.

PR #972, opened 2024-08-20
Fix invalid test case in FindDocInfosByKeys

kokodak:

The current test case, the intent of the test(there is a length validation so that no errors are thrown) and the actual test are different.

In practice, current FindDocInfosByKeys test passes without length validation on the key list.

This is because `var keys []key.Key` is a **nil slice**, and `[]key.Key{}` is a **non-nil slice**.

https://github.com/yorkie-team/yorkie/blob/eb0c2d32070b750997dc77b051e1ab18324e9fb2/server/rpc/admin_server.go#L271

https://github.com/yorkie-team/yorkie/blob/eb0c2d32070b750997dc77b051e1ab18324e9fb2/server/backend/database/testcases/testcases.go#L146

The difference between the two is important, so in this PR, it modifies the test code to match the intent of the original test.

If a nil value is inserted into a DB query, it can lead to unintended results, so it would be nice to be aware of this.

Related to https://github.com/yorkie-team/yorkie/issues/943

Thanks for your review!

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

## Summary by CodeRabbit

- **Chores**
	- Enhanced code readability by adjusting the representation of an empty slice of keys in the test case to a more idiomatic format.

### DIFF
--- a/server/backend/database/testcases/testcases.go
+++ b/server/backend/database/testcases/testcases.go
@@ -143,7 +143,8 @@ func RunFindDocInfosByKeysTest(
 		}
 
 		// 02. Find documents
-		infos, err := db.FindDocInfosByKeys(ctx, projectID, []key.Key{})
+		var emptyKeys []key.Key
+		infos, err := db.FindDocInfosByKeys(ctx, projectID, emptyKeys)
 		assert.NoError(t, err)
 		assert.Len(t, infos, 0)
 	})
--- a/server/backend/database/testcases/testcases.go
+++ b/server/backend/database/testcases/testcases.go
@@ -143,8 +143,7 @@ func RunFindDocInfosByKeysTest(
 		}
 
 		// 02. Find documents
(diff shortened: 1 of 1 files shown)


server/backend/database/testcases/testcases.go
```diff
@@ -143,7 +143,8 @@ func RunFindDocInfosByKeysTest(
 		}
 
 		// 02. Find documents
-		infos, err := db.FindDocInfosByKeys(ctx, projectID, []key.Key{})
+		var emptyKeys []key.Key
+		infos, err := db.FindDocInfosByKeys(ctx, projectID, emptyKeys)
```
hackerwins, line 147:
```suggestion
		infos, err := db.FindDocInfosByKeys(ctx, projectID, nil)
```

hackerwins (approved):
Thanks for your contribution.

PR #974, opened 2024-08-21
Optimize FindChangeInfosBetweenServerSeqs to prevent unnecessary Query

kokodak:

This PR modifies the `FindChangeInfosBetweenServerSeqs` method to avoid executing DB queries when the `from` > `to`. This change minimizes unnecessary database resource consumption, particularly in scenarios where the most recent document editor, User A, continues editing without any interference from other users (B, C, etc.).

In such cases, in PushPull, User A's `Checkpoint.serverSeq` always equal to the server's `initialServerSeq` (DocInfo.serverSeq), leading to a situation where `from > to` in the `FindChangeInfosBetweenServerSeqs` method. By preventing the execution of a query under these circumstances, we can reduce the load on database resources.

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling in the database query methods to prevent execution with invalid input parameters, enhancing overall stability and robustness.

- **Tests**
	- Added new test cases to validate the behavior of the database operations, ensuring consistent server sequence values during user editing without interference and verifying correct retrieval of document snapshots.

### DIFF
--- a/server/backend/database/memory/database.go
+++ b/server/backend/database/memory/database.go
@@ -1051,6 +1051,9 @@ func (d *DB) FindChangeInfosBetweenServerSeqs(
 	txn := d.db.Txn(false)
 	defer txn.Abort()
 
+	if from > to {
+		return nil, nil
+	}
 	var infos []*database.ChangeInfo
 
 	iterator, err := txn.LowerBound(
--- a/server/backend/database/mongo/client.go
+++ b/server/backend/database/mongo/client.go
@@ -1030,6 +1030,9 @@ func (c *Client) FindChangeInfosBetweenServerSeqs(
 	from int64,
 	to int64,
 ) ([]*database.ChangeInfo, error) {
+	if from > to {
+		return nil, nil
+	}
 	cursor, err := c.collection(ColChanges).Find(ctx, bson.M{
 		"project_id": docRefKey.ProjectID,
 		"doc_id":     docRefKey.DocID,
--- a/server/backend/database/testcases/testcases.go
+++ b/server/backend/database/testcases/testcases.go
@@ -359,7 +359,7 @@ func RunFindChangeInfosBetweenServerSeqsTest(
 		assert.NoError(t, clientInfo.AttachDocument(docInfo.ID, false))
 		assert.NoError(t, db.UpdateClientInfoAfterPushPull(ctx, clientInfo, docInfo))
 
-		updatedClientInfo, err := db.FindClientInfoByRefKey(ctx, clientInfo.RefKey())
+		updatedClientInfo, _ := db.FindClientInfoByRefKey(ctx, clientInfo.RefKey())
 
 		// Record the serverSeq value at the time the PushPull request came in.
 		initialServerSeq := docInfo.ServerSeq
--- a/server/backend/database/testcases/testcases.go
+++ b/server/backend/database/testcases/testcases.go
@@ -378,6 +378,148 @@ func RunFindChangeInfosBetweenServerSeqsTest(
 		assert.NoError(t, err)
 		assert.Len(t, changeInfos, 0)
 	})
+
+	t.Run("retrieving a document with snapshot that reflect the latest doc info test", func(t *testing.T) {
+		ctx := context.Background()
+
+		docKey := key.Key(fmt.Sprintf("tests$%s", t.Name()))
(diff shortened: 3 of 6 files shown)


hackerwins (commented):
Thanks for your contribution.

Can you leave the problem case described in the issue as a test case?

https://github.com/yorkie-team/yorkie/blob/main/server/backend/database/testcases/testcases.go

kokodak:
I've written test code to cover what was mentioned in the PR text. But I'm worried that the test case for the database has too business-like terms in it.

Also, I'm not sure how to express in the test that the `from > to` condition caused an early return. How do you think it should be expressed?

If you have any suggestions, please feel free to let me know.

server/backend/database/memory/database.go
```diff
@@ -1051,6 +1051,9 @@ func (d *DB) FindChangeInfosBetweenServerSeqs(
 	txn := d.db.Txn(false)
 	defer txn.Abort()
```
sejongk, line 1053: How about simply explaining the case when this early return is necessary as a comment?

server/backend/database/mongo/client.go
```diff
@@ -1030,6 +1030,9 @@ func (c *Client) FindChangeInfosBetweenServerSeqs(
 	from int64,
 	to int64,
 ) ([]*database.ChangeInfo, error) {
+	if from > to {
```
sejongk, line 1033: Here too.

server/backend/database/testcases/testcases.go
```diff
@@ -343,6 +343,43 @@ func RunFindChangesBetweenServerSeqsTest(
 	})
 }
 
+// RunFindChangeInfosBetweenServerSeqsTest runs the FindChangeInfosBetweenServerSeqs test for the given db.
+func RunFindChangeInfosBetweenServerSeqsTest(
+	t *testing.T,
+	db database.Database,
+	projectID types.ID,
+) {
+	t.Run("continues editing without any interference from other users test", func(t *testing.T) {
```
sejongk, line 352: I think it would be clearer to focus on testing whether the function works well in both cases (`from > to` and `from <= to`). This approach effectively defines the function's spec as test cases, allowing us to verify that the function performs correctly regardless of any internal optimizations, even if more optimizations are applied later.

server/backend/database/memory/database.go
kokodak: That comment could be useful.
On the other hand, I wonder if developers need to know when the `from > to` condition occurs. Because, since the DB query has the information that `from <= server_seq <= to`, we can intuitively realize that the case `from > to` doesn't exist.
Think it's still better to give more information? @sejongk

server/backend/database/mongo/client.go
kokodak: Same as https://github.com/yorkie-team/yorkie/pull/974#discussion_r1729318827.

server/backend/database/memory/database.go
kokodak: Actually, retrieving a document can also cause a `from > to` condition.
When a snapshot of a document reflects the most recent state of the document, there are no additional changes to look for.
If we're going to comment out a short description, it would be nice to specify both cases.

sejongk (approved):
LGTM 👍.

sejongk:
@hackerwins Would you mind reviewing all the changes?

hackerwins (commented):
It looks good overall. I am curious about the following:

> This change minimizes unnecessary database resource consumption, particularly in scenarios where the most recent document editor, User A, continues editing without any interference from other users (B, C, etc.).

Could you explain me how to reproduce this scenario? I want to ensure there isn't another underlying issue.

hackerwins (approved):

kokodak:
> Could you explain me how to reproduce this scenario? I want to ensure there isn't another underlying issue.

Reproducing at the user level is very simple: you simply continue editing the document by yourself, without concurrent editing by other users.

Here is an example scenario:

A. Open the CodeMirror example of the JS SDK in a single browser
B. Type 'abc' in the browser
C. Check the server for queries from the corresponding editing client

In the above scenario, unnecessary queries were still occurring.

PR #977, opened 2024-08-25
Set `updated_at` with `created_at` when creating Document

### LINKED ISSUE #968: Set updatedAt value for newly created documents
**What happened**:

When creating a new document through the dashboard example, the `updatedAt` value is not displaying correctly in the document list. (Ref: https://github.com/yorkie-team/dashboard/issues/99)

When a document is first created, the `updatedAt` should be set to the same value as `createdAt`.
Reasons for this approach:
1. Initial state indication: It clearly shows that the record has not been modified since its creation.
2. Data consistency: All records will have an `updatedAt` value from the moment of creation.
 
1. Navigate to the dashboard
2. Create a new document using the provided example
image
3. Check the document list
image
4. Observe that the `updatedAt` value is not correctly set

**Environment**:
- Operating system:
- Browser and version:
- Yorkie version (use `yorkie version`): 0.4.30
- Yorkie JS SDK version: 0.4.28

window9u:

This PR ensures that when a new `Document` is created and attached, the `updated_at` field is initialized with the `created_at` timestamp. 

Before the commit [b494fa2d](https://github.com/yorkie-team/yorkie/commit/b494fa2d02c17e229ad3fc5cebaa7ea33231077a), the `updated_at` field was correctly set during the attachment process through a `pushpull` operation that included presence. However, after this commit, the `updated_at` field is no longer updated when the document is attached, leading to inconsistencies.

Fixes #968

The underlying issue stems from the `FindDocInfoByKeyAndOwner` function, which returned a `DocInfo` object without setting the "owner", "server_seq", "created_at", and "updated_at" fields, leading to zero values being assigned. While this functionality is acceptable within the specific context where only `project_id` and `id` are required, it creates issues when the `AttachDocument` function relies on these values. It currently works around this issue since `server_seq` has a zero value of 0. 

However, I believe we should return `DocInfo` with the updated "owner", "server_seq", "created_at", "updated_at" fields to maintain consistency and improve clarity.

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

## Summary by CodeRabbit

- **New Features**
	- Introduced an `UpdatedAt` field for improved tracking of document update times.
	- Enhanced document update functionality to return the latest version post-update.

- **Bug Fixes**
	- Improved test assertions to ensure document ownership and timestamp accuracy.

- **Documentation**
	- Updated comments for improved readability in test cases.

### DIFF
--- a/server/backend/database/memory/database.go
+++ b/server/backend/database/memory/database.go
@@ -746,6 +746,7 @@ func (d *DB) FindDocInfoByKeyAndOwner(
 			Owner:      clientRefKey.ClientID,
 			ServerSeq:  0,
 			CreatedAt:  now,
+			UpdatedAt:  now,
 			AccessedAt: now,
 		}
 		if err := txn.Insert(tblDocuments, info); err != nil {
--- a/server/backend/database/mongo/client.go
+++ b/server/backend/database/mongo/client.go
@@ -741,7 +741,7 @@ func (c *Client) FindDocInfoByKeyAndOwner(
 				"created_at": now,
 				"updated_at": now,
 			},
-		})
+		}, options.FindOneAndUpdate().SetReturnDocument(options.After))
 	} else {
 		result = c.collection(ColDocuments).FindOne(ctx, filter)
 		if result.Err() == mongo.ErrNoDocuments {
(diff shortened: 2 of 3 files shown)


hackerwins (approved):
Thanks for your contribution.

Issue #723, opened 2023-12-13
Incorrect Garbage Collection Due to MinSyncedSeq Determination with Lamport Clocks


chacha912:
**What happened**:

## 1. Inaccurate Garbage Collection using Lamport Clocks

In Yorkie, [the Garbage Collection (GC) feature](https://github.com/yorkie-team/yorkie/blob/main/design/garbage-collection.md) is designed to eliminate tombstone nodes once every peer has acknowledged the deletion operation (meaning these nodes are no longer referenced by remote peers). The determination of whether all peers have received a deletion operation is based on the time of the min synced sequence change. This change (minSyncedSeq) is determined using the smallest Lamport clocks. However, while Lamport clocks enable total ordering, they lack the capability to distinguish between concurrency and causality of events, leading to inaccuracies in GC decisions.

Consider the scenario where concurrent edits occur:

![GC(as-is)](https://github.com/yorkie-team/yorkie/assets/81357083/1fe9c87b-75e2-40fd-80e0-023ec90ea17d)
<img width="300" alt="" src="https://github.com/yorkie-team/yorkie/assets/81357083/92d9c4f2-1929-478f-b90c-e3a142b7c4ed">

1. Initially, the document starts with the string `abd`, with clients A and B both synced up to the change `3@B`.
2. Simultaneously, A and B perform edits. A deletes `bd`, and B adds `c` between `b` and `d`.
3. B syncs first, marking the server with `4@B` as synced.
4. A syncs next, pulling `4@B` and pushing `4@A`. The server registers `4@A` as synced.
5. A applies the edit for `4@B`, triggering GC since min synced seq is `4@A`, and the `removedAt` matches `4@A`.
6. B inserts `1` after `b`.
7. B syncs again, pushing `5@B` and pulling `4@A`.
8. A syncs, pulling `5@B`, but encounters an error since it can't find `2@A`.

The bug stems from B not yet syncing `4@A` in step 3 but marking it as synced since we use Lamport clock to determine the min synced sequence. Although `4@A < 4@B`, it doesn't guarantee the sync of `4@A` due to Lamport clock limitations.
- Lamport Clocks 
    - `a->b` => `L(a)<L(b)`
    - but, `L(a)<L(b)` => `a->b` or `a||b`(concurrency)

### 2-1. Use of Vector Clocks

Replace Lamport clocks with vector clocks to manage the concurrency of changes. Vector clocks maintain an array of counters for each client, representing the ordering of changes.

[...]

PR #981, opened 2024-08-27
Implementing GC Using Version Vector

JOOHOJANG:
The purpose of this PR is to address a defect found in the existing GC that used syncedSeq, by introducing VersionVector to resolve the issue.

The diagram of the structure can be found at this [link](https://miro.com/app/board/o9J_kidjCtk=/?moveToWidget=3458764594957276897&cot=14).​

Here are the four key points to focus on:

1. What is Lamport Synced Version Vector and how to use it
2. How to store version vector into database and update it when receives change
3. What is min version vector and how to solve GC problem
4. How to handle detached client's version vector

### 1. What is Lamport Synced Version Vector
[Definition of Version Vector](https://en.wikipedia.org/wiki/Version_vector)
Since we're using lamport timestamp to handle CRDT, we implement Lamport Synced Version Vector which is quite different from original Version Vector.
Lamport Synced Version Vector allows us to continue using the existing method that relied on Lamport.
It is key:value map consist of actorID and lamport.

image
As you can see from the image above, which is typical version vector example, client's value(lamport in our case) is not the largest at every moment.

image
At the above image, you can see that the client’s Lamport is always the largest, which is consistent with how we used to update and utilize lamport in the past.

For convenience, I will refer to Lamport Synced Version Vector as the version vector from now on.

When creating a new change, we increment both the lamport and the lamport in the version vector to use as the ID. Thus, we don't need to use version vector as ID, but use as a part of ID. You can see how it is used in the modified code (see changeID parts).

When receiving a new change, compute max version vector by using version vector from change and document's version vector.

```
max(vv1, vv2) =>
for (key in vv1 || key in vv2) {
    maxVV[key] = max(vv1[key] || 0, vv2[key] || 0)
}


ex)
max([c1:2, c2:3, c3:4], [c1:3, c2:1, c3:5, c4:3])
=> [c1:3, c2:3, c3:5, c4:3]

```
```
min(vv1, vv2) =>
for (key in vv1 || key in vv2) {
    minVV[key] = min(vv1[key] || 0, vv2[key] || 0)
}

ex)
min([c1:2, c2:3, c3:4], [c1:3, c2:1, c3:5, c4:3])
=> [c1:2, c2:1, c3:4, c4:0]

```

### 2. How to store version vector into database and update it when receives change
1. Attach
image

- set initial version vector ([c1: 0])
- make change which includes updated changeID (version vector [c1: 1])
- push change into server and change table in database.
- push version vector into version vector table in database
    - you can see the detailed logic from modified codes (`updateVersionVector`, `updateAndFindMinSyncedVersionVectorAfterPushPull`)

2. Pushpull
- pushonly
    - there's no received changes when its pushonly, thus no need to update document's version vector.
- pushpull
    - after receives change, we update document's version vector when every change has been applied. 
```
1. update changeID
(except vv, everything is same as before)

doc.vv = max(doc.vv, doc.vv)
doc.vv[doc.actorid] = maxLamport + 1
```

### 3. What is min version vector and how to solve GC problem
Min version vector is the vector which consists of minimum value of every version vector stored in version vector table in database.

Conceptually, min version vector is version vector that is uniformly applied to all users. It ensures that all users have at least this version applied. Which means that it is safe to perform GC.

You can see detailed logic from `UpdateAndFindMinVersionVectorAfterPushPull` in modified codes.
Suppose clientA is performing pushpull

1. if there're version vectors in database, collect them.
2. collect clientA's version vector from change pack
3. store clientA's version vector into database
4. compute min from collected version vectors

GC algorithm
```
if (removedAt.lamport <= minSyncedVersionVector[removedAt.actor]) {
    GC()
}
```
If node.removedAt.lamport is smaller or equal to every value in min version vector, then run GC.
It's okay to run GC when its equal, because the meaning of min version vector is that every client's version vector is at least larger then min version vector.

### 4. How to handle detached client's version vector
If detached client's version vector remains in database and local document, its memory wasting and also GC will not performed.

```
1. minVV = [c1:3, c2:4, c3:4]
2. c3 detach
3. c2 remove nodes k , k > 4
4. c1, c2 keep updating and perform pushpull
5. minVV at some moment [c1: n, c2: m, c3:4] where n, m > k > 4
6. k > 4, can't perform GC
```
image

So it's necessary to remove detached client's version vector and its lamport from other's version vector.

1. When client detaches, remove its version vector from version vector table
2. remove client's lamport from every version vector stored in database.
3. when other client perform pushing and its version vector includes detached client's lamport, filter it by using min version vector.
    - the reason why we're using min version vector in this case is that we already compute min version vector (to return min version vector to perform GC).
    - min version vector includes all attached users in the document.
4. store filtered version vector into database.
5. when other client perform pulling, we also use min version vector to filter it after apply change.
6. after few times of pushpull, detached client's lamport will be totally removed from database and others document.

---

By the way, the code is a bit messy, so any reviews are always welcome. If you have any ideas for improvements, feel free to share them.

---

Address #723

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


pkg/document/document.go
```diff
@@ -219,7 +219,12 @@ func (d *Document) ApplyChangePack(pack *change.Pack) error {
 		d.GarbageCollect(pack.MinSyncedVersionVector)
 	}
 
-	// 05. Update the status.
+	// 05. Remove detached client's lamport from version vector if it exists
+	if pack.MinSyncedVersionVector != nil {
+		d.doc.changeID = d.doc.changeID.SetVersionVector(d.doc.changeID.VersionVector().Filter(pack.MinSyncedVersionVector.Keys()))
+	}
```
hackerwins, line 225: The same logic also exists in `internal_document`. It seems necessary to check why it is implemented repeatedly.

pkg/document/time/version_vector.go
```diff
@@ -38,6 +40,11 @@ func (v VersionVector) Set(id *ActorID, i int64) {
 	v[id.bytes] = i
 }
 
+// UnSet unsets the given actor's version to the given value.
+func (v VersionVector) UnSet(id *ActorID) {
```
hackerwins, line 44:
```suggestion
func (v VersionVector) Unset(id *ActorID) {
```

pkg/document/time/version_vector.go
```diff
@@ -81,6 +96,7 @@ func (v VersionVector) Marshal() string {
 
 // AfterOrEqual returns whether this VersionVector is causally after or equal
 // the given VersionVector.
+// TODO(JOOHOJANG) check below function is necessary
```
hackerwins, line 99: Did we need this function? If we didn't need it, I think we could delete it.

```suggestion
// TODO(JOOHOJANG): check below function is necessary
```

pkg/document/time/version_vector.go
```diff
@@ -97,9 +113,102 @@ func (v VersionVector) AfterOrEqual(other VersionVector) bool {
 	return true
 }
 
+// EqualToOrAfter returns whether this VersionVector's every field is equal or after than given ticket.
+func (v VersionVector) EqualToOrAfter(other *Ticket) bool {
+	return v[other.actorID.bytes] >= other.lamport
+}
+
 // After returns whether this VersionVector is causally after the given ticket.
+// TODO(JOOHOJANG) check below function is necessary
```
hackerwins, line 122: Did we need this function? If we didn't need it, I think we could delete it.

```suggestion
// TODO(JOOHOJANG): check below function is necessary
```

pkg/document/time/version_vector.go
```diff
@@ -97,9 +113,102 @@ func (v VersionVector) AfterOrEqual(other VersionVector) bool {
...
+}
+
+// TODO(JOOHOJANG) remove func
+func (v VersionVector) Keys() []actorID {
+	var actors []actorID
+	for id := range v {
+		actors = append(actors, id)
+	}
+
+	return actors
```
hackerwins, line 213: What is the intention of the comment?

server/backend/database/mongo/client.go
```diff
@@ -1149,62 +1149,170 @@ func (c *Client) UpdateAndFindMinSyncedTicket(
 	), nil
 }
 
-// UpdateAndFindMinSyncedVersionVector updates the given serverSeq of the given client
-// and returns the SyncedVersionVector of the document.
-func (c *Client) UpdateAndFindMinSyncedVersionVector(
+// UpdateAndFindMinSyncedVersionVectorAfterPushPull returns min synced version vector
+func (c *Client) UpdateAndFindMinSyncedVersionVectorAfterPushPull(
```
hackerwins, line 1153: Since we are not using MongoDB Transaction, we need to handle exceptions for partial failures. I will request a meeting about this later.

server/backend/database/version_vector.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// VersionVectorInfo is a structure representing information about the version vector for each document and client.
+type VersionVectorInfo struct {
+	ID            types.ID           `bson:"_id"`
+	ProjectID     types.ID           `bson:"project_id"`
+	DocID         types.ID           `bson:"doc_id"`
+	ClientID      types.ID           `bson:"client_id"`
+	VersionVector time.VersionVector `bson:"version_vector"`
+}
```
hackerwins, line 31: Is there an advantage to storing `VersionVector` per `Client` rather than the min value of VersionVector per `Document`?

test/integration/gc_test.go
```diff
@@ -20,6 +20,7 @@ package integration
 
 import (
 	"context"
+	"github.com/yorkie-team/yorkie/pkg/document/time"
```
hackerwins, line 23: https://github.com/uber-go/guide/blob/master/style.md#import-group-ordering

hackerwins (commented):
Thanks for your contribution. I hope the GC issue will be resolved by this.

I left a few comments.

pkg/document/time/version_vector.go
JOOHOJANG: my bad I should've remove the comment.

server/backend/database/version_vector.go
JOOHOJANG: Storing `VersionVector` its not an option, it's inevitable.

I'll explain this when we have meeting

pkg/document/document.go
JOOHOJANG: oh okay, I'll figure it out and remove this if it's unnecessary

server/backend/database/version_vector.go
hackerwins: @JOOHOJANG wrote and shared the exception situation below.

Screenshot 2024-09-23 at 5 52 14 PM

The problem with storing only the min value of VersionVector on the server is that we can properly calculate the min value for other peer's VV, but when calculating the min value for own previous version vector on the server, we cannot update it to the next version vector.

e.g.) 
- min(A, B, C, ....): ✅
- min(prevA, nextA) => prevA: ❌

hackerwins: https://discord.com/channels/928301813785038878/1263454638254985226/1284163377320427550

server/backend/database/mongo/client.go
JOOHOJANG: I wrote new TC about this case https://github.com/yorkie-team/yorkie/pull/981/commits/ada30e04fd01ccad8cb50b11b90389b85ae9e76a
Assume there're three clients, and one detached but its version vector remains in DB. (partial failure)

The result was same as we expected, no GC runs because min version vector stays the same as when one client detached. Except this, others work properly.

If this case really happened, I think housekeeping should handle this situation

pkg/document/document.go
hackerwins: Let’s see if it can be resolved after merging.

server/backend/database/memory/database.go
```diff
@@ -1177,63 +1178,195 @@ func (d *DB) UpdateAndFindMinSyncedTicket(
...
+		iterator, err := txn.Get(tblVersionVector, "doc_id", docRefKey.DocID.String())
+
+		if err != nil {
+			return err
+		}
+		actorID, err := clientInfo.ID.ToActorID()
+
+		if err != nil {
+			return err
+		}
```
hackerwins, line 1205:
```suggestion
		iterator, err := txn.Get(tblVersionVector, "doc_id", docRefKey.DocID.String())
		if err != nil {
			return err
		}

		actorID, err := clientInfo.ID.ToActorID()
		if err != nil {
			return err
		}
```

server/backend/database/mongo/client.go
```diff
@@ -1149,62 +1149,170 @@ func (c *Client) UpdateAndFindMinSyncedTicket(
...
+	cursor, err := c.collection(ColVersionVector).Find(ctx, bson.M{
 		"project_id": docRefKey.ProjectID,
 		"doc_id":     docRefKey.DocID,
-	}, options.FindOne().SetSort(bson.D{
-		{Key: "server_seq", Value: 1},
-	}))
-	if errors.Is(result.Err(), mongo.ErrNoDocuments) {
-		return nil, nil
+	})
+	defer cursor.Close(ctx)
```
hackerwins, line 1169: Do we need to call this function?

server/backend/database/mongo/client.go
```diff
@@ -1149,62 +1149,170 @@ func (c *Client) UpdateAndFindMinSyncedTicket(
...
-		return nil, fmt.Errorf("decode change: %w", err)
+	if !isAttached {
+		cursor, err := c.collection(ColVersionVector).Find(ctx, bson.M{
+			"project_id": docRefKey.ProjectID,
+			"doc_id":     docRefKey.DocID,
+		})
+		if err != nil {
+			return fmt.Errorf("find all version vectors: %w", err)
+		}
+		defer cursor.Close(ctx)
```
hackerwins, line 1256: Do we need to call this function?

server/backend/database/memory/database.go
```diff
@@ -1177,63 +1178,194 @@ func (d *DB) UpdateAndFindMinSyncedTicket(
...
+
+		minVersionVector = minVersionVector.Min(vvi.VersionVector)
 	}
 
-	var syncedSeqInfo *database.SyncedSeqInfo
-	if raw := iterator.Next(); raw != nil {
-		info := raw.(*database.SyncedSeqInfo)
-		if info.DocID == docRefKey.DocID {
-			syncedSeqInfo = info
+	// 02. Compute min version vector without current client's version vector strored in db
```
hackerwins, line 1313:
```suggestion
	// 02. Compute min version vector without current client's version vector stored in db
```

server/backend/database/memory/indexes.go
```diff
@@ -19,13 +19,14 @@ package memory
...
-	tblSnapshots  = "snapshots"
-	tblSyncedSeqs = "syncedseqs"
+	tblProjects      = "projects"
+	tblUsers         = "users"
+	tblClients       = "clients"
+	tblDocuments     = "documents"
+	tblChanges       = "changes"
+	tblSnapshots     = "snapshots"
+	tblSyncedSeqs    = "syncedseqs"
+	tblVersionVector = "versionvector"
```
hackerwins, line 29: It would be good to fit this into convention.
```suggestion
	tblVersionVectors = "versionvectors"
```

server/backend/database/mongo/client.go
JOOHOJANG: I added it just in case we needed it. Is it okay not to have it?

server/backend/database/mongo/client.go
JOOHOJANG: I added it just in case we needed it. Is it okay not to have it?

server/backend/database/mongo/indexes.go
```diff
@@ -40,6 +40,8 @@ const (
 	ColSnapshots = "snapshots"
 	// ColSyncedSeqs represents the syncedseqs collection in the database.
 	ColSyncedSeqs = "syncedseqs"
+	// ColVersionVector represents the versionvector collection in the database.
+	ColVersionVector = "versionvector"
```
hackerwins, line 44:
```suggestion
	ColVersionVector = "versionvectors"
```

hackerwins (approved):
Thanks for your contribution.

PR #986, opened 2024-08-28
Implement `InitialRoot` option for Document attachment

### LINKED ISSUE #669: Provide an interface to set the initial value when creating a document

window9u:

This PR introduces an `initialRoot` option when users attach a Document. The goal is to set initial values for editing documents with a predefined structure. Rather than checking if the document is in its initial state or empty, this feature checks if the document has been set up with initial values for editing, and if not, it applies them.

**Implementation details**:
- The `attach` function in the client now accepts an `initialRoot` option.
- `initialRoot` can be partially applied. For each element in `initialRoot`:
    - If the key already exists in the Document, that element will be discarded.
    - If the key doesn't exist, the element will be applied.
- In this implementation, `initialRoot` is not sent to the server during `Attach`. It is applied locally to the Document after pushpull during `Attach`.

**How to use the initialRoot option**: 
When attaching a document, you can use the `initialRoot` option. Here's a comparison of the old and new approaches:

Before (without `initialRoot`):
```go
doc := document.New("doc_key")
client1.Attach(ctx, doc)
doc.Update(func(root *json.Object, p *presence.Presence) error {  
    if root.GetCounter("k")!=nil{  
       root.GetCounter("k").Increase(1)  
    }else{  
       root.SetNewCounter("k", crdt.LongCnt, 0).Increase(1)  
    }  
    return nil  
})
```

After (with `initialRoot`):
```go
doc := document.New("doc_key")
client1.Attach(ctx, doc, client.WithInitialRoot(map[string]any{
    "k": json.NewCounter(0, crdt.LongCnt),  
}))
doc.Update(func(root *json.Object, p *presence.Presence) error {  
    root.GetCounter("k").Increase(1)  
    return nil  
})
```

In this case:
- If a counter for "k" already exists, setting it to 0 is discarded.
- If a counter for "k" doesn't exist, it's initialized with 0.
- Users don't need to worry about overwriting existing valid counters.

**Note**: 
- This function doesn't perform type checking to ensure "k" is a counter. Yorkie recommends using the same type for the same key to maintain consistency.
- Currently, when documents are created simultaneously, one user's document may overwrite another's due to LWW (Last Write Wins). To see an example of this case, please refer `concurrent attach with InitialRoot test case`

Addresses #669

1. This implementation uses a key-based criterion to determine whether to apply the `initialRoot`.
2. It works for both new and existing documents, preventing accidental overwriting of content.

```
Users can use an initialRoot option when attaching documents.
```

**Additional documentation**:

**Future plans** (as discussed):

1. Consider creating the Document instance in the `attach` method in future work.
2. Implement offline editing features.

**Checklist**:
* [x] Added relevant tests
* [x] Didn't break existing functionality

<details>
  <summary>Previous discussion</summary>
  <div>
  This PR introduces an "InitialRoot" option when users attach a Document. The goal is to allow initialization under specific conditions. Two approaches are considered:

  1. Apply InitialRoot only when the Document is first created
  2. Apply InitialRoot when the Document is empty

  Here's how each option works:
  1. Apply InitialRoot only when Document is first created:
     * Determined by checking if the Checkpoint's Serverseq is 1.
     * Guarantees InitialRoot is only applied to newly created Documents.
     * Drawback: Won't apply InitialRoot to existing but empty Documents.
  2. Apply InitialRoot when Document is empty:
     * Determined by checking if the Document has no elements and no garbage elements.
     * Allows applying InitialRoot to any empty Document, regardless of creation time.
     * Concern: Potential unintended behavior in concurrent scenarios.

  Current Implementation:
  - Option 2 is currently implemented.
  - Option 1 is being considered as potentially more predictable and safer in concurrent scenarios.

  Implementation Details:
  - Client's Attach function checks Document state after attachment.
  - InitialRoot is applied to Local Document but not immediately pushed.

  Notes for Reviewer:
  1. Consider which option is more appropriate.
  2. Evaluate trade-offs between options:
     - Option 1: More predictable, safer in concurrent scenarios, can't initialize existing empty Documents.
     - Option 2: More flexible, can initialize any empty Document, potentially risky in concurrent scenarios.
  3. If Option 1 is chosen, implementation will be updated.
  </div>
</details>

### DIFF
--- a/client/options.go
+++ b/client/options.go
@@ -102,8 +102,9 @@ func WithPresence(presence innerpresence.Presence) AttachOption {
 	return func(o *AttachOptions) { o.Presence = presence }
 }
 
-// WithInitialRoot sets the initial root of the document.
-// If the document already exists before attachment, this initial root will not be applied.
+// WithInitialRoot sets the initial root of the document. Values in the initial
+// root will be discarded if the key already exists in the document. If some
+// keys are not in the document, they will be added.
 func WithInitialRoot(root map[string]any) AttachOption {
 	return func(o *AttachOptions) {
 		o.InitialRoot = root
--- a/client/options.go
+++ b/client/options.go
@@ -92,13 +92,22 @@ type AttachOption func(*AttachOptions)
 // AttachOptions configures how we set up the document.
 type AttachOptions struct {
 	// Presence is the presence of the client.
-	Presence innerpresence.Presence
-	IsManual bool
+	Presence    innerpresence.Presence
+	InitialRoot map[string]any
+	IsManual    bool
 }
 
 // WithPresence configures the presence of the client.
 func WithPresence(presence innerpresence.Presence) AttachOption {
 	return func(o *AttachOptions) { o.Presence = presence }
+}
+
+// WithInitialRoot sets the initial root of the document.
+// If the document already exists before attachment, this initial root will not be applied.
+func WithInitialRoot(root map[string]any) AttachOption {
+	return func(o *AttachOptions) {
+		o.InitialRoot = root
+	}
 }
 
 // WithManualSync configures the manual sync of the client.
--- a/client/client.go
+++ b/client/client.go
@@ -349,12 +349,14 @@ func (c *Client) Attach(ctx context.Context, doc *document.Document, options ...
 
 	if opts.InitialRoot != nil {
 		if !doc.HasElement() {
-			err = doc.Update(func(root *json.Object, p *presence.Presence) error {
+			if err = doc.Update(func(root *json.Object, p *presence.Presence) error {
 				for k, v := range opts.InitialRoot {
 					root.SetDynamicValue(k, v)
 				}
 				return nil
-			})
+			}); err != nil {
+				return err
+			}
 		} else {
 			return ErrDocumentExists
 		}
--- a/client/client.go
+++ b/client/client.go
@@ -344,16 +344,17 @@ func (c *Client) Attach(ctx context.Context, doc *document.Document, options ...
 		}
 	}
 
-	if opts.InitialRoot != nil {
-		if err = doc.Update(func(root *json.Object, p *presence.Presence) error {
-			for k, v := range opts.InitialRoot {
+	if err = doc.Update(func(root *json.Object, p *presence.Presence) error {
+		for k, v := range opts.InitialRoot {
+			if root.Get(k) == nil {
 				root.SetDynamicValue(k, v)
 			}
-			return nil
-		}); err != nil {
-			return err
 		}
+		return nil
+	}); err != nil {
+		return err
 	}
+
 	return nil
 }
 
--- a/client/client.go
+++ b/client/client.go
@@ -63,6 +63,9 @@ var (
 	// this client.
 	ErrDocumentNotAttached = errors.New("document is not attached")
 
+	// ErrDocumentExists occurs when use initialRoot Option but document already exists.
+	ErrDocumentExists = errors.New("document already exists in server")
+
 	// ErrDocumentNotDetached occurs when the given document is not detached from
 	// this client.
 	ErrDocumentNotDetached = errors.New("document is not detached")
@@ -344,6 +347,18 @@ func (c *Client) Attach(ctx context.Context, doc *document.Document, options ...
 		}
 	}
 
+	if opts.InitialRoot != nil {
+		if !doc.HasElement() {
+			err = doc.Update(func(root *json.Object, p *presence.Presence) error {
+				for k, v := range opts.InitialRoot {
+					root.SetDynamicValue(k, v)
+				}
+				return nil
+			})
+		} else {
+			return ErrDocumentExists
+		}
+	}
 	return nil
 }
 
--- a/client/client.go
+++ b/client/client.go
@@ -63,9 +63,6 @@ var (
 	// this client.
 	ErrDocumentNotAttached = errors.New("document is not attached")
 
-	// ErrDocumentExists occurs when use initialRoot Option but document already exists.
-	ErrDocumentExists = errors.New("document already exists in server")
-
 	// ErrDocumentNotDetached occurs when the given document is not detached from
 	// this client.
 	ErrDocumentNotDetached = errors.New("document is not detached")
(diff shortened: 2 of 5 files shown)


client/options.go
```diff
@@ -92,15 +92,24 @@ type AttachOption func(*AttachOptions)
...
 }
 
 // WithPresence configures the presence of the client.
 func WithPresence(presence innerpresence.Presence) AttachOption {
 	return func(o *AttachOptions) { o.Presence = presence }
 }
 
+// WithInitialRoot sets the initial root of the document.
+// If the document already exists before attachment, this initial root will not be applied.
+func WithInitialRoot(root map[string]any) AttachOption {
```
window9u, line 107: We could provide an option like 'ensureFirstDocument'. If this option is set to true, we can use a condition checking if ServerSeq == 1. However, I'm not certain if this is the best approach.

chacha912 (approved):
Thanks for your contribution! This looks good to me. 
Could you please add information in the PR description about how to use the initialRoot option?

window9u:
@chacha912 

Thank you for your feedback and the summary of our discussions. As you mentioned, the current PR description doesn't accurately reflect the implementation. I've rewritten the entire PR description to align with the current implementation and include information on how to use the initialRoot option.

However, I'm concerned about whether it's the right process to completely overwrite the previous PR description. Could I replace the entire PR description with the updated version below?

### Updated PR Description

This PR introduces an `initialRoot` option when users attach a Document. The goal is to set initial values for editing documents with a predefined structure. Rather than checking if the document is in its initial state or empty, this feature checks if the document has been set up with initial values for editing, and if not, it applies them.

**Implementation details**:
- The `attach` function in the client now accepts an `initialRoot` option.
- `initialRoot` can be partially applied. For each element in `initialRoot`:
    - If the key already exists in the Document, that element will be discarded.
    - If the key doesn't exist, the element will be applied.
- In this implementation, `initialRoot` is not sent to the server during `Attach`. It is applied locally to the Document after pushpull during `Attach`.

**How to use the initialRoot option**: 
When attaching a document, you can use the `initialRoot` option. Here's a comparison of the old and new approaches:

Before (without `initialRoot`):
```go
doc := document.New("doc_key")
client1.Attach(ctx, doc)
doc.Update(func(root *json.Object, p *presence.Presence) error {  
    if root.GetCounter("k")!=nil{  
       root.GetCounter("k").Increase(1)  
    }else{  
       root.SetNewCounter("k", crdt.LongCnt, 0).Increase(1)  
    }  
    return nil  
})
```

After (with `initialRoot`):
```go
doc := document.New("doc_key")
client1.Attach(ctx, doc, client.WithInitialRoot(map[string]any{
    "k": json.NewCounter(0, crdt.LongCnt),  
}))
doc.Update(func(root *json.Object, p *presence.Presence) error {  
    root.GetCounter("k").Increase(1)  
    return nil  
})
```

In this case:
- If a counter for "k" already exists, setting it to 0 is discarded.
- If a counter for "k" doesn't exist, it's initialized with 0.
- Users don't need to worry about overwriting existing valid counters.

**Note**: 
- This function doesn't perform type checking to ensure "k" is a counter. Yorkie recommends using the same type for the same key to maintain consistency.
- Currently, when documents are created simultaneously, one user's document may overwrite another's due to LWW (Last Write Wins). To see an example of this case, please refer `concurrent attach with InitialRoot test case`


**Which issue(s) this PR fixes**: Fixes #669

1. This implementation uses a key-based criterion to determine whether to apply the `initialRoot`.
2. It works for both new and existing documents, preventing accidental overwriting of content.

```
Users can use an initialRoot option when attaching documents.
```

**Additional documentation**:

**Future plans** (as discussed):

1. Consider creating the Document instance in the `attach` method in future work.
2. Implement offline editing features.

**Checklist**:
* [x] Added relevant tests
* [x] Didn't break existing functionality

window9u:
@chacha912 
I apologize for accidentally deleting your comment. This was unintentional and my mistake. Here's what happened:

1. I mistakenly clicked on the checklist in your comment.
2. I thought I could only delete my own interaction with the comment.
3. Intending to remove only my misclicked checklist, I deleted the entire comment.

To rectify this, I've restored your deleted comment below for reference:


--- 
Thanks for your contribution! This looks good to me.
Could you please add information in the PR description about how to use the initialRoot option?

> I've summarized the discussions we had in the Discord thread and left them here for reference.
> 
> ### Future plans
> 
> `Plan 1` (add initialRoot to attach option) → `future work` (create document in attach) → implement `offline editing` feature
> 
> - Reference: [https://codepair.yorkie.dev/hackerwins/66b17903ff67609ce4be81b7/share?token=1xs8do](https://codepair.yorkie.dev/hackerwins/66b17903ff67609ce4be81b7/share?token=1xs8do)
> - [x]  Add initialRoot to attach option
>     - [ ]  Reflect this change in the Yorkie-JS-SDK as well and update the examples for testing.
> - [ ]  Delegate Document instance creation to attach, similar to Liveblocks
> - [ ]  Consider and implement how to create and edit documents in offline mode.
> 
> ### Setting initial values
> 
> - Currently, when documents are created simultaneously, one user's document may overwrite another's due to LWW (Last Write Wins). If this causes issues (when the initial value needs to be set accurately at the time of document creation), we can improve by generating the initial value on the server.
> - Reference: [https://codepair.yorkie.dev/yorkie-2024-summer/66d94468e535e3f72191c4ef/share?token=ykgneg](https://codepair.yorkie.dev/yorkie-2024-summer/66d94468e535e3f72191c4ef/share?token=ykgneg)

hackerwins (approved):
Thanks for your contribution.

Since the implementation is simpler than first approach, it would be good to apply this to JS SDK and see the actual usability.

PR #988, opened 2024-08-31
Reduce CI test time by optimizing task execution

### LINKED ISSUE #789: Reduce CI test time by optimizing task execution
**Description**:

Currently, when a PR is opened, the CI process runs various tasks including `Build and Integration Test`, `Benchmark`, `Sharding Test`, `Releasing Charts`, and `Publishing Docker Images`.

- Build and Integration Test: 8 mins
- Benchmark: 5 mins
- Sharding Test: 3 mins
- Release Charts: 10 secs
- Publish Docker Images: 7 mins

The integration Test takes to complete these tasks is 8 minutes.

To ensure that the testing process finishes within 10 minutes, it is necessary to optimize the tasks. Specifically, we should consider the following changes:

- Rename "Sharding Test" to "Complex Test" to better reflect its purpose.
- Move the "Tree Concurrency Test" from the Integration Test to the Complex Test.

These changes will help reduce the total test time and ensure that the testing process is completed within the required 10 minutes.

**Why**:

Currently, the integration test with the Tree Concurrency Test takes up most of the 8-minute test time, causing delays in the CI process. By optimizing the task execution and reducing the test time, we can improve the efficiency of our CI pipeline and ensure timely feedback on pull requests.

binary-ho:
# 1. **What this PR does / why we need it**:
Reduce CI Time and experiment with various methods

<br>

here is the existing ci time [('be335eb' commit ci run result)](https://github.com/yorkie-team/yorkie/actions/runs/10647332524)
image

<br>

Through experimentation, we could significantly reduce CI time to the level shown below,
but I decided not to adopt this approach.

image

<br>

### I discarded various methods and, in the end, was able to reduce the time by just a little.

image

- `build` : -1m 20s
- `bench` : -1m
- `sharding-test` : +1m

<br> <br>

# 2. **Which issue(s) this PR fixes**:
Fixes #789

# 3. Implements

I tried various methods:

Some were actually adopted, while others were only experimented with and not adopted.

The ones adopted are marked as `Adopt`.

The ones not adopted are marked as `Not Adopt`, with reasons provided. Please review these and consider adopting them where appropriate.

1. `Adopt` **Remove unnecessary 'Get tools dependencies' step from bench, sharding jobs.**
    - Reduced 1m 20s from `bench` and `sharding-test` jobs.
    - **There is no disadvantage.** <br> <br>
2. `Adopt` **Renamed `sharding-test` to `complex-test` and moved `tree_concurrency_test` to `complex-test`. 
It might be worth considering whether this change is beneficial.**
    - Reduced 1m 20s from 2m.
    - `Disadvantage` **:** But `complex-test` execution time increased by 2m.
    - Expanded target packages for `complex-test`. While the original `sharding-test` only required testing changes in `server/backend/database/**`, now changes in `pkg/document/**` and `client/**` packages also trigger tests. <br> <br>
3. `Not Adopt` **Run integration test in parallel.**
    - Reduced 1m 20s from build job.
    - `Disadvantage` **:** There are variables shared for integration tests that have concurrent write operations, causing tests to fail due to the `-race` option. Preventing race conditions now is easy, but it cannot be guaranteed for the future. This requires more consideration. Further details are provided below. <br> <br>
4. `Not Adopt` **Cache `make tools` results.**
    - Reduced 1m 20s from build job.
    - `Disadvantage`: Caching the result of `go install` is easy, but utilizing it is inconvenient, and it is prone to break if the `make tools` process changes. More details are provided below. <br> <br>

For a more convenient review, the detailed content is collapsed. Click to expand and read the details.

<br> <br>

# 3.1 Removing Unnecessary Dependency Step

The dependencies installed by `make tools` are not needed in `bench` and `sharding-test` jobs.

Therefore, I removed the unnecessary 'Get tools dependencies' steps from the bench and sharding jobs.

image

This reduces 1m 20s in bench and sharding-test jobs.

The bench only reduced by 1m 20s. The 3m 20s in the previous CI might look less due to caching of the bench results in the previous CI, where caching time is not displayed.

<br> <br>

# 3.2 Parallel Test in Integration Package

Applying parallel tests to the integration package can reduce 1m 20s from the build job.

**However, I did not apply it.**

<details>

<summary> Here are the reasons why </summary>

1. The current integration tests have shared states for test environment setups, and some tests perform writes to these shared states. Therefore, race conditions occur, and the tests fail due to Yorkie's `-race` test option.

2. It is possible to prevent these race conditions by removing shared writes or using methods like Go's write mutexes that prevent simultaneous writes.

   Eventually, I believe these shared writes should be managed. The purpose of the `-race` option is to validate race conditions in the code being tested, not in the test setup process.

   However, even after removing shared writes, other issues remain and require further thought.
    1. When writing tests in the integration package after introducing parallel tests, contributors must consider these race conditions.
    2. Even if considered, it is challenging to be certain that there are no race conditions. Potential race conditions do not always fail tests. In my experience, the same code only caused build test failures about once every 20-30 times.

    Therefore, it is not possible to know upfront if a test writer avoided race conditions correctly. I am exploring a test approach to prevent this situation, and I plan to share it if I find a solution.

    Once these issues are resolved, we could consider applying parallel tests to Yorkie's longer-running tests, potentially reducing CI or test times significantly. (Benchmarks could also be reduced)

Therefore, I have removed the parallel test introduced earlier.

If test times get longer in the future, adopting it might be worth considering.

</details>

<br> <br>

# 3.3 Reasons for Not Caching `make tools`

All CI jobs initially had a step to call `make tools`. Therefore, I attempted to cache the results of `make tools`, achieving savings of 1m 20s in all jobs.

When the cache hits, the `make-tools` job execution time in the capture below is less than 10 seconds.

image

Check the test results. [after make-tools cache result](https://github.com/binary-ho/yorkie/actions/runs/10652632833)

**However, I decided not to adopt caching for `make tools`.**

<details>

<summary> Here are the reasons  </summary>

- **Advantages:**
    - Saves 1m 20s in the build test when the cache hits.
- **Disadvantages:**
    - The compressed cache takes up about 250MB of space. Also, this cache needs to be managed separately. GitHub Actions deletes caches not accessed for more than 30 days. However, the GitHub Actions Free Plan only provides 500MB.
    You could add steps to manage this separately, but this complicates the CI process.
    - It is difficult to utilize caching.
    If a dependency installation other than `go install` is added to the `make tools` process later, the cache target must be changed.

    Even if `go install` finds the output files, calling it still takes a significant amount of time. From my tests, even if the results of `make tools` are pre-installed in `$GOPATH/bin` and `$GOPATH/pkg`, it still takes about 1 minute, not much different from the original 1m 20s.

    Removing the `make tools` step would allow you to take advantage of caching. Currently, `make tools` only calls `go install` 5 times. However, if dependency installation methods other than `go install` are added later, additional folder caching must be included in the CI.
    - Caching could result in at least one job and numerous steps being added to the YAML file.
    This significantly affects the readability of the YAML file. If caching is used, the file would look like this:
    [ci.yml after make-tools cache](https://github.com/binary-ho/yorkie/actions/runs/10652632833/workflow)

</details>

<br> <br>

# 3.4 Moving `tree_concurrency_test` to `complex-test`

As proposed in the issue, `tree_concurrency_test` was moved to `sharding-test`, 
and some common types and methods were moved to `main_test.go`.

Additionally, `sharding-test` was renamed to `complex-test`.

result:
image

**Advantages:**

1. Reduces build CI time by 1m 20s.
~~(If parallel tests are introduced, CI time will not decrease)~~

**Disadvantages:**

1. After moving, the scope of the target for `complex-test` is broadened, so it runs more frequently.
2. `complex-test` time increases by about 2m.
3. The name `complex-test` feels a bit ambiguous (to me).

<br> 

### As-Is

Currently, `sharding-test` is only triggered by changes in `server/backend/database/**`.

image
<br>

### After Move

After the move, the target package for `complex-test` expands, causing it to run more often!

image

## Summary by CodeRabbit

- **New Features**
	- Introduced a new testing framework focusing on complex scenarios, enhancing the scope of functionality being tested.
	- Added a comprehensive suite of unit tests for both SDK and Admin RPC server backends with sharded databases.

- **Bug Fixes**
	- Improved organization and clarity of test results through the introduction of a new struct for assertions.

- **Refactor**
	- Renamed and reorganized several test functions and packages to better reflect their purpose and improve code maintainability.

- **Chores**
	- Simplified CI workflow by updating job names and paths for testing, streamlining the testing process.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
     outputs:
       build: ${{ steps.ci-target-check.outputs.build }}
       bench: ${{ steps.ci-target-check.outputs.bench }}
-      sharding-test: ${{ steps.ci-target-check.outputs.sharding-test }}
+      complex-test: ${{ steps.ci-target-check.outputs.complex-test }}
 
     steps:
     - name: Checkout code
@@ -42,8 +42,10 @@ jobs:
             - 'admin/**'
             - 'api/converter/**'
           
-          sharding-test:
+          complex-test:
             - 'server/backend/database/**'
+            - 'pkg/document/**'
+            - 'client/**'
 
   build:
     name: build
@@ -133,12 +135,12 @@ jobs:
         github-token: ${{ secrets.GITHUB_TOKEN }}
         comment-always: true
 
-  sharding-test:
-    name: sharding-test
+  complex-test:
+    name: complex-test
     runs-on: ubuntu-latest
 
     needs: ci-target-check
-    if: ${{ needs.ci-target-check.outputs.sharding-test == 'true' }}
+    if: ${{ needs.ci-target-check.outputs.complex-test == 'true' }}
 
     steps:
 
@@ -171,5 +173,5 @@ jobs:
     - name: Initialize the Mongos
       run: docker compose -f build/docker/sharding/docker-compose.yml exec mongos1 mongosh test /scripts/init-mongos1.js
 
-    - name: Run the tests with sharding tag
-      run: go test -tags sharding -race -v ./test/sharding/...
+    - name: Run the tests with complex tag
+      run: go test -tags complex -race -v ./test/complex/...
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,9 +109,6 @@ jobs:
     - name: Check out code
       uses: actions/checkout@v4
 
-    - name: Get tools dependencies
-      run: make tools
-
     - name: Stack
       run: docker compose -f build/docker/docker-compose.yml up --build -d
 
@@ -152,9 +149,6 @@ jobs:
 
     - name: Check out code
       uses: actions/checkout@v4
-
-    - name: Get tools dependencies
-      run: make tools
 
     - name: Check Docker Compose Version
       run: docker compose --version
--- a/server/backend/database/testcases/testcases.go
+++ b/server/backend/database/testcases/testcases.go
@@ -48,6 +48,7 @@ const (
 
 // RunFindDocInfoTest runs the FindDocInfo test for the given db.
 // TODO(binary-ho) : temporary comment for measure CI test time
+// TODO(binary-ho) : temporary comment for measure CI test time
 func RunFindDocInfoTest(
 	t *testing.T,
 	db database.Database,
--- a/server/backend/database/testcases/testcases.go
+++ b/server/backend/database/testcases/testcases.go
@@ -47,6 +47,7 @@ const (
 )
 
 // RunFindDocInfoTest runs the FindDocInfo test for the given db.
+// TODO(binary-ho) : temporary comment for measure CI test time
 func RunFindDocInfoTest(
 	t *testing.T,
 	db database.Database,
--- a/server/backend/database/testcases/testcases.go
+++ b/server/backend/database/testcases/testcases.go
@@ -48,7 +48,6 @@ const (
 
 // RunFindDocInfoTest runs the FindDocInfo test for the given db.
 // TODO(binary-ho) : temporary comment for measure CI test time
-// TODO(binary-ho) : temporary comment for measure CI test time
 func RunFindDocInfoTest(
 	t *testing.T,
 	db database.Database,
--- a/server/backend/database/testcases/testcases.go
+++ b/server/backend/database/testcases/testcases.go
@@ -47,7 +47,7 @@ const (
 )
 
 // RunFindDocInfoTest runs the FindDocInfo test for the given db.
-// TODO(binary-ho): temporary comment for measure CI test time
+// TODO(binary-ho) : temporary comment for measure CI test time
 func RunFindDocInfoTest(
 	t *testing.T,
 	db database.Database,
--- a/server/backend/database/testcases/testcases.go
+++ b/server/backend/database/testcases/testcases.go
@@ -47,7 +47,6 @@ const (
 )
 
 // RunFindDocInfoTest runs the FindDocInfo test for the given db.
-// TODO(binary-ho) : temporary comment for measure CI test time
 func RunFindDocInfoTest(
 	t *testing.T,
 	db database.Database,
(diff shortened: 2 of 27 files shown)


krapie:
@binary-ho Could you rebase your commit to current upstream? There were some changes in previous commit.

krapie (commented):

I understand that this PR modifies two things. If my understanding is incorrect, please let me know.

1. Remove unnecessary `make tools` in CI workflow
2. Rename `sharding-test` to `complex-text` with moving `tree_concurrency_test.go` in `complex-text`, as @hackerwins suggested.

Seems like your work has reduced the CI time about 1 min, and I think this is huge 👍🏼
Considering that we are only running specific CI task like `complex-test` on specific file changes, this will synergy with that CI behavior. 

#### Leaving my small curiously:

Is there any reason for modifying `test/integration/main_test.go`? It seems like you switched function order and created `testResult` struct which seems to be originated from `tree_concurrency_test.go` and seems this struct is not being used?

binary-ho:
Thank you for review kevin!
also, sorry for not providing more context about testResult.

`testResult` type was defined in `concurrent_tree_test` but was also being used in array_test. 
Therefore, i could't leave the `testResult` type in `concurrent_tree_test`
it needed to stay within the integration package.

<br>

image

<br>

However, since the `testResult` type was used in multiple tests within the integration package 
and could potentially be used in the future as well, I decided to place it in test_main instead of `array_test`.

If you feel that placing the `testResult` type in `array_test.go` would be more appropriate according to principles like `YANGI`, please let me know.

krapie:
@binary-ho Then how about moving `testResult` struct in `test/complex/tree_concurrency_test.go` to `test/complex/main_test.go`, just like `integration` package?

Also I will understand switched function order as to keep consistency between two `main_test.go` files in `integration` and `complex` package.

binary-ho:
you're right.. Why didn’t I think of that? 
I have moved the `testResult` of the complex package to `main_test.go`! 

And the change in the function order was my mistake....
I thought I had checked it, but I missed it. I apologize for that

krapie (approved):
LGTM 👍🏼

PR #1010, opened 2024-09-17
Use random generated key

window9u:
## What this PR does / why we need it

This PR introduces a new method for generating API keys using short UUID  instead of the current time-based approach. This change aims to enhance security and reduce the predictability of our API keys.

Fixes #

#### 1. Custom Charset [0-9a-v]

- **Description**: Uses a custom charset of 32 characters (0-9 and a-v)
- **Process**: Generates random bytes and maps them to the charset
- **Output**: String of specified length
- **Example**: "f3c5h9m937etoad7l48ioc5oo1jr5p1h"
- **Pros**: Utilizes the same charset used in `xid` ([0-9a-v]).
- **Cons**: This encoding method may be less efficient or use more space.

```go
const charset = "0123456789abcdefghijklmnopqrstuv"
randomBytes := make([]byte, length)
_, err := rand.Read(randomBytes)
result := make([]byte, length)
for i := 0; i < length; i++ {
    result[i] = charset[randomBytes[i]%uint8(len(charset))]
}
```

#### 2. Base64 Encoding

- **Description**: Generates random bytes and encodes using base64 URL-safe encoding
- **Output**: Base64 encoded string
- **Example**: "zehf_23Bg2vrYk4VMZvgiQDqcgRgLRDcLBx3IvONyTY="
- **Pros**:  Easily allows setting the length of the key.
- **Cons**: Requires error handling, which makes the code more complex.

```go
bytes := make([]byte, length)
_, err := rand.Read(bytes)
if err != nil {
    return "", fmt.Errorf("failed to generate random bytes: %w", err)
}
return base64.URLEncoding.EncodeToString(bytes), nil
```

#### 3. ShortUUID(selected option)

- **Description**: Uses the shortuuid library
- **Process**: Generates a short UUID
- **Note**: Ignores the length parameter
- **Example**: "cqz7HZQp7VDm4Lg5otd3ZE"
- **Pros**: 
	- The code is the cleanest. 
	- Compresses the UUID using base57, reducing the string length.
- **Cons**: Cannot set the length of the key.
- **Selected reason**: I believe this option strikes a good balance between security and efficiency, and it offers the simplest and cleanest code.

```go
shortuuid.New()
```

#### 4. Standard UUID

- **Description**: Uses the standard UUID library
- **Process**: Generates a standard UUID (version 4)
- **Example**: "3023ef2e-3a04-404d-9f84-c08d77f43ab0"
- **Cons**:  Does not reduce the string length as effectively as the ShortUUID method, which compresses using base57.

```go
uuid.NewString()
```

## Notes for Reviewers

1. **Provide new API**: Consider creating an issue to implement an API for key refreshment.
2. **Uniqueness**: ~~This approach does not ensure absolute uniqueness, but the probability of collision is extremely low. See [UUID comment](https://github.com/google/uuid/blob/v1.6.0/version4.go#L34) for details.~~ MongoDB uniqueness constraint ensure this uniqueness [comment](https://github.com/yorkie-team/yorkie/pull/1010#issuecomment-2371580774)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/golang-jwt/jwt v3.2.2+incompatible
 	github.com/hashicorp/go-memdb v1.3.3
 	github.com/jedib0t/go-pretty/v6 v6.4.9
+	github.com/lithammer/shortuuid/v4 v4.0.0
 	github.com/prometheus/client_golang v1.13.0
 	github.com/rs/cors v1.10.1
 	github.com/rs/xid v1.5.0
@@ -29,8 +30,6 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.1
 )
-
-require github.com/stretchr/objx v0.5.2 // indirect
 
 require (
 	github.com/beorn7/perks v1.0.1 // indirect
--- a/go.sum
+++ b/go.sum
@@ -168,6 +168,7 @@ github.com/google/pprof v0.0.0-20201203190320-1bf35d6f28c2/go.mod h1:kpwsk12EmLe
 github.com/google/pprof v0.0.0-20201218002935-b9804c9f04c2/go.mod h1:kpwsk12EmLew5upagYY7GY0pfYCcupk39gWOCRROcvE=
 github.com/google/renameio v0.1.0/go.mod h1:KWCgfxg9yswjAJkECMjeO8J8rahYeXnNhOm40UhjYkI=
 github.com/google/uuid v1.1.2/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
+github.com/google/uuid v1.3.0/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
 github.com/google/uuid v1.6.0 h1:NIvaJDMOsjHA8n1jAhLSgzrAzy1Hgr+hNrb57e+94F0=
 github.com/google/uuid v1.6.0/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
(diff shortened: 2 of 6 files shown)


krapie (commented):
Below are some comments regarding this change.

#### Reason for this change?

> // TODO(hackerwins): Use random generated Key.

Are we changing because current key generation `xid` uses the Mongo Object ID algorithm and we don't want to be predictable, just like @window9u explained? Knowing the exact reason will help us determine if this approach is suitable. @hackerwins 

#### API Key rolling

> API Gateway: Consider creating an issue to implement an API gateway for key refreshment.

I agree with your opinion of having API for rolling (refreshing) API key by various cases like key exposure. But what does it mean by implementing `API Gateway`?

window9u:
@krapie 

 1. I noticed that the public and private keys are almost identical except for the last word. You can also verify this in your dashboard. This is why I attempted to change the xid to a randomly generated key.


2. I misused the term "gateway." As you correctly understood, I meant to add a key rolling API. I think we should reset the previously generated key using the xid.

krapie:
@sejongk Is there any constraints when we use UUID as project key considering our MongoDB environment especially for indexing and sharding?

sejongk:
Regarding the uniqueness of UUID, the `public_key` and `secret_key` in `project_info` are ensured to be unique in MongoDB thanks to the use of a uniqueness constraint.
https://github.com/yorkie-team/yorkie/blob/0871fac98488c7de245f19a395c0be1d0317ac19/server/backend/database/mongo/indexes.go#L73

sejongk:
> @sejongk Is there any constraints when we use UUID as project key considering our MongoDB environment especially for indexing and sharding?

@krapie In the current MongoDB sharding settings, `doc_key` and `client_key` are used as shard keys. Therefore, I believe there are no side effects on sharded data or the MongoDB sharding setup.

krapie (approved):
Since we do not put any information in our API key, I think (shorten) UUID is good approach for generating API key in terms of its randomness, uniqueness.

~~I was little worried about indexing of this API key in terms of retrieving `Project`, since we retrieve `Project` every time the API call is made. But since we are using cache while retrieving this, it will have little impact on the performance.~~

@window9u Could you create subsequence issue or PR for providing API for rolling these API keys?

window9u:
@krapie 

Thank you for reviewing my PR. I've created the issue!!

PR #1019, opened 2024-09-26
Update design for garbage collection and version vector

JOOHOJANG:
Since we're gonna use `version vector` to run `GC` after this. 
I just updated design document for garbage collection and version vector.

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything


design/garbage-collection.md
```diff
@@ -39,18 +77,85 @@ In the initial state, both clients have `"ab"`.
...
+min(c2.vv, c3.vv) = min({c1:3, c2:4, c3: 5}, {c1:3, c2:3, c3:5}) = 
+{c1:3, c2:3, c3:5}
+c2, c3 are acitve(attached).
+
+minVersionVector = {c1:3, c2:3, c3:5}.Filter([c1]) = {c2:3, c3:5}
+```
+
+After client receive this minVersionVector, it will filter its version vector to remove detached client's lamport.
+The next pushpull request will contains filtered version vector so that eventually db.version vector will store attached client's version vector only.
+![filter-version-vector](media/filter-version-vector.jpg)
```
emplam27, line 160: I think the image is different from the code above. If the image sync with the code above, It would be easier to understand!

design/garbage-collection.md
```diff
@@ -39,18 +77,85 @@ In the initial state, both clients have `"ab"`.
 
 ![garbage-collection-2](media/garbage-collection-2.png)
 
-Client `c1` deletes `"b"`, which is recorded as a change with logical timestamp `3`. The text node of `"b"` can be referenced by remote, so it is only marked as tombstone. And the client `c1` sends change `3` to server through PushPull API and receives as a response that `min_synced_seq` is `2`. Since all clients did not receive the deletion `change 3`, the text node is not purged by garbage collection.
+`Client a` deletes `"b"`, which is recorded as a change with versionVector `{b:1}`. The text node of `"b"` can be referenced by remote, so it is only marked as tombstone. And the `Client a` sends change `3a` to server through PushPull API and receives as a response that `minVersionVector` is `{a:1, b:2}`. Since all clients did not receive the deletion `change 3a`, the text node is not purged by garbage collection.
 
-Meanwhile, client `c2` inserts `"c"` after textnode `"b"`.
+Meanwhile, `client b` inserts `"c"` after textnode `"b"`.
```
emplam27, line 82: `client b`'s change is local change, so there are not send change `2b` to server and it changes only in local. Am I understanding this correctly?

design/garbage-collection.md
```diff
@@ -13,22 +19,54 @@ Yorkie provides garbage collection to solve this problem.
...
+
+ex)
+min([c1:2, c2:3, c3:4], [c1:3, c2:1, c3:5, c4:3])
+=> [c1:2, c2:1, c3:4, c4:0]
+
+```
+### An example of garbage collection:
 #### State 1
 
 ![garbage-collection-1](media/garbage-collection-1.png)
```
emplam27, line 72: How is the image's situation created? Is it like this below or just an example situation?

image

design/garbage-collection.md
```diff
@@ -39,18 +77,85 @@ In the initial state, both clients have `"ab"`.
...
 
-Client `c2` pushes change `4` to server and receives as a response that `min_synced_seq` is `3`. After the client applies change `4`, the contents of document are changed to `ac`. This time, all clients have received change `3`, so textnode `"b"` is completely removed.
+`Client b` pushes change `3b` to server and receives as a response that `minVersionVector` is `{a:1, b:1}`. After the client applies change `4`, the contents of document are changed to `ac`. This time, all clients have received change `3a`. Since node "b" is removed at `3a`, it's still marked as tombstone for every clients, because `minVersionVector[a] = 1 < 3` 
 
 #### State 4
 
 ![garbage-collection-4](media/garbage-collection-4.png)
 
-Finally, after client `c1` receives change `4` from server, purges textnode `"b"` because it is no longer referenced remotely.
+`Client a` pushpull but nothing to push or pull. `minVersionVector` is still `{a:1, b:1}`, so no GC happens.
```
emplam27, line 94: `client a` is pull change `3b` in this image. is it correct?

design/garbage-collection.md
```diff
@@ -39,18 +77,85 @@ In the initial state, both clients have `"ab"`.
...
+
+#### State 5
+
+![garbage-collection-4](media/garbage-collection-5.png)
+
+`Client b` pushpull but nothing to push or pull. `minVersionVector` is now `{a:3, b:1}`, node "b" removedAt is `3@a`, and `minVersionVector[a] = 3 >= 3`, thus `client b` meets GC condition
```
emplam27, line 100: why `client b` pull `minVersionVector{a:3, b:1}`? Above image `garbage-collection-4.png`, `client a` pull `minVersionVector{a:1, b:1}`.

emplam27 (commented):
What a great document! it's good to understand about VV and GC👍👍👍

design/version-vector.md
```diff
@@ -0,0 +1,66 @@
...
+
+### The reason for choosing Version Vector To Replace Lamport Based Logical Clock System
+Yorkie provides Garbage Collection for removed (tombstone) nodes. The previously implemented Lamport-based ID system triggers a crucial problem with causality.
+You can see detail from [garbage collection design document](./garbage-collection.md).
+
+### What is Version Vector
+
+A Version Vector is a mechanism used in distributed systems to track the causal relationships between different replicas or nodes. Each node maintains a vector, where each element corresponds to a version counter for every other node in the system, including itself. Whenever a node updates its state, it increments its own counter. When nodes exchange information, they merge their vectors by taking the maximum value for each element, ensuring that the system knows which updates happened concurrently or which one came first. Version Vectors are particularly useful for detecting conflicts and maintaining consistency in distributed databases by providing a more detailed view of the system's state than a single logical clock.
+
+![version-vector-example](media/version-vector-example.jpg)
```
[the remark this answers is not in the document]
JOOHOJANG:
```
Whenever a node updates its state, it increments its own counter. When nodes exchange information, they merge their vectors by taking the maximum value for each element, 
```
This part can be confusing because the case that you've mentioned is `Vector Clock`.
Also, implementation can be diverged but the important part is maintaining the concept of Version Vector. 
version at B@3 is always identical at any time. 

For your information, here is a specific explanation. A's lamport increases at A@2, and when B gets A@2, it'll be max([2, 0, 0], [0, 0, 0])  => [2, 0, 0], then we need to compute B's new lamport which is max(2, 1).

design/garbage-collection.md
```diff
@@ -13,22 +19,54 @@ Yorkie provides garbage collection to solve this problem.
...
+![prev-gc-issue](media/prev-gc-issue.jpg)
+
+As you can see from the image above, `min_synced_seq` doesn't guarantee every client receives change.  
+
 ### How It Works
 
 Garbage collection checks that deleted nodes are no longer referenced remotely and purges them completely.
 
-Server records the logical timestamp of the last change pulled by the client whenever the client requests PushPull. And Server returns the smallest logical timestamp, `min_synced_seq` of all clients in response PushPull to the client. `min_synced_seq` is used to check whether deleted nodes are no longer to be referenced remotely or not.
+Server records the version vector of the last change pulled by the client whenever the client requests PushPull. And Server returns the vin version vector, `minVersionVector` of all clients in response PushPull to the client. `minVersionVector` is used to check whether deleted nodes are no longer to be referenced remotely or not.
```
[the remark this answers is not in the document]
JOOHOJANG: Oh my bad. I'll fix it

design/garbage-collection.md
JOOHOJANG: Ah my bad, I'll fix it 
Thanks for your correction

design/garbage-collection.md
JOOHOJANG: You're right. client b's change has not been sent in that moment(so it stays as local change).
Only client a's change has been sent.

design/garbage-collection.md
JOOHOJANG: A updates and sends changes to the server, and at the same time, B does the same thing. This situation could happen.

design/garbage-collection.md
emplam27: This code filtered `c1`'s vv. i think this code will be filterd `c3`'s vv
```
minVersionVector = {c1:3, c2:3, c3:5}.Filter([c1]) = {c1:3, c2:3}
```

emplam27 (approved):
LGTM!😄

design/garbage-collection.md
emplam27: How about explaining local change for unfamiliar member to yorkie?

design/version-vector.md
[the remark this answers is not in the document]
emplam27: thx! understand it!

PR #1035, opened 2024-10-16
Add CLI for migration

JOOHOJANG:
Define semi-auto migration architecture and add migration scripts for add version vector.

**Process**
```
1. make new version directory in migrations
2. add migration script files.
3. add version into migrationMap in /cmd/migration.go
```
**Usage example** 
```
$ ./yorkie migration --from v0.5.1 --to v0.6.0 --mongo-connection-uri mongodb://localhost:27017 --batch-size 10000 --database-name yorkie-meta
```

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

### DIFF
--- a/cmd/yorkie/migration.go
+++ b/cmd/yorkie/migration.go
@@ -19,8 +19,7 @@ package main
 import (
 	"context"
 	"fmt"
-	"os"
-	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


hackerwins (commented):
Thanks for your contribution. Overall looks good. I have one question.

Currently, Yorkie is built and distributed as a single binary(`yorkie`). In this change, it seems that Yorkie CLI is designed to execute by referencing the `migrations` directory.

While `migrations` directory is included in the binary during the build, it is unclear why the command(`yorkie migration`) still needs to reference the directory.

It would be better to retrieve and execute the target scripts based on the `migrationMap` rather than external directory. I'm curious about your opinion.

JOOHOJANG:
> Thanks for your contribution. Overall looks good. I have a few questions.
> 
> A. `migrations` directory
> 
> Currently, Yorkie is built and distributed as a single binary(`yorkie`). In this change, it seems that Yorkie CLI is designed to execute by referencing the `migrations` directory.
> 
> While `migrations` directory is included in the binary during the build, it is unclear why the command(`yorkie migration`) still needs to reference the directory.
> 
> It would be better to retrieve and execute the target scripts based on the `migrationMap` rather than external directory. I'm curious about your opinion.
> 
> B. Batch Processing for `Find`
> 
> Please, consider implementing batch processing for `Find`. The current approach of loading the entire collection into memory could lead to excessive memory usage.
> 
> ```go
> collection := db.Database(databaseName).Collection("changes")
> 
> cursor, err := collection.Find(ctx, bson.M{})
> if err != nil {
>   return err
> }
> ```

A. Agree, you're right. Watching migration map is enough. I'll fix it.
B. `Cursor` has default batch size (101 documents) so we should not increase it unless we need a larger batch size. https://www.mongodb.com/docs/manual/tutorial/iterate-a-cursor/

hackerwins (approved):
Thanks for your contribution.

PR #1047, opened 2024-10-22
Introduce VersionVector

### LINKED ISSUE #723: Incorrect Garbage Collection Due to MinSyncedSeq Determination with Lamport Clocks

JOOHOJANG:
The purpose of this PR is to address a defect found in the existing GC that used syncedSeq, by introducing VersionVector to resolve the issue.

The diagram of the structure can be found at this [link](https://miro.com/app/board/o9J_kidjCtk=/?moveToWidget=3458764594957276897&cot=14).​

Here are the four key points to focus on:

1. What is Lamport Synced Version Vector and how to use it
2. How to store version vector into database and update it when receives change
3. What is min version vector and how to solve GC problem
4. How to handle detached client's version vector

### 1. What is Lamport Synced Version Vector
[Definition of Version Vector](https://en.wikipedia.org/wiki/Version_vector)
Since we're using lamport timestamp to handle CRDT, we implement Lamport Synced Version Vector which is quite different from original Version Vector.
Lamport Synced Version Vector allows us to continue using the existing method that relied on Lamport.
It is key:value map consist of actorID and lamport.

image
As you can see from the image above, which is typical version vector example, client's value(lamport in our case) is not the largest at every moment.

image
At the above image, you can see that the client’s Lamport is always the largest, which is consistent with how we used to update and utilize lamport in the past.

For convenience, I will refer to Lamport Synced Version Vector as the version vector from now on.

When creating a new change, we increment both the lamport and the lamport in the version vector to use as the ID. Thus, we don't need to use version vector as ID, but use as a part of ID. You can see how it is used in the modified code (see changeID parts).

When receiving a new change, compute max version vector by using version vector from change and document's version vector.

```
max(vv1, vv2) =>
for (key in vv1 || key in vv2) {
    maxVV[key] = max(vv1[key] || 0, vv2[key] || 0)
}


ex)
max([c1:2, c2:3, c3:4], [c1:3, c2:1, c3:5, c4:3])
=> [c1:3, c2:3, c3:5, c4:3]

```
```
min(vv1, vv2) =>
for (key in vv1 || key in vv2) {
    minVV[key] = min(vv1[key] || 0, vv2[key] || 0)
}

ex)
min([c1:2, c2:3, c3:4], [c1:3, c2:1, c3:5, c4:3])
=> [c1:2, c2:1, c3:4, c4:0]

```

### 2. How to store version vector into database and update it when receives change
1. Attach
image

- set initial version vector ([c1: 0])
- make change which includes updated changeID (version vector [c1: 1])
- push change into server and change table in database.
- push version vector into version vector table in database
    - you can see the detailed logic from modified codes (`updateVersionVector`, `updateAndFindMinSyncedVersionVectorAfterPushPull`)

2. Pushpull
- pushonly
    - there's no received changes when its pushonly, thus no need to update document's version vector.
- pushpull
    - after receives change, we update document's version vector when every change has been applied. 
```
1. update changeID
(except vv, everything is same as before)

doc.vv = max(doc.vv, doc.vv)
doc.vv[doc.actorid] = maxLamport + 1
```

### 3. What is min version vector and how to solve GC problem
Min version vector is the vector which consists of minimum value of every version vector stored in version vector table in database.

Conceptually, min version vector is version vector that is uniformly applied to all users. It ensures that all users have at least this version applied. Which means that it is safe to perform GC.

You can see detailed logic from `UpdateAndFindMinVersionVectorAfterPushPull` in modified codes.
Suppose clientA is performing pushpull

1. if there're version vectors in database, collect them.
2. collect clientA's version vector from change pack
3. store clientA's version vector into database
4. compute min from collected version vectors

GC algorithm
```
if (removedAt.lamport <= minSyncedVersionVector[removedAt.actor]) {
    GC()
}
```
If node.removedAt.lamport is smaller or equal to every value in min version vector, then run GC.
It's okay to run GC when its equal, because the meaning of min version vector is that every client's version vector is at least larger then min version vector.

### 4. How to handle detached client's version vector
If detached client's version vector remains in database and local document, its memory wasting and also GC will not performed.

```
1. minVV = [c1:3, c2:4, c3:4]
2. c3 detach
3. c2 remove nodes k , k > 4
4. c1, c2 keep updating and perform pushpull
5. minVV at some moment [c1: n, c2: m, c3:4] where n, m > k > 4
6. k > 4, can't perform GC
```
image

So it's necessary to remove detached client's version vector and its lamport from other's version vector.

1. When client detaches, remove its version vector from version vector table
2. remove client's lamport from every version vector stored in database.
3. when other client perform pushing and its version vector includes detached client's lamport, filter it by using min version vector.
    - the reason why we're using min version vector in this case is that we already compute min version vector (to return min version vector to perform GC).
    - min version vector includes all attached users in the document.
4. store filtered version vector into database.
5. when other client perform pulling, we also use min version vector to filter it after apply change.
6. after few times of pushpull, detached client's lamport will be totally removed from database and others document.

---

By the way, the code is a bit messy, so any reviews are always welcome. If you have any ideas for improvements, feel free to share them.

---

Address #723

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

## Summary by CodeRabbit

- **New Features**
  - Introduced version vector support across various components, enhancing change tracking and synchronization.
  - Added `VersionVector` field to multiple response schemas and structures, including `GetSnapshotMetaResponse` and `ServerPack`.
  - New methods for managing version vectors added to the database interface, improving client synchronization.

- **Bug Fixes**
  - Improved error handling in multiple functions, ensuring robust responses to invalid inputs and operations.
  - Enhanced garbage collection logic to utilize version vectors, ensuring consistent behavior across concurrent operations.

- **Documentation**
  - Updated OpenAPI specifications to reflect changes in versioning and added new schemas for version vectors.
  - Revised design documents to clarify the role of version vectors in garbage collection and synchronization.

- **Tests**
  - Enhanced test coverage for version vector functionality, including garbage collection and document manipulation scenarios.
  - Added new tests to validate version vector behavior under concurrent updates and edge cases.

- **Chores**
  - Refactored internal logic to replace deprecated ticket-based systems with version vector-based mechanisms.
  - Updated database schema to include new tables and indexes for version vectors, improving data retrieval.


hackerwins (approved):
Thanks for your contribution.

PR #1068, opened 2024-11-14
Fix version vector cleanup during client deactivation

### LINKED ISSUE #1058: Deletion of the version vector of a deactivated client is not performed properly
**What happened**:
When performing the `deactivate` operation on the client, it detaches all attached documents and then attempts to remove the corresponding version vector using `updateVersionVector()`. However, `updateVersionVector()` currently checks the status of `clientDocInfo` and removes the version vector even if the status is not "attached," due to the reliance on the outdated information prior to detachment. This results in the version vector not being removed as expected.

https://github.com/yorkie-team/yorkie/blob/19c818e390a83c292fa090b3bd141645e69d7ecc/server/clients/clients.go#L87-L101

The version vector should only be removed if the `clientDocInfo` status is confirmed to be not "attached" after the successful detachment of documents.

1. Attach a document to a client.
2. Execute the `deactivate` function for the client. 
3. Observe the outcome of the version vector removal process. (You can check by attaching a new client and compare the size of the document's version vector, or simply check whether the version vector of the deactivated client remains in the database.)

**Environment**:
- Operating system:
- Browser and version:
- Yorkie version (use `yorkie version`): 0.5.4
- Yorkie JS SDK version:

raararaara:
This PR resolves the issue where the version vector is not being correctly removed during the document detach process caused by client deactivation. 

When `deactivate` is called, the `pushpull` process within `detachDocument` properly deletes the version vector. However, after a successful completion of `detachDocument`, an additional call to `UpdateVersionVector` mistakenly attempts to upsert the version vector for a client that has already been removed, causing it to appear as though the deletion did not occur.

By removing the call to `UpdateVersionVector` during the client deactivation process, this issue is effectively resolved.

Fixes #1058 

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Didn't break anything

## Summary by CodeRabbit

- **Bug Fixes**
	- Removed unnecessary updates to document version tracking upon client deactivation to streamline functionality.

- **Tests**
	- Enhanced clarity of test assertions related to garbage collection, ensuring explicit comparisons for expected outcomes.
	- Introduced a new test case for garbage collection involving tree structures to validate version vector checks.


raararaara:
Because the fact that no error occurred during the detach process guarantees that the detach was completed successfully, I manually manipulated `clientInfo`.

Are there any other problems I might encounter with this?

hackerwins (approved):
Thanks for your contribution.

PR #1073, opened 2024-11-20
Provide additional explanation for GC design document.

JOOHOJANG:
I just added additional explanation for GC for the case when last active client deactivates.
It leads to a such result that min version vector is empty. 
`VersionVector.minLamport()` returns int64max value when min version vector is empty, this is because empty min version vector means there's no more client using document and it means its safe to run GC to remove all tombstones.

By the way, this concept can lead readers to confusion, I just added additional explanation of such case. 

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

- **Documentation**
	- Updated the garbage collection document to include a new section on handling empty `minVersionVector` scenarios.
	- Clarified the logic for calculating the minimum Lamport value and its implications for the garbage collection process.

### DIFF
--- a/design/garbage-collection.md
+++ b/design/garbage-collection.md
@@ -173,6 +173,21 @@ What’s essential here is having a consistent criterion. If we take the node’
 
 ![remove-detached-clients-tombstone](media/remove-datached-clients-tombstone.jpg)
 
+### What if MinVersionVector is empty?
+Since minVersionVector is consist of lamport from active clients, there's a case where min version vector is empty.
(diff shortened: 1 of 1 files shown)


chacha912:
Currently, we have decided to maintain the detached client lamport in VV, and all related changes have been updated. (#1090) I have documented the progress in a separate issue.(#1144) For now, I will close this PR and proceed with supplementing the documentation later.

PR #1088, opened 2024-12-03
Replace maxCreatedAtMap with version vector for causal-concurrent operations

chacha912:

This PR replaces the usage of `MaxCreatedAtMapByActor` with `version vector` for operations that handle causal and concurrent relationships. (`Text.Edit`, `Text.Style`, `Tree.Edit`, and `Tree.Style`)

The introduction of version vector allows us to track which client's Lamport time existed when a change was created. For example, during delete operations, we compare the Lamport time of the deleted character's creation with the Lamport time from the version vector to maintain proper concurrency handling.

As shown in the diagram, when checking if a node should be removed, we compare:
- Node's createdAt (e.g., `6@A`)  
- Client's Lamport time at change creation (e.g., `2@A`)

image

Related #723 #1047

There are two issues with replacing maxCreatedAt with VV (Version Vector):

1. Since the information of detached clients is removed from VV, the client's lamport value is set to 0 at that point. When a client's lamport value is not present in VV, we need to be able to distinguish between two cases: whether we haven't received the client's changes yet, or whether the client has been detached.

image

[Resolved] Modified to retain client information in change VV even after detachment, ensuring all client states remain traceable. (#1090)

2. Considering backward compatibility, since the version vectors for previous changes are not accurate (when performing DB migration, only the lamport of the change actor ID was added), we cannot use the lamport values obtained from there. Therefore, to maintain backward compatibility, we need to keep maxCreatedAt.

[Resolved] Since the version vectors in existing data are uncertain, we can proceed after removing all this information. If a change has no version vector, maxCreatedAt will be used; otherwise, we use the version vector to verify nodeExisted.

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

- **New Features**
	- Introduced new methods for managing version vectors.
	- Added tests for document lifecycle and garbage collection interactions.

- **Bug Fixes**
	- Enhanced handling of node deletion and styling based on updated parameters.

- **Documentation**
	- Updated method signatures to reflect new parameters and structures.

- **Refactor**
	- Streamlined method signatures across various operations to include options for enhanced flexibility.

### DIFF
--- a/pkg/document/change/change.go
+++ b/pkg/document/change/change.go
@@ -54,7 +54,7 @@ func New(id ID, message string, operations []operations.Operation, p *innerprese
 // Execute applies this change to the given JSON root.
 func (c *Change) Execute(root *crdt.Root, presences *innerpresence.Map) error {
 	for _, op := range c.operations {
-		if err := op.Execute(root); err != nil {
+		if err := op.Execute(root, operations.WithVersionVector(c.ID().versionVector)); err != nil {
 			return err
 		}
 	}
--- a/pkg/document/change/change.go
+++ b/pkg/document/change/change.go
@@ -52,10 +52,17 @@ func New(id ID, message string, operations []operations.Operation, p *innerprese
 }
 
 // Execute applies this change to the given JSON root.
-func (c *Change) Execute(root *crdt.Root, presences *innerpresence.Map) error {
+func (c *Change) Execute(root *crdt.Root, presences *innerpresence.Map, opSource string) error {
 	for _, op := range c.operations {
-		if err := op.Execute(root, operations.WithVersionVector(c.ID().versionVector)); err != nil {
-			return err
+		// TODO(chacha912): Remove opSource after testing.
+		if opSource == "remote" {
+			if err := op.Execute(root, operations.WithVersionVector(c.ID().versionVector)); err != nil {
+				return err
+			}
+		} else {
+			if err := op.Execute(root); err != nil {
+				return err
+			}
 		}
 	}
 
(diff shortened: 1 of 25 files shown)


pkg/document/change/change.go
```diff
@@ -54,7 +54,7 @@ func New(id ID, message string, operations []operations.Operation, p *innerprese
 // Execute applies this change to the given JSON root.
 func (c *Change) Execute(root *crdt.Root, presences *innerpresence.Map) error {
 	for _, op := range c.operations {
-		if err := op.Execute(root); err != nil {
+		if err := op.Execute(root, operations.WithVersionVector(c.ID().versionVector)); err != nil {
```
hackerwins, line 57: Currently we only have a single option(`VersionVector`) which makes the pattern overhead unnecessary and JS SDK doesn't introduce option pattern so keeping the interfaces aligned would improve consistency.

So it seems that removing the Option pattern from the Operation interface is good for now. Is there a any other reason what I missing for this?

hackerwins (approved):
Thanks for your contribution.

I left a trivial comment.

pkg/document/change/change.go
chacha912: Thanks for the review. I've modified it.

pkg/document/operations/style.go
```diff
@@ -63,14 +63,14 @@ func NewStyle(
...
-func (e *Style) Execute(root *crdt.Root) error {
+func (e *Style) Execute(root *crdt.Root, versionVector time.VersionVector) error {
 	parent := root.FindByCreatedAt(e.parentCreatedAt)
 	obj, ok := parent.(*crdt.Text)
 	if !ok {
 		return ErrNotApplicableDataType
 	}
 
-	_, pairs, err := obj.Style(e.from, e.to, e.maxCreatedAtMapByActor, e.attributes, e.executedAt)
+	_, pairs, err := obj.Style(e.from, e.to, e.maxCreatedAtMapByActor, e.attributes, e.executedAt, versionVector)
```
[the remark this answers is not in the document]
chacha912: I've checked the parameter order carefully, and it appears to be in the correct order. Could you point out where you see the mismatch?

PR #718, opened 2023-12-10
Add document hard deletion functionality to housekeeping tasks.

fourjae:
**What this PR does / why we need it**:
A housekeeping task was added to automatically physically delete documents at certain intervals.

**Which issue(s) this PR fixes**:

Fixes # https://github.com/yorkie-team/yorkie/issues/706

Although it does not directly solve the above issue, I think it is somewhat related to the need to improve the housekeeping structure.
I think the housekeeping structure could be improved if necessary.

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Added relevant tests or not required
- [x] Didn't break anything

## Summary by CodeRabbit

- **New Features**
	- Enhanced housekeeping functionality for managing document deletions and client deactivations with more granular control.
	- New configuration parameters for housekeeping intervals and limits, including document deletion grace periods.
	- Introduction of methods for identifying and deleting documents eligible for hard deletion based on project specifics.

- **Bug Fixes**
	- Improved error handling during document deletion processes.

- **Documentation**
	- Updated configuration files and sample files to reflect new housekeeping parameters and their usage.
	- Expanded test coverage to validate new configuration structures and parameters.

PR #484, opened 2023-03-06
Add RemoveDocument API

hackerwins:
**What this PR does / why we need it**:

Add document deletion API.

Document deletion is crucial feature to save resources. Without deletion, accumulated documents and its relevant data will waste storage resources.

To overall document deletion process will look like this:

1. Client sends document delete API(`RemoveDocument`) with  `DocIsRemoved = true` in request `ChangePack`
2. In `PushPull` of *`RemoveDocument`(or other APIs),* check `DocIsRemoved`. And if `DocIsRemoved = true`, simply set `RemovedAt` date in `docInfo` to logically(soft) remove document.
3. Set response `ChangePack`'s `DocIsRemoved` to `true` to inform document deletion.
4. After certain period of time(`DocumentRemoveDuration`), document is physically removed from database.

To successfully implement this API, we need to seperate our tasks and implement them step by step.

- [x]  1. Implement basic document deletion API with simply setting `RemovedAt` date in `docInfo`
- [ ]  2. Determine whether to include removed documents in `ListDocuments` API used by Dashboard and CLI(Admin)
- [ ]  3. Implement housekeeping background process to physically remove document(`is_removed = true`) and its relevant data.

**Which issue(s) this PR fixes**:

Address #464

**Special notes for your reviewer**:

Discussion about document deletion API in codepair: [https://codepair.yorkie.dev/ylvpas](https://codepair.yorkie.dev/ylvpas)

Previous closed PR for implementing document deletion API: [https://github.com/yorkie-team/yorkie/pull/468](https://github.com/yorkie-team/yorkie/pull/468)

WatchDocuments API also needs to find documents by `ID`, not by `key`.

**Does this PR introduce a user-facing change?**:

```
adds Client.Remove(document), *RemoveDocument API*
```

**Additional documentation**:

```docs

```

**Checklist**:
- [x] Added relevant tests or not required
- [x] Didn't break anything

PR #1113, opened 2024-12-26
Introduce project event webhook

### LINKED ISSUE #1002: Provide DocEvent Webhook

We are currently implementing an LLM-based document search functionality in CodePair. As part of this, we need to maintain a vector of document content in Vector Store. It's crucial that any updates to the document are reflected in the Vector Store by continually editing the content.

To achieve this, we require a mechanism that notifies external services like CodePair when documents are modified in Yorkie. We propose the introduction of a Webhook system that triggers when a document event occurs.

Specifically, we suggest that when handling the PushPullChanges requests, the server should check if a Webhook for the DocEvent is registered for the project. If it is, the server would call that Webhook during the background routine of the PushPullChanges API execution, right before publishing the DocEvent.

I think it will have a similar structure to the [Auth Webhook](https://yorkie.dev/docs/cli#auth-webhook), and if changes occur frequently, an event control device such as debouncing will also be needed.

**Why is this needed**:

This enhancement would enable seamless integration with external services, allowing for real-time updates to Search Engine or Vector Store based on document changes in Yorkie, thereby enhancing the overall document management and search capabilities of our application.

window9u:

This PR implements **document creation** and **document removal** events for a new **project event webhook**. Currently, users who uses Yorkie as a SaaS are not always aware of document status changes. By introducing project-level events—rather than relying solely on doc-level events like `attach` or `push-pull`—we can more clearly represent the lifecycle of documents and make those events available to external systems.

**Which issue(s) this PR fixes**: Fixes #1002

1. **Project event webhook**
- The original issue suggested a `DocEvent`, but instead, I'm introducing a `Project Event`. We already have `DocEvent` events (`DocumentChanged` `DocumentWatched`, `DocumentUnwatched`, `broadcast`), and using the same naming might be confusing.
- Additionally, doc-level events(e.g., `attach`, `document-changed`, `watched`, `unwatched`) don't clearly convey CRUD semantics. For instances:
	- **Create**: A document is automatically created when a user attaches to a non-existent document. `documentAttached` alone doesn't explicitly indicate **creation**.
	- **Read/Update**: `push-pull` may imply either reading or updating a document, and it's not always obvious which. (A `Change` can contain both `presence` and `operations`; if it only has `presence`, it might not mean a real update of document's content.)
     - **Delete**: Yorkie's `removeDocument` function only marks a document as "removed" without fully deleting it. In contrast, the `removeDocumentByAdmin` function permanently deletes the document. When a `removeDocument` action occurs, Yorkie sends webhook events, allowing the customer's server to decide the next steps—such as moving the document to a recycle bin or directly deleting it using `removeDocumentByAdmin`.

2. **Add HMAC signature in request header**
- For the authentication webhook, we provide a token so the receiver can validate the request. However, for the project event webhook, the endpoint may not know which service is calling.
- To address this, I introduced an HMAC signature in the `X-Signature` header, generated using the project's secret key. The receiving server can verify this signature to ensure the request is from Yorkie.
- In another approach, we could set our public key in `projectKey` in our request attributes.

```go
type WebhookAttribute struct {
        ProjectKey string `json:"projectKey"`
	DocumentKey string `json:"documentKey"`
	ClientKey   string `json:"clientKey"`
	IssuedAt    string `json:"issuedAt"`
}
```

3. **Handling timeouts and unreachable events**
	- **Asynchronous work**: Authentication must be verified before proceeding, so the auth webhook is synchronous. In contrast, a project event occurs after some action is completed, so it can be handled asynchronously without impacting user response times.
	- **Unreachable events and replay**: If an endpoint is down or encounters an error, we need a way to track failed events. In this PR, we only log these failures; in the future, the Yorkie dashboard could handle or additional recovery steps.
	- **More relaxed timeouts**: Currently, the auth webhook uses exponential backoff with a base of 200ms and a maximum wait of 3 seconds. This PR applies the same settings for consistency, although we might later adopt a more extended retry schedule(e.g., Liveblocks' approach: `[immediately, 5s, 5m, 30m, 2h, 5h, 10h]`, etc.).

4. **Add an HTTP client with timeout**
-  Since `http.POST` does not allow specifying a request timeout directly, using an HTTP client with a configurable timeout is more predictable. Although the auth webhook still uses `http.POST`, we may consider migrating it to an HTTP client for consistency.

#### In this PR

This design proposal differs somewhat from the original issue’s requirements. If the idea seems sound, I’ll proceed with the remaining tasks:

- Add config validation and tests
- Add `event_webhook` tests
- Update the admin server’s `UpdateProjectRequest` API spec
- Integrate the auth webhook’s exponential backoff logic into `pkg/webhook`

### In a subsequent PR
- Introduce a `document-changed` event (with caching)

### Does this PR introduce a user-facing change?

```release-note

```

### Additional documentation
There are four configuration options related to timeouts, which can be confusing. Below is a brief explanation:

- **EventWebhookRequestTimeout**: How long to wait for a response before timing out
- **EventWebhookMaxRetries**: Number of retry attempts
- **EventWebhookBaseWaitInterval**: Base interval for exponential backoff
- **EventWebhookMaxWaitInterval**: Maximum interval between retries

image

```docs
N/A
```

### Checklist

- [ ]  Added relevant tests or not required
- [ ]  Addressed and resolved all CodeRabbit review comments
- [x]  Ensured no breaking changes

## Summary by CodeRabbit

- **New Features**
	- Added event webhook functionality for project documents.
	- Introduced support for tracking document creation and removal events.
	- Implemented HMAC-based authentication for webhook requests.
	- Added configurable webhook retry and timeout mechanisms.

- **Improvements**
	- Enhanced server configuration with new webhook-related parameters.
	- Implemented exponential backoff for webhook retries.
	- Improved error handling and logging for webhook events.
	- Expanded API documentation to include new webhook properties.
	- Enhanced validation for new event webhook fields in project configurations.


window9u:
I notice that the `RemoveDocument` feature has more context discussed in issue #484. In that PR, three tasks were outlined:
1. **Implement a basic document deletion API** that sets the `RemovedAt` date in `docInfo`.  
2. **Decide whether to include removed documents** in the `ListDocuments` API, which is used by the Dashboard and CLI (Admin).  
3. **Create a housekeeping background process** to physically delete documents (`is_removed = true`) and their related data.  

The first task has been completed, and work on the third task is ongoing in PR #718.

Given this, I think it makes sense to publish the `documentRemoved` webhook as part of the third task, which physically removing documents. So I'll remove the publishing `removeDocument` event.

window9u:
This PR was opened in a forked repository and will be reopened from a branch in the origin repository

PR #1131, opened 2025-01-31
Introduce project event webhook

window9u:

This PR implements **document creation** and **document removal** events for a new **project event webhook**. Currently, users who uses Yorkie as a SaaS are not always aware of document status changes. By introducing project-level events—rather than relying solely on doc-level events like `attach` or `push-pull`—we can more clearly represent the lifecycle of documents and make those events available to external systems.

**Which issue(s) this PR fixes**: Fixes [#1002](https://github.com/yorkie-team/yorkie/issues/1002)

1. **Project event webhook**

- The original issue suggested a `DocEvent`, but instead, I'm introducing a `Project Event`. We already have `DocEvent` events (`DocumentChanged` `DocumentWatched`, `DocumentUnwatched`, `broadcast`), and using the same naming might be confusing.
- Additionally, doc-level events(e.g., `attach`, `document-changed`, `watched`, `unwatched`) don't clearly convey CRUD semantics. For instances:
    - **Create**: A document is automatically created when a user attaches to a non-existent document. `documentAttached` alone doesn't explicitly indicate **creation**.
    - **Read/Update**: `push-pull` may imply either reading or updating a document, and it's not always obvious which. (A `Change` can contain both `presence` and `operations`; if it only has `presence`, it might not mean a real update of document's content.)
    - **Delete**: Yorkie's `removeDocument` function only marks a document as "removed" without fully deleting it. In contrast, the `removeDocumentByAdmin` function permanently deletes the document. When a `removeDocument` action occurs, Yorkie sends webhook events, allowing the customer's server to decide the next steps—such as moving the document to a recycle bin or directly deleting it using `removeDocumentByAdmin`.

2. **Add HMAC signature in request header**

- For the authentication webhook, we provide a token so the receiver can validate the request. However, for the project event webhook, the endpoint may not know which service is calling.
- To address this, I introduced an HMAC signature in the `X-Signature` header, generated using the project's secret key. The receiving server can verify this signature to ensure the request is from Yorkie.
- In another approach, we could set our public key in `projectKey` in our request attributes.

```go
type WebhookAttribute struct {
        ProjectKey string `json:"projectKey"`
	DocumentKey string `json:"documentKey"`
	ClientKey   string `json:"clientKey"`
	IssuedAt    string `json:"issuedAt"`
}
```

3. **Handling timeouts and unreachable events**
    
    - **Asynchronous work**: Authentication must be verified before proceeding, so the auth webhook is synchronous. In contrast, a project event occurs after some action is completed, so it can be handled asynchronously without impacting user response times.
    - **Unreachable events and replay**: If an endpoint is down or encounters an error, we need a way to track failed events. In this PR, we only log these failures; in the future, the Yorkie dashboard could handle or additional recovery steps.
    - **More relaxed timeouts**: Currently, the auth webhook uses exponential backoff with a base of 200ms and a maximum wait of 3 seconds. This PR applies the same settings for consistency, although we might later adopt a more extended retry schedule(e.g., Liveblocks' approach: `[immediately, 5s, 5m, 30m, 2h, 5h, 10h]`, etc.).
4. **Add an HTTP client with timeout**
    

- Since `http.POST` does not allow specifying a request timeout directly, using an HTTP client with a configurable timeout is more predictable. Although the auth webhook still uses `http.POST`, we may consider migrating it to an HTTP client for consistency.

#### In this PR

This design proposal differs somewhat from the original issue’s requirements. If the idea seems sound, I’ll proceed with the remaining tasks:

- Add config validation and tests
- Add `event_webhook` tests
- Update the admin server’s `UpdateProjectRequest` API spec
- Integrate the auth webhook’s exponential backoff logic into `pkg/webhook`

### In a subsequent PR

- Introduce a `document-changed` event (with caching)

### Additional documentation

There are four configuration options related to timeouts, which can be confusing. Below is a brief explanation:

- **EventWebhookRequestTimeout**: How long to wait for a response before timing out
- **EventWebhookMaxRetries**: Number of retry attempts
- **EventWebhookBaseWaitInterval**: Base interval for exponential backoff
- **EventWebhookMaxWaitInterval**: Maximum interval between retries

image

```
N/A
```

### Checklist

- [ ]  Added relevant tests or not required
- [ ]  Addressed and resolved all CodeRabbit review comments
- [x]  Ensured no breaking changes

## Summary by CodeRabbit

Based on the comprehensive changes, here are the release notes:

## Release Notes

- **New Features**
	- Added support for project event webhooks
	- Introduced event webhook configuration for document creation and removal events
	- Implemented HMAC-based webhook authentication mechanism

- **Improvements**
	- Enhanced project configuration with event webhook URL and events
	- Added retry and backoff strategies for webhook requests
	- Improved error handling for webhook interactions

- **Configuration**
	- New configuration options for event webhook max retries, wait intervals, and timeouts
	- Ability to specify webhook events and URL at the project level

- **Documentation**
	- Updated OpenAPI specifications to include new event webhook fields
	- Refined documentation for project-level configurations

These release notes summarize the key additions and improvements related to the new event webhook functionality across the project.


hackerwins:
Thanks for your contribution.

I left a few things to think about this PR.

## Document Indexing and Webhooks
- Webhooks need to be introduced for document indexing
- While CodePair can control document creation and deletion
- Document editing currently bypasses the CodePair server
- Therefore, `DocChanged` is essential for document indexing requirements

## Regarding DocEvent
- This PR introduces new events(`ProjectEvent`)
- It would be preferable to have events managed in a unified format within the project
- DocEvent alone cannot distinguish CRUD operations, and ChangePack needs to be interpreted together

## Webhook and DocEvent Integration
- Initially, it seems better to trigger webhooks for DocChanged events
- Current DocEvent structure:
  - types: DocumentChanged, DocumentWatched, DocumentUnwatched, DocumentBroadcast
  - metadata: type, publisher, dockey
  - body: (for DocumentBroadcast)
    - topic: documentkey
    - payload: includes message for broadcasting

## Limitations of DocumentChanged for Content Changes
- Cannot properly represent document content changes due to presence change inclusion
- The current `DocEvent.DocumentChanged` has a limitation in distinguishing between presence and content changes
- Proposed solution: Split into DocumentRootChanged and DocumentPresenceChanged

This approach would provide clearer event separation and better align with the document indexing requirements. Instead of introducing a new ProjectEvent, enhancing the existing DocEvent structure would maintain consistency while addressing the core needs.

For now, it seems better to close this PR, divide it into several pieces, and proceed again.

PR #1156, opened 2025-02-17
Add Event Webhook Support with DocRootChanged Event

window9u:
### **What this PR Does and Why We Need It**

This PR introduces a basic implementation of event webhooks. The key changes include:

1. **Adding Event Webhook Attributes to various types**
    - Introduced `EventWebhookURL` and `EventWebhookTypes` attributes.
	    - Update `project_info` about database.
	    - Update `updatable_field` for yorkie admin service
	    - Updated CLI to support `event-webhook` updates.
2. **Introducing the `DocRootChanged` Event for Event Webhooks**
    - Currently, _codepair_ needs to detect when a document is updated, not just when it is viewed. To address this, I added the `DocRootChanged` event.
    - I considered using `DocPresenceChanged` or `DocChanged` for this purpose, but their usage is ambiguous at the moment, so I did not include them.

#### Request
```json
{
  "type": "string",  // Only 'DocumentRootChanged' for now
  "attributes": {
    "key": "string",  // Document Key
    "issuedAt": "2025-02-17T08:35:27.558Z" //ISO 8601
  }
}
```
#### Response

- HTTP **200 OK**
- No response body

---

### **Which Issue(s) This PR Fixes**

This PR addresses issue: **#1002**

---

#### 1. Caching Concerns
    - This implementation follows the structure of `auth-webhook`.
    - Throttling tests are included but currently skipped.
    - A follow-up PR will address proper webhook caching.

#### 2. Event Webhook Configuration
- I have added six configuration parameters, with default values set the same as `auth-webhook`:
    1. `EventWebhookMaxRetries`
    2. `EventWebhookMaxWaitInterval`
    3. `EventWebhookMinWaitInterval`
    4. `EventWebhookRequestTimeout`
- Configuration details need further discussion.

### After Discussing Caching with Reviewer
- The reviewer and I agreed that removing the cache at this stage is beneficial for clarity and future enhancements. 
- This approach allows us to focus on core functionality now while paving the way for implementing robust throttling and rate limiting in upcoming PRs.

---

### **Does This PR Introduce a User-Facing Change?**

```release-note
Users can receive real-time updates when a document is modified.
```

---

### **Additional Documentation**

```docs

```

---

### **Checklist**

- [x]  Added relevant tests (or not required)
- [x]  Addressed and resolved all CodeRabbit review comments
- [x]  Verified that nothing is broken

## Summary by CodeRabbit

- **New Features**
  - Introduced event webhook support to provide real-time notifications for project events.
  - Enhanced project settings with new options to configure a webhook URL and event triggers.
  - Updated command-line tools and server configurations for easier management of event webhooks.

- **Documentation**
  - Revised API documentation and configuration samples to include the new event webhook parameters.


server/rpc/webhook/event.go
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package webhook provides publishing events to project endpoint.
+package webhook
```
window9u, line 18: "I've thought about the name for this package. Initially, I considered naming it `event`. However, there are three components involved in sending an `event`: `client`, `kafka`, and `webhook`. Naming the package `event` would be misleading since it specifically relates to `webhook`.

Therefore, I named this package `webhook`. However, if this name is still unclear, consider alternative names such as:
- `eventwebhook`
- `event`"

api/types/events/events.go
```diff
@@ -30,6 +30,10 @@ const (
 	// modified by a change.
 	DocChangedEvent DocEventType = "document-changed"
 
+	// DocRootChangedEvent is an event indicating that document's root content
+	// is being changed by operation.
+	DocRootChangedEvent DocEventType = "document-root-changed"
```
window9u, line 35: I considered adding `DocPresenceChangedEvent` to complement `DocRootChangedEvent` for symmetry. However, I decided not to add it for now since its usage is unclear.  

This isn't a critical issue, just an idea. I think it could be useful for sending events to a batch component that removes `Change` entries that only contain presence updates.

server/rpc/webhook/event.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil
+	}
+
+	body, err := buildEventWebhookBody(docKey, webhookType)
+	if err != nil {
+		return fmt.Errorf("marshal event webhook request: %w", err)
+	}
+
+	cacheKey := generateEventCacheKey(prj.PublicKey, docKey, string(webhookType))
+	if _, found := be.EventWebhookCache.Get(cacheKey); found {
```
window9u, line 54: Currently, this caching logic is the same as `auth-webhook`, but it has an issue with throttling:

1. **Get Cache**
2. **Send Webhook**
3. **Set Cache**

With this implementation, if step **2 (Send Webhook)** becomes a bottleneck, the event webhook is not throttled properly.

For example, if the webhook latency is **1s** and an `event` occurs every **0.3s**, the system would send **3 events** instead of throttling them.

As suggested by CodeRabbit, we need to modify the process as follows:

- **1. Get Cache** → **Get or Set Cache**
- **2. Send Webhook**
- **3. Set Cache** → **None**

I believe the key difference between `auth-webhook` and `event-webhook` is this:

- **Auth-webhook** requires a response from the authentication endpoint to validate requests.
- **Event-webhook**, on the other hand, does not need to reuse the `200 OK` response. But proper throttling is more important.

server/packs/packs.go
```diff
@@ -195,6 +196,19 @@ func PushPull(
 				},
 			)
 
+			if reqPack.OperationsLen() > 0 {
+				if err := webhook.SendEvent(
```
window9u, line 200: This operation could take up to (`RequestTimeout` + `MaxWaitInterval`) * `Retries`. With the default settings, this amounts to (3s + 3s) * 10 = 60s.

As a result, it may impact the timing of snapshot creation. This scenario occurs when webhook endpoints crash or take longer than the request timeout (3s) to respond.

server/config.go
```diff
@@ -70,8 +70,16 @@ const (
 	DefaultAuthWebhookMinWaitInterval = 100 * time.Millisecond
 	DefaultAuthWebhookCacheSize       = 5000
 	DefaultAuthWebhookCacheTTL        = 10 * time.Second
-	DefaultProjectCacheSize           = 256
-	DefaultProjectCacheTTL            = 10 * time.Minute
+
+	DefaultEventWebhookRequestTimeout  = 3 * time.Second
```
window9u, line 74: Currently, I have set the default timeout and cache configuration to match `auth-webhook`. We should discuss whether this is the best approach.

window9u:
There are too many changed files, making it difficult to identify the important parts. So, I've left a self-review detailing my considerations and concerns.

test/integration/event_webhook_test.go
```diff
@@ -0,0 +1,303 @@
...
+	})
+	assert.NoError(t, err)
+
+	cli := newActivatedClient(t, ctx, svr.RPCAddr(), project.PublicKey)
+	assert.NoError(t, cli.Attach(ctx, doc, client.WithInitialRoot(map[string]any{
+		"counter": json.NewCounter(0, crdt.LongCnt),
+	})))
+
+	waitWebhookReceived := 20 * time.Millisecond
+	t.Run("throttling event test", func(t *testing.T) {
```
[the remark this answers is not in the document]
hackerwins: Would it be a good idea to cache webhook results for event trigger controlling?

- When an event occurs, it is basic to convey everything to the user.
- However, if too many calls are made, it may be unnecessary.
- This seems to be intended to debouncing or throttle event control.
- We can achieve a similar effect by caching the webhook results, but it makes the intent appear ambiguous.

server/backend/config.go
```diff
@@ -197,6 +232,39 @@ func (c *Config) ParseAuthWebhookCacheTTL() time.Duration {
...
+// ParseEventWebhookRequestTimeout returns request timeout.
+func (c *Config) ParseEventWebhookRequestTimeout() time.Duration {
+	result, err := time.ParseDuration(c.EventWebhookRequestTimeout)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "parse event webhook request timeout: %w", err)
+		os.Exit(1)
+	}
+
+	return result
+}
```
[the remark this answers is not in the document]
window9u: This is fine because we execute all the parsing duration flags before server runs.

cmd/yorkie/server.go
```diff
@@ -74,6 +80,11 @@ func newServerCmd() *cobra.Command {
 			conf.Backend.AuthWebhookMinWaitInterval = authWebhookMinWaitInterval.String()
 			conf.Backend.AuthWebhookRequestTimeout = authWebhookRequestTimeout.String()
 			conf.Backend.AuthWebhookCacheTTL = authWebhookCacheTTL.String()
+
+			conf.Backend.EventWebhookMaxWaitInterval = eventWebhookMaxWaitInterval.String()
+			conf.Backend.EventWebhookMinWaitInterval = eventWebhookMinWaitInterval.String()
+			conf.Backend.EventWebhookRequestTimeout = eventWebhookRequestTimeout.String()
+
```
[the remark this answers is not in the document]
window9u: we decided to remove event webhook cache in this pr

cmd/yorkie/server.go
```diff
@@ -348,6 +359,30 @@ func init() {
...
+		"event-webhook-min-wait-interval",
+		server.DefaultEventWebhookMinWaitInterval,
+		"Minimum wait interval between retries(exponential backoff).",
+	)
+	cmd.Flags().DurationVar(
+		&eventWebhookMaxWaitInterval,
+		"event-webhook-max-wait-interval",
+		server.DefaultEventWebhookMaxWaitInterval,
+		"Maximum wait interval between retries(exponential backoff).",
+	)
```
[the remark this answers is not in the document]
window9u: we decided to remove event webhook cache in this pr

server/packs/packs.go
window9u: I believe this issue can be resolved natively by implementing a `rate-limiter`.

1. Since sending webhooks does not hold a lock, it does not directly block snapshot creation. However, it can still affect the timing of snapshot generation.
2. By applying a `rate-limit`, webhooks may be throttled(skiped). This ensures that even if one event delays snapshot creation, other events can still proceed with triggering snapshots.

hackerwins (approved):

PR #1278, opened 2025-05-21
Filter detached client from vv

### LINKED ISSUE #1144: Remove Detached Client's Lamport from Version Vectors
## Summary
We investigated the possibility of removing Lamport clock information of detached clients from Version Vectors (VV) to optimize VV size. Currently, VV grows indefinitely as more clients participate in document editing. While removing detached client information could reduce VV size, we concluded it's better to maintain this information for now due to several technical challenges, particularly in concurrent operation handling.

## Background
- VV was introduced in v0.5.3 to solve GC (Garbage Collection) problems that couldn't be resolved with Lamport clock alone
- Lamport clock provides total ordering but cannot determine concurrency
  - If `a->b` then `L(a)<L(b)`
  - However, if `L(a)<L(b)`, it could mean either `a->b` or `a||b` (concurrent)
- VV contains editing information of all clients participating in the document
- Current implementation maintains detached client information in VV to ensure proper GC and concurrent operation handling

### 1. Introduction of Version Vector (v0.5.3)
- Purpose: Solve GC problems by enabling concurrency detection (https://github.com/yorkie-team/yorkie-js-sdk/pull/899)
- Key changes:
  - Introduced VV to overcome Lamport clock limitations (#723)
  - Changed GC condition to: `removedAt.lamport <= minVersionVector[removedAt.actor]`
  - Image

#### Initial Attempt to Remove Detached Client's VV 
- Implemented in:
  - Removed detached client data from DB.versionvectors
  - Added filtering of detached clients from minVV
  - Updated local client VV during applyChangePack
    - Image
- Challenges identified:
  - N+1 query problem when removing from other clients' VV
  - Solution: Filter after calculating minVV instead of modifying all VVs
  - Image

### 2. GC Enhancement for Detached Client Nodes (v0.5.6)
- #1060 
- Implementation:
  - Added logic to handle missing Lamport in minVV
  - Used `int64max` when minVersionVector.minLamport() has no value
  - Added condition: `removedAt.lamport < minVersionVector.minLamport()`

| Image|Image| 
| :--: | :--: | 
| Nodes deleted by detached clients are not GC'd because they have no value in minVV | When there is no value in minVV, int64max is used to enable GC | 

### 3. Bug Discovery 
- Issue identified: #1089
- Problem: Premature GC of nodes deleted by detached clients
- Key findings:
  - GC should only occur when all clients are aware of node deletion
  - Current implementation filters minVV immediately upon client detachment
  - This leads to premature filtering before other clients receive the detach change

### 4. Final Implementation (v0.5.7)
Current process:
- Only remove from ColVersionVectors during detach
- Keep detached client information in minVV
- Allow GC for detached client nodes while maintaining VV size

### 1. Conditional Removal:
   - Store detached user data separately with `{actorID:lamport}` format (detachedClientLamport)
   - Remove from minVV only when: `minVV[detachedClientID] === detachedClientLamport`
   - This ensures all users have received the detach change

### 2. Limitation: Concurrent Operation Detection

Main challenge: Cannot reliably determine if missing VV information means:
- Client is detached
- Client Op hasn't been received yet

Image

Example scenario:
```
Client A's delete Op VV: [7@A, 4@B]
Client B state: includes nodes 8@C, 9@C
Problem: Cannot determine if C's absence in VV is due to detachment or lack of information
```

### 3. Future Considerations:   
- DAG (Directed Acyclic Graph) structure could help by:
  - Tracking operation parent information
  - Verifying if client information was previously received

Image

## Decision
We decided to maintain detached client information in VV for now because:
1. Concurrent operation handling requires complete version history
2. Current architecture lacks reliable way to distinguish between detached and unknown clients
3. Removing information could lead to incorrect operation application

## Future Work
1. Implement performance benchmarks to measure VV size impact
2. Consider DAG implementation for better operation history tracking
3. Compare with other editors/libraries (e.g., [Zed editor's approach](https://zed.dev/blog/crdts#deletion))
4. Monitor and analyze real-world performance impacts

JOOHOJANG:
Since version vector's size grow infinitely, it causes performance and memory issues.

We used the min version vector to remove the Lamport of detached clients.

A separate database table was created to store detached clients. When a user detaches, their information is added to this table. Later, during a push-pull operation, we compare the computed min version vector with the list of detached clients. If `minVV[client] == detached client's Lamport`, we remove the corresponding key from the min version vector. This ensures that when a user receives a sync pack, their version vector is filtered accordingly.

Conceptually, removing a user's Lamport from the min version vector implies that all users have acknowledged the user's detachment. This condition can be confirmed when `minVV[detached client ID] == detached client’s Lamport`

Although introducing a separate table may raise concerns about potential performance overhead, the impact is expected to be minimal.

**Changed**

- [x] Add new database table (detached clients)
- [x] Filter min version vector using detached clients table(using cache to improve performance)
- [x] Filter every client's local version vector when receiving change pack
- [x] Add related TCs

image

### Next tasks
- [ ] Update memory db in same ways
- [ ] Implement local version vector update logic into SDKs
- [ ] Testing in many ways 

Fixes https://github.com/yorkie-team/yorkie/issues/1144

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

- **New Features**
  - Introduced tracking and management of detached clients to improve document synchronization and garbage collection.
- **Bug Fixes**
  - Enhanced version vector handling to exclude detached clients, ensuring accurate document state and cleanup.
- **Tests**
  - Added integration tests to verify correct garbage collection behavior after client detachment.
- **Chores**
  - Updated database schema and indexing to support detached client tracking.
  - Extended document editing and styling operations with enhanced version vector parameters for improved consistency.

### DIFF
--- a/admin/client.go
+++ b/admin/client.go
@@ -402,7 +402,7 @@ func (c *Client) ListChangeSummaries(
 	}
 	var summaries []*types.ChangeSummary
 	for _, c := range changes {
-		if _, err := newDoc.ApplyChanges(c); err != nil {
+		if _, err := newDoc.ApplyChanges(nil, c); err != nil {
 			return nil, err
 		}
 
--- a/pkg/document/change/change.go
+++ b/pkg/document/change/change.go
@@ -51,9 +51,9 @@ func New(id ID, message string, operations []operations.Operation, pc *innerpres
 }
 
 // Execute applies this change to the given JSON root.
-func (c *Change) Execute(root *crdt.Root, presences *innerpresence.Map) error {
+func (c *Change) Execute(root *crdt.Root, presences *innerpresence.Map, vector time.VersionVector) error {
 	for _, op := range c.operations {
-		if err := op.Execute(root, c.ID().versionVector); err != nil {
+		if err := op.Execute(root, c.ID().versionVector, vector); err != nil {
 			return err
 		}
 	}
--- a/pkg/document/crdt/rga_tree_split.go
+++ b/pkg/document/crdt/rga_tree_split.go
@@ -554,7 +554,7 @@ func (s *RGATreeSplit[V]) deleteNodes(
 			if ok {
 				clientLamportAtChange = lamport
 			} else {
-				clientLamportAtChange = 0
+				clientLamportAtChange = vector.MaxLamport()
 			}
 		}
 
--- a/pkg/document/crdt/rga_tree_split.go
+++ b/pkg/document/crdt/rga_tree_split.go
@@ -471,6 +471,7 @@ func (s *RGATreeSplit[V]) edit(
 	content V,
 	editedAt *time.Ticket,
 	versionVector time.VersionVector,
+	minVersionVector time.VersionVector,
 ) (*RGATreeSplitNodePos, []GCPair, error) {
 	// 01. Split nodes with from and to
 	toLeft, toRight, err := s.findNodeWithSplit(to, editedAt)
@@ -484,7 +485,7 @@ func (s *RGATreeSplit[V]) edit(
 
 	// 02. delete between from and to
 	nodesToDelete := s.findBetween(fromRight, toRight)
-	removedNodes := s.deleteNodes(nodesToDelete, editedAt, versionVector)
+	removedNodes := s.deleteNodes(nodesToDelete, editedAt, versionVector, minVersionVector)
 
 	var caretID *RGATreeSplitNodeID
 	if toRight == nil {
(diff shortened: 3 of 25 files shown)


hackerwins:
@JOOHOJANG Thanks for your contribution.

VersionVector is used for both `causality tracking` and `garbage collection`. We want to remove Lamport of a detached Actor, but we first need to clearly define a safe removal point that does not lose essential causality information.

I'll close this PR for now until we have a clearer definition of that safe point.

PR #1296, opened 2025-06-01
Add project API key rotation to enhance security

### LINKED ISSUE #1020: Provide API Key Rotation

Currently, the `public_key` and `private_key` are immutable once set. However, in cases such as key exposure or security concerns, there should be a way to rotate or update the API keys. 

Additionally, the current key generation method uses MongoDB `xid`, which should be updated to a more secure random key generation logic. This is addressed in [PR#1010](https://github.com/yorkie-team/yorkie/pull/1010).

**Why is this needed**:

Implementing this enhancement would improve security and make key management easier for administrators, allowing them to rotate keys as needed in response to security events.

minwoo1999:
Add project API key rotation functionality to enhance security by allowing periodic key updates. This feature enables project owners to rotate their API keys when needed, following security best practices.

Fixes #1020 

- Added comprehensive test cases covering success and failure scenarios
- Implemented proper authentication checks for project ownership
- Used secure key generation methods
- Added proper error handling for various edge cases

```release-note
Add project API key rotation functionality
- Added new endpoint `/yorkie.v1.AdminService/RotateProjectKeys`
- Project owners can now rotate their project's API keys
- Proper authentication and authorization checks implemented
```

**Additional documentation**:

```docs
- [API Documentation]: api/docs/yorkie/v1/admin.openapi.yaml
- [Implementation Details]: server/projects/projects.go
```
**TestCase Image**:
image

**Checklist**:
- [x] Didn't break anything

## Summary by CodeRabbit

- **New Features**
  - Added the ability to rotate API keys for projects via a new endpoint in the Admin Service.
- **Documentation**
  - Updated API documentation to include the new project key rotation endpoint and clarified usage notes for certain legacy types.
- **Tests**
  - Introduced tests to verify key rotation functionality and error handling for various scenarios.


minwoo1999:
> Thank you for your contribution.
> 
> Overall, it looks good. And in my personal opinion, I think we should invalidate the project cache in the RPC server's interceptor to ensure consistency with the updated database.
> 
> As you can see, we cache the project, and whenever a request comes in, the RPC server's interceptor builds the context using the cached project based on the API key from the request.
> 
> https://github.com/yorkie-team/yorkie/blob/4c65d115dbf8b07bf1b6c6bed969b25189282c4a/server/rpc/interceptors/yorkie.go#L170-L179
> 
> So, when an API key is rotated, the RPC server may end up with two cache entries—each with a different API key but referring to the same (or nearly the same) project. This could be a security issue if the old key is compromised. Even after rotating the key, an attacker with access to the old key might still be able to access resources.
> 
> Therefore, I believe we need project cache invalidation. It would also be beneficial to add test cases to cover scenarios where a rotated key could cause issues.

Thanks again for the helpful feedback!

As you mentioned, there was indeed a chance that rotating an API key could leave behind stale or duplicated cache entries, which might cause inconsistencies or even security issues. To address this, I updated the interceptor logic to invalidate the old cache entry when the API key is rotated. This way, we can make sure that only the latest key is used and prevent any unintended access with outdated credentials.

I’ve also added test cases to cover this scenario — specifically checking that access is properly denied when an old key is used after rotation, and that the context is rebuilt correctly with the new key.

Thanks again for raising this point!

hackerwins:
@minwoo1999 Thanks for your contribution.

Invalidating the cache in the interceptor to avoid stale projects in single node.

But, I’d like to note that in cluster mode, each node maintains its own in-memory project cache. So invalidating the cache on a single node may not be sufficient to ensure consistency across the entire cluster.

https://github.com/yorkie-team/yorkie/blob/main/design/sharded-cluster-mode.md

At the moment, we don’t have a mechanism to broadcast across nodes. I think it would be good to leave a TODO for this for invalidation in the future.

hackerwins (approved):

PR #1330, opened 2025-06-18
Change MongoDB project wide shard key (clients, documents)

emplam27:

Before:

Now `clients` and `documents` collections are sharded by `project_id`. But users typically create a project, and then create clients and documents in that project. So `clients` and `documents` queries are tend to be crowded onto one shard due to the project_id ranged sharding policy.

image

After:

Update MongoDB shard key project-wide sharded collections(`clients` and `documents`) from `project_id` ranged to `key` hasded collections

- `clients` collection have `key` field. it means string value set by the SDKs when activating the clients.
- `documents` collection have `key` field. it means string value set by the SDKs when attaching the documents.

Queries be distributed equally across both shards with key. However, If a user intentionally creates a key that can only be queried by one shard, it can cause performance issues.

image

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:

- [ ] Didn't break anything


hackerwins:
This PR continues in #1371.

PR #1425, opened 2025-08-02
Prevent project secret key from granting full user access

### LINKED ISSUE #813: Project Secret Key grants the same permissions as the owner of the project.
**What happened**:

The Project Secret Key grants the same permissions as the owner of the project. For example, when Admin A owns Project B and uses the Secret Key of that project, they can access other projects owned by Admin A as well. This occurs because accessing a project via the Project Secret Key returns the user who owns the project.

image

The Project Secret Key should only provide access to the specific project it belongs to and not grant access to other projects owned by the same user.

The issue can be reproduced by using the Project Secret Key to access a project and observing that it inadvertently provides access to other projects owned by the same user.

It is suggested that a solution could involve adding a flag to the User to distinguish between Admin and Project access rights.

```go
claims, err := i.tokenManager.Verify(authorization)
if err == nil {
    user, err := users.GetUserByName(ctx, i.backend, claims.Username)
    if err == nil {
        return user, nil
    }
}

project, err := projects.GetProjectFromSecretKey(ctx, i.backend, authorization)
if err == nil {
    user, err := users.GetUserByID(ctx, i.backend, project.Owner)
    if err == nil {
        return user, nil // Need to return Project permissions, not Admin permissions
    }
}
``` 

**Environment**:
- Operating system: ALL
- Browser and version: ALL
- Yorkie version (use `yorkie version`): v0.4.14
- Yorkie JS SDK version: NONE

SANGHEEJEONG:

<h3>Issue</h3>
<ul>
<li><strong>Problem</strong>: When Admin A (owner of Project B) uses the secret key of Project B, they are unintentionally granted access to other projects they own.</li>
<li><strong>Cause</strong>: The current logic returns the project owner as a user, which implicitly grants full access across all projects owned by that user.</li>
</ul>
<pre><code class="language-go">project, err := projects.GetProjectFromSecretKey(ctx, i.backend, authorization)
if err == nil {
    user, err := users.GetUserByID(ctx, i.backend, project.Owner)
    if err == nil {
        return user, nil // This returns full user-level access instead of project-scoped access
    }
}

</code></pre>

<br>

<h3>Ideal Behavior</h3>

Item | Current (As-Is) | Desired (To-Be)
-- | -- | --
Project X's secret key | Grants access to all of Developer A’s projects | Grants access to Project X only
Developer A’s user token | Grants access to all of their own projects | Unchanged — full access to all own projects
---
Fixes #813 

---

### Summary of Changes

To resolve this, I introduce the concept of access scope in the `User` model.

- **User Token** → Returns `User` with `AccessScopeAdmin`
- **Project Key** → Returns `User` with `AccessScopeProject` and sets the `Project` in context

This allows downstream handlers to determine the access level and use it appropriately.
```go
// If access scope is project, verify the project from context
if user.AccessScope == types.AccessScopeProject {
	ctxProject := projects.From(ctx)
	if ctxProject.Name != projectName {
		return nil, fmt.Errorf("project mismatch: key is for %s, got %s", ctxProject.Name, projectName)
	}
	return ctxProject, nil
}

// For admin scope, fetch the project
project, err := projects.GetProject(ctx, s.backend, user.ID, projectName)
if err != nil {
	return nil, err
}
```
---
<br>

```release-note
NONE

```

**Additional documentation**:

```docs

```

**Checklist**:

- [ ] Didn't break anything

## Summary by CodeRabbit

* **New Features**
  * User accounts now include an access scope, distinguishing between full administrative access and project-specific access.

* **Refactor**
  * Improved authorization logic for project access, ensuring consistent enforcement of user access scopes.
  * Authentication process now enriches request context with user and project information for streamlined access control.

* **Tests**
  * Added integration test to verify that document access is restricted across different projects.
  * Introduced helper functions to simplify project creation and unauthorized access testing in tests.

### DIFF
--- a/server/rpc/admin_server.go
+++ b/server/rpc/admin_server.go
@@ -836,7 +836,12 @@ func (s *adminServer) GetProjectWithAuth(
 		// Verify project from context for project scope
 		ctxProject := projects.From(ctx)
 		if ctxProject.Name != projectName {
-			return nil, fmt.Errorf("project mismatch: key is for %s, got %s: %w", ctxProject.Name, projectName, auth.ErrUnauthenticated)
+			return nil, fmt.Errorf(
+				"project mismatch: key is for %s, got %s: %w",
+				ctxProject.Name,
+				projectName,
+				auth.ErrUnauthenticated,
+			)
 		}
 		return ctxProject, nil
 	case types.AccessScopeAdmin:
--- a/server/rpc/admin_server.go
+++ b/server/rpc/admin_server.go
@@ -836,7 +836,7 @@ func (s *adminServer) GetProjectWithAuth(
 		// Verify project from context for project scope
 		ctxProject := projects.From(ctx)
 		if ctxProject.Name != projectName {
-			return nil, fmt.Errorf("project mismatch: key is for %s, got %s", ctxProject.Name, projectName)
+			return nil, fmt.Errorf("project mismatch: key is for %s, got %s: %w", ctxProject.Name, projectName, interceptors.ErrUnauthenticated)
 		}
 		return ctxProject, nil
 	case types.AccessScopeAdmin:
--- a/server/rpc/admin_server.go
+++ b/server/rpc/admin_server.go
@@ -831,21 +831,22 @@ func (s *adminServer) GetProjectWithAuth(
 ) (*types.Project, error) {
 	user := users.From(ctx)
 
-	// If access scope is project, verify the project from context
-	if user.AccessScope == types.AccessScopeProject {
+	switch user.AccessScope {
+	case types.AccessScopeProject:
+		// Verify project from context for project scope
 		ctxProject := projects.From(ctx)
 		if ctxProject.Name != projectName {
 			return nil, fmt.Errorf("project mismatch: key is for %s, got %s", ctxProject.Name, projectName)
 		}
-
 		return ctxProject, nil
+	case types.AccessScopeAdmin:
+		// Fetch project for admin scope
+		project, err := projects.GetProject(ctx, s.backend, user.ID, projectName)
+		if err != nil {
+			return nil, err
+		}
+		return project, nil
+	default:
+		return nil, fmt.Errorf("invalid access scope: %s", user.AccessScope)
 	}
-
-	// For admin scope, fetch the project
-	project, err := projects.GetProject(ctx, s.backend, user.ID, projectName)
-	if err != nil {
-		return nil, err
-	}
-
-	return project, nil
 }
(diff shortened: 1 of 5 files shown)


server/rpc/admin_server.go
```diff
@@ -839,3 +823,29 @@ func (s *adminServer) RotateProjectKeys(
...
+
+	// If access scope is project, verify the project from context
+	if user.AccessScope == types.AccessScopeProject {
+		ctxProject := projects.From(ctx)
+		if ctxProject.Name != projectName {
+			return nil, fmt.Errorf("project mismatch: key is for %s, got %s", ctxProject.Name, projectName)
+		}
+
+		return ctxProject, nil
+	}
```
blurfx: I suggest explicitly checking for every `AccessScope` and returning an error if it is not a valid `AccessScope` to prevent granting incorrect permissions from several edge cases, like when new `AccessScope` types are added in the future. If you think it's okay with this change, it would be better to use a `switch` statement instead of `if`.

server/rpc/admin_server.go
```diff
@@ -839,3 +823,29 @@ func (s *adminServer) RotateProjectKeys(
 		Project: converter.ToProject(newProject),
 	}), nil
 }
+
+// GetProjectWithAuth fetches the project based on the user's access scope.
+func (s *adminServer) GetProjectWithAuth(
```
kokodak, line 828: Authorization is a critical aspect, so I believe it's essential to have test coverage to verify that #813 has been truly resolved. Specifically, we should add tests for the `GetProjectWithAuth`.

The tests should cover the following cases:
1. When a request is made using a Project Secret Key, ensure that it cannot access any project other than the one it belongs to.
2. When a request is made using a User Token, ensure that it continues to behave as it did before.

server/rpc/admin_server.go
```diff
@@ -839,3 +823,29 @@ func (s *adminServer) RotateProjectKeys(
...
+	ctx context.Context,
+	projectName string,
+) (*types.Project, error) {
+	user := users.From(ctx)
+
+	// If access scope is project, verify the project from context
+	if user.AccessScope == types.AccessScopeProject {
+		ctxProject := projects.From(ctx)
+		if ctxProject.Name != projectName {
+			return nil, fmt.Errorf("project mismatch: key is for %s, got %s", ctxProject.Name, projectName)
```
kokodak: I think it would be good to return `ErrUnauthenticated` error.
Something like this:
```suggestion
			return nil, fmt.Errorf("project mismatch: key is for %s, got %s: %w", ctxProject.Name, projectName, ErrUnauthenticated)
```

This will allow for verification in this way:
```go
assert.Equal(t, connect.CodeUnauthenticated, connect.CodeOf(err))
```

server/rpc/admin_server.go
SANGHEEJEONG: Thank you for letting me know!

server/rpc/admin_server.go
SANGHEEJEONG: I've confirmed that the second case is already covered by existing tests, so I added a test only for the first case. 
Thanks!

hackerwins (approved):
Thanks for your contribution.

Issue #1352, opened 2024-09-03
Introduce REST API for event broadcasting


devleejb:
**Description**:
The recent addition of the event Broadcast functionality in the `yorkie-js-sdk` (referenced in [this PR](https://github.com/yorkie-team/yorkie-js-sdk/pull/884)) has been a great enhancement. However, currently, this feature is only accessible through the `yorkie-js-sdk`. It would be beneficial to extend this functionality to allow for event broadcasting through a REST API.

For reference, LiveBlocks provides a REST API for broadcasting events, as detailed [here](https://liveblocks.io/docs/api-reference/rest-api-endpoints#post-broadcast-event). I propose implementing a similar REST API endpoint in Yorkie.

**Why**:
Introducing a REST API for event broadcasting will enhance flexibility and enable broader usage scenarios, allowing users who prefer RESTful interactions to leverage the event functionality without being limited to the SDK. This would cater to a wider audience and potentially increase adoption of the feature.

jojaeng2:
Hi, I'm interested in this issue. can I try this?

kokodak:
Hi @jojaeng2! Of course you can. If you have any question, feel free to ask via Discord or GitHub.

jojaeng2:
Thanks, @kokodak,
Before diving into how to solve this issue, I have a few questions.

First, I want to fully understand the reason why this issue is necessary. 
After reviewing several related issues, here’s my current understanding of the situation Yorkie is facing:

1. Yorkie already has a broadcast API implemented, as introduced in this [PR](https://github.com/yorkie-team/yorkie/pull/631).

2. The necessary code to use this API is already implemented in both the JS and Go SDKs.

3. If a developer does not use the JS or Go SDK, they would need to access the API via gRPC, which is not very user-friendly.

4. Supporting event broadcasts through a developer-friendly REST API could lower the entry barrier and potentially attract more users.

This is my current understanding. Is that correct?
Since I’m still getting familiar with the Yorkie project and not yet experienced with Go and gRPC, I might have misunderstood some parts.
If anything I commented on is incorrect, please let me know.

kokodak:
@jojaeng2 
It seems you’ve understood the situation very well. As you mentioned, the Broadcast API is implemented in both the JS SDK and Go SDK, but to improve accessibility, support for a REST API is also necessary.

[...]

PR #1449, opened 2025-08-16
Introduce REST API for event broadcasting

jojaeng2:
- Added a REST API for publishing broadcast events.

Fixes # https://github.com/yorkie-team/yorkie/issues/1352

To properly test the feature I added, it would require implementing a subscription and verifying that the subscriber receives the DocBroadcast correctly. However, since it’s tricky to perform that verification through the REST API approach, I simply added a test case that checks whether the API call returns successfully without errors.

Please let me know if there’s a better way to handle this.
Also, since my understanding of Go is still limited, I’d appreciate any comments if you notice something that has been incorrectly implemented.

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:

- [ ] Didn't break anything


test/integration/restapi_test.go
```diff
@@ -197,6 +197,18 @@ func TestRESTAPI(t *testing.T) {
 		assert.Len(t, summaries.Documents, 0)
 	})
 
+	t.Run("broadcast test", func(t *testing.T) {
```
kokodak, line 200: It would be great if the existing test scenarios for restapi_test could be expanded with more concrete steps.
For example, a scenario like the following:
1. Create and activate clients C1, C2, and C3
2. Create a document and attach it to each client
3. Establish WatchDocument stream connections for each client
4. Verify that when C1 publishes an event, it is broadcasted to C2 and C3

You can refer to the following test as a helpful example:
https://github.com/yorkie-team/yorkie/blob/813071407b46f743a40698c5f21d02e17a22a186/server/rpc/testcases/testcases.go#L599-L649

Also, it would be nice to add an e2e test for broadcasting over RPC. Currently, there is no E2E test for this in the RPC layer.
Such a test could be defined in `server/rpc/testcases/testcases.go`.

kokodak (commented):
Thanks for your contribution. I've added some comments.

hackerwins:
@kokodak @jojaeng2 Thanks for your contribution.

Regarding [Yorkie cluster architecture](https://github.com/yorkie-team/yorkie/blob/main/design/sharded-cluster-mode.md#how-does-it-work), I think we need to clarify a few points:
- The cluster uses Consistent Hashing so that each Document is handled by a specific server.
- Locks for document editing and Subscriptions are not distributed across the cluster but are instead managed locally on each server.
- For this reason, when admin.Client sends a request to admin.Server, it specifies the document shard key to ensure the request is routed to the correct server.

The issue is that the Admin REST API does not specify the shard key in its requests, which may cause incorrect behavior.
To make it work properly, admin.Server should forward REST API requests to the appropriate shard server via the internal ClusterServer connection.

hackerwins:
Thanks for your contribution.

I’m closing this PR for now as there hasn’t been any recent progress.
We can reopen it later if necessary.

Issue #1446, opened 2025-08-12
Increase coverage by adding more tests


ggyuchive:
**What would you like to be added**
 Currently, Yorkie coverage is [38.31%](https://app.codecov.io/gh/yorkie-team/yorkie). Some lines such as `Client.Attach`, remain uncovered even though the tests execute `Client.Attach` multiple times. But it seems some functionality is missing in testscript.

**Why is this needed**
 Test enhancement can assure backward compatibility, no performance regression in case of new feature development or refactoring.

**Good to add**
1. Add database interface unit-test : Cover `server/backend/database/mongo/client.go`, `server/backend/database/memory/database.go`
2. Add config test : Cover `server/backend/**/config.go` (https://github.com/yorkie-team/yorkie/pull/1450)
3. Add cmd test : Cover under `cmd/yorkie`
4. Add packs unit-test : Cover under `server/packs`
5. Add rpc test: Cover `server/rpc/yorkie_server.go`, `server/rpc/admin_server.go`, `server/rpc/cluster_server.go`
6. Add test cases which expects error (e.g. Return proper log in wrong parameter value, database error, ...)
7. Introduce new test package and tests that includes scenario with bug reported in issue, remove in ci and makefile
8. ...

**How to measure coverage enhancement in local**
1. Run `go test -tags integration -race -coverprofile=coverage_main.out -covermode=atomic ./...` in latest yorkie change
2. Add test - I newly added AdminRPC CreateSchema test with below.
- `./server/rpc/testcases/testcases.go`
```go
func RunAdminCreateSchemaTest(
	t *testing.T,
	testClient v1connect.YorkieServiceClient,
	testAdminClient v1connect.AdminServiceClient,
	testAdminAuthInterceptor *admin.AuthInterceptor,
) {
	resp, err := testAdminClient.LogIn(
		context.Background(),
		connect.NewRequest(&api.LogInRequest{
			Username: helper.AdminUser,
			Password: helper.AdminPassword,
		},
		))
	assert.NoError(t, err)

	testAdminAuthInterceptor.SetToken(resp.Msg.Token)

	_, err = testClient.ActivateClient(
		context.Background(),
		connect.NewRequest(&api.ActivateClientRequest{ClientKey: t.Name()}))
	assert.NoError(t, err)

	schemaRule1 := []types.Rule{
		{
			Path: "$.title",
			Type: "string",
		},
	}
	rule1 := converter.ToRules(schemaRule1)

[...]

PR #1450, opened 2025-08-16
Update default value while configure server with yml file

ggyuchive:
- Update `ensureDefaultValue` and `config.sample.yml` to updated config
- Delete unused configuration variable
- Add tests related to configuration methods

Subtask of #1446 

Fixes #

NONE

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:

- [X] Didn't break anything

### DIFF
--- a/cmd/yorkie/server.go
+++ b/cmd/yorkie/server.go
@@ -70,7 +70,6 @@ var (
 	eventWebhookMaxWaitInterval time.Duration
 	eventWebhookMinWaitInterval time.Duration
 	eventWebhookRequestTimeout  time.Duration
-	eventWebhookCacheTTL        time.Duration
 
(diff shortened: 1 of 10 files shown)


ggyuchive:
Hi team, while reading yml file, boolean value with `conf.Backend.UseDefaultProject` is hard to initialize because `server.DefaultUseDefaultProject = true`.

We are now set default value when `int = 0`, `string = ""`. However, it's hard to know boolean type set false explicitly or just initialized. For resolving this, change from `bool` to `UseDefaultProject *bool yaml:"UseDefaultProject"` can be a solution.

Is it okay to change like this? I'm not changed this yet because I thought confirm this is needed.
 Thanks in advance :)

PR #1477, opened 2025-09-02
Add Dev Container configuration

### LINKED ISSUE #1451: Consider using devcontainer for development environment
[Visual Studio Code Dev Containers](https://code.visualstudio.com/docs/devcontainers/containers) makes it easier to create a development environment inside a container. I think It would be great to have a devcontainer configuration for yorkie.

FYI, GoLand seems like it [supports](https://www.jetbrains.com/help/go/connect-to-devcontainer.html) devcontainers.

**Why is this needed**:

Making a development environment inside a container has these benefits below:

##### 1. Host environment and development environment can be separated.

This is great when we want contributors to develop on fixed or unified environment. Contributors would always develop with same versions of the tools and same system configuration.

For example, currently yorkie uses golangci-lint `v1.64.6` and I already have `v2.1.5` installed. So even after `make tools`, running `golangci-lint run` would use `v2.1.5` and result in below:
```
golangci-lint run ./...
Error: can't load config: unsupported version of the configuration: "" See https://golangci-lint.run/product/migration-guide for migration instructions
Failed executing command with error: can't load config: unsupported version of the configuration: "" See https://golangci-lint.run/product/migration-guide for migration instructions
```
We could avoid situations like this by separating the development environment.

Plus, IMO some people with paranoia wouldn't want to install the tools on their host environment.

##### 2. Contributors can start their contribution with minimum effort.

The requirements for contributors are 1) Docker; 2) IDE that supports devcontainer. 
We could even remove the manual installation procedure of protobuf compiler by adding something like post-start script of the container.

onee-only:
This PR makes it available for contributors of Yorkie to work on development container environment.
Additional information can be found on #1451 .

Fixes #1451

The commit ad48df373a3dde935dcc8ee9cac5be5d22ce223e isn’t directly related to this PR.
I included it since it’s a small change, but I can remove it and open a separate PR if you think that’s more appropriate.

```release-note
NONE
```

**Additional documentation**:

```docs

```

**Checklist**:

- [X] Didn't break anything

## Summary by CodeRabbit

- Chores
  - Added a VS Code Dev Container for a Go-based development environment with protobuf tooling and Docker access, including an automated post-create setup that downloads modules and installs tools.
- Documentation
  - Updated contributing guide: raised required Go version to 1.23+ and added instructions for using Dev Containers via supported IDEs or the CLI.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,12 @@
+{
+    "name": "Yorkie",
+    "image": "mcr.microsoft.com/devcontainers/go:1.24",
+    "features": {
+        "ghcr.io/devcontainers-extra/features/protoc": "latest",
+        "ghcr.io/devcontainers/features/docker-outside-of-docker": "latest"
+    },
+    "postCreateCommand": "sh .devcontainer/setup.sh",
+    "runArgs": ["--network=host"],
+    "forwardPorts": [],
+    "customizations": {}
+}
--- a/.devcontainer/setup.sh
+++ b/.devcontainer/setup.sh
@@ -1,0 +1,4 @@
+#! /bin/sh
+
+go mod download
+make tools
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,6 +37,14 @@ Below are needed for developing and building Yorkie.
 - [Protobuf Compiler](https://github.com/protocolbuffers/protobuf/releases) (version 3.4.0+)
 - [Docker](https://www.docker.com/)
 
+#### Using Dev Container
+Alternatively, you can use [Development Containers](https://containers.dev/). The only requirement is Docker.
+
+To use it, you can:
+
+- Use an IDE that supports Dev Containers (VS Code, GoLand, etc.)
+- Set up manually with the [CLI](https://github.com/devcontainers/cli)
+
 ### Building & Testing
 
 You must install required Go packages to build Yorkie executable.


krapie (commented):

I tried starting the dev containers with this configuration, but it fails when installing `ghcr.io/devcontainers/features/docker-outside-of-docker`, with below error logs:

```bash
ERROR: failed to solve: process "/bin/sh -c cp -ar /tmp/build-features-src/docker-outside-of-docker_1 /tmp/dev-container-features  && chmod -R 0755 /tmp/dev-container-features/docker-outside-of-docker_1  && cd /tmp/dev-container-features/docker-outside-of-docker_1  && chmod +x ./devcontainer-featur
[2025-09-04T13:45:27.906Z] es-install.sh  && ./devcontainer-features-install.sh  && rm -rf /tmp/dev-container-features/docker-outside-of-docker_1" did not complete successfully: exit code: 1
``` 

Removing the `docker-outside-of-docker` feature allowed the dev containers to start successfully. Could you check on your local machine to see if you encounter the same error?

onee-only:
@krapie You are right. Weirdly the build didn't fail at the time I created this PR. I tested it on GitHub Codespaces too. 
But now it fails on both my local environment and codespaces.

We need [`docker-outside-of-docker`](https://github.com/devcontainers/features/tree/main/src/docker-outside-of-docker) feature to use host's docker engine inside of container. So we cannot remove it.

I've read some logs and this is what I got:
```sh
[2025-09-05T01:53:24.587Z] 2.162 E: Unable to locate package moby-cli
2.162 E: Unable to locate package moby-buildx
2.163 (!) It seems packages for moby not available in OS debian trixie (arm64). To resolve, either: (1) set feature option '"moby": false' , or (2) choose a compatible OS version (eg: 'ubuntu-24.04').
2.163 ERROR: Feature "Docker (docker-outside-of-docker)" (ghcr.io/devcontainers/features/docker-outside-of-docker) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/docker-outside-of-docker for help troubleshooting this error.
```

Apparently the "trixie" release doesn't include `moby-cli` and `moby-buildx` in their package list by default. 

I changed it to "bookworm" release (`"mcr.microsoft.com/devcontainers/go:1.24-bookworm"`) and it works again.
Alternatively, specifying `{"moby": false}` on the configuration also works. This will install docker-cli instead of moby.

I'm not sure which approach to choose on this. AFAIK, trixie is the newer release with performance enhancements. So it might be better to use trixie and use docker-cli instead of moby.

onee-only:
+I just found out that there is go:1.25 image for devcontainer. I'll update it as well after we resolve the issue above.

hackerwins:
@onee-only 
Thanks for your contribution. But I would like to close this PR for the following reasons:
https://github.com/yorkie-team/yorkie/issues/1451#issuecomment-3264804461

PR #1539, opened 2025-10-20
Improve diagnostics by including details on Webhook failure

downfa11:

im angry. Improves error messages in webhook client by including response body when JSON decoding fails or unexpected responses are received. 

Changes:
- modified `pkg/webhook/client.go` to read and include response body in error messages when JSON decoding fails 
- updated `server/rpc/auth/webhook.go` to include status code, allowed flag, and reason in unexpected response errors 
- added test case in `pkg/webhook/client_test.go` to verify error messages contain response body content

This enhancement makes debugging webhook integration issues significantly easier by showing the actual response content that caused the error.

<br>

the error now includes information to make it clear why the response was unexpected.

<br>

```release-note
none
```

**Additional documentation**:

```docs

```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

* **Bug Fixes**
  * Webhook responses are now fully read and JSON parse failures surface the raw response and clearer diagnostic context (status, allowed flag, reason) with wrapped errors for easier troubleshooting.
* **Tests**
  * Added a test that verifies handling of invalid JSON webhook responses and that errors include the raw response and status.

### DIFF
--- a/pkg/webhook/client.go
+++ b/pkg/webhook/client.go
@@ -119,7 +119,7 @@ func (c *Client[Req, Res]) Send(
 			if logger != nil {
 				logger.Errorf("webhook json decode failed: status=%d, body=%s", resp.StatusCode, string(bodyBytes))
 			}
-			return resp.StatusCode, fmt.Errorf("%w: invalid json response", ErrUnexpectedResponse)
+			return resp.StatusCode, fmt.Errorf("invalid json response: %w", ErrUnexpectedResponse)
 		}
 
 		return resp.StatusCode, nil


hackerwins:
@downfa11

Thanks for adding detailed error information for webhook debugging!

However, I have a concern about exposing the webhook server’s response body directly to clients. This could unintentionally leak sensitive infrastructure details(e.g. database connection strings, internal paths, etc.).

It might be safer to log this information only in Dashboard for debugging purposes, instead of returning it to clients. We might also consider providing a tab on Dashboard that can show webhook errors.

But I think this will take a long time to work on, How about sending a more granular error now, like "invalid json response"?

downfa11:
@hackerwins 

Thanks to reveiw! I also think returning it all the way to the client is risky. 

Wouldn’t it be better to log it only on the server? It should still make debugging convenient.

hackerwins:
@downfa11 I think it would be good to log it on the server. Could we chat for a moment?
https://discord.gg/MVEAwz9sBy

pkg/webhook/client.go
```diff
@@ -108,8 +109,17 @@ func (c *Client[Req, Res]) Send(
...
+		if err != nil {
+			return resp.StatusCode, fmt.Errorf("read response body: %w", err)
+		}
+
+		if err := json.Unmarshal(bodyBytes, &res); err != nil {
+			logger := logging.From(ctx)
+			if logger != nil {
+				logger.Errorf("webhook json decode failed: status=%d, body=%s", resp.StatusCode, string(bodyBytes))
+			}
+			return resp.StatusCode, fmt.Errorf("%w: invalid json response", ErrUnexpectedResponse)
```
hackerwins:
```suggestion
			return resp.StatusCode, fmt.Errorf("invalid json response: %w", ErrUnexpectedResponse)
```

https://github.com/uber-go/guide/blob/master/style.md#error-wrapping

PR #1586, opened 2025-11-27
Prevent empty API key routing and disable default tenant in cluster

raararaara:

This PR addresses the behavior of the SDK when making API calls without an API Key. 
The current Lua filter uses a default value for the `x-shard-key` header, which results in incorrect routing in a cluster environment. Specifically, if the `x-shard-key` is missing, it defaults to "default," which can lead to requests being routed to an incorrect node, causing operational inconsistencies.

The proposed solution removes the default fallback in cluster mode and returns a 400 error if an API Key or Shard Key is missing. This change ensures that requests are correctly routed to the node associated with the actual document, enhancing the robustness of our system.

#### Any background context you want to provide?

Currently, when an SDK API call is made without an API Key, it triggers a Lua filter that assigns a default `x-shard-key`. This results in various nodes processing the same document due to mismatched handling between SDK requests and [server-based cluster services](https://github.com/yorkie-team/yorkie/pull/1036).

In non-cluster mode, this issue does not occur as requests can be routed appropriately due to the single-server environment.

Fixes #

### Checklist
- [x] Didn't break anything

## Summary by CodeRabbit

* **Chores**
  * Removed automatic insertion of the x-shard-key request header at the cluster ingress.
  * Retained the HTTP connection manager's stream idle timeout; no other ingress routing behavior changed.
  * Updated the backend startup argument to use an inline flag assignment for the backend default-project setting.

### DIFF
--- a/build/charts/yorkie-cluster/templates/yorkie/deployment.yaml
+++ b/build/charts/yorkie-cluster/templates/yorkie/deployment.yaml
@@ -88,7 +88,7 @@ spec:
           "{{ .Values.yorkie.name }}-gateway.{{ .Values.yorkie.namespace }}.svc.cluster.local",
           "--backend-rpc-addr",
           "$(POD_IP):{{ .Values.yorkie.ports.rpcPort }}",
-          "--backend-use-default-project", "{{ .Values.yorkie.args.useDefaultProject }}",
+          "--backend-use-default-project={{ .Values.yorkie.args.useDefaultProject }}",
           {{- if .Values.yorkie.logLevel }}
           "--log-level",
           "{{ .Values.yorkie.logLevel }}",


build/charts/yorkie-cluster/templates/istio/ingress-envoy-filter.yaml
```diff
@@ -44,20 +44,30 @@ spec:
...
-                end
+            "@type": "type.googleapis.com/envoy.extensions.filters.http.header_mutation.v3.HeaderMutation"
+            mutations:
+              request_mutations:
+                - append:
+                    header:
+                      key: x-shard-key
+                      value: "%REQ(x-api-key)%"
+                    append_action: APPEND_IF_EXISTS_OR_ADD
+                  condition:
```
krapie: Could you please check whether `condition` field exists in `request_mutations`? Unfortunately I do not see that field in the [proto](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/common/mutation_rules/v3/mutation_rules.proto#envoy-v3-api-msg-config-common-mutation-rules-v3-headermutation).

build/charts/yorkie-cluster/templates/istio/ingress-envoy-filter.yaml
```diff
@@ -44,20 +44,30 @@ spec:
...
-                    end
-                  end
-                end
+            "@type": "type.googleapis.com/envoy.extensions.filters.http.header_mutation.v3.HeaderMutation"
+            mutations:
+              request_mutations:
+                - append:
+                    header:
+                      key: x-shard-key
+                      value: "%REQ(x-api-key)%"
```
krapie: Maybe the value should be `%REQUEST_HEADER(x-api-key)%` as per [header format specifier](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#config-access-log-format)?

build/charts/yorkie-cluster/templates/istio/ingress-envoy-filter.yaml
```diff
@@ -44,20 +44,30 @@ spec:
...
+                    append_action: APPEND_IF_EXISTS_OR_ADD
```
krapie: How about using `ADD_IF_ABSENT` instead if the action should be taken if no `x-shard-key` present? I think `APPEND_IF_EXISTS_OR_ADD` will concat `x-api-key` if `x-shard-key` exists in this case as per [header append action](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-enum-config-core-v3-headervalueoption-headerappendaction)?

FYI: if `x-api-key` header is absent, then `%REQUEST_HEADER(x-api-key)%` will return ``, not absent.

> The same format specifier as used for HTTP access logging applies here, however unknown header values are replaced with the empty string instead of -. Header value is encoded as string. This does not work for non-utf8 characters. Only one of value or raw_value can be set.

https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-msg-config-core-v3-headervalue

raararaara:
Thanks for the review comments, @krapie ! Sorry for the late response. Before proceeding with the current approach, I think we need to consider the following fundamental issue.

## Analysis: Fundamental Issue Discovered Before Lua → Native Filter Replacement

### A. Current Flow: SDK API Call Without API Key

1. Request enters yorkie-gateway
2. Envoy Filter checks: if `x-shard-key` is missing, set it to `"default"`
3. Consistent hashing routes request to the node responsible for `"default"`
4. Yorkie server processes the request

### B. The Problem (Cluster Mode Only)

The node responsible for `"default"` is NOT necessarily the node responsible for the actual document.

| Request Source | x-shard-key Value | Routed To |
|----------------|-------------------|-----------|
| SDK/Direct call (without API Key) | `"default"` (set by filter) | Node X (handles "default") |
| ClusterService | `apiKey/docKey` | Node Y (handles actual document) |

**The same document can be processed by different nodes!**

- SDK request: `Hash("default")` → Node X
- ClusterService: `Hash("apiKey/docKey")` → Node Y (correct node for the document)

The filter's `"default"` fallback doesn't route to the correct document owner - it just routes to whichever node happens to own the hash of the string "default".

### C. Question

Based on the above analysis, the `"default"` fallback in the filter appears to be invalid in cluster mode. I'm considering removing this logic, but since you introduced `useDefaultProject` / Default Project, I wanted to ask: was there a different intention or use case I might be missing?

### D. Proposed Solution

**Non-Cluster Mode (single server, development):**
- Allow API Key omission → Map to Default Project
- This provides development convenience
- No routing issues since there's only one server

**Cluster Mode (production):**
- Remove Lua Filter's `default` fallback logic (it routes to the wrong node anyway)
- Return 400 error when API Key/Shard Key is missing (at Gateway or Server level)
- This aligns with the existing `useDefaultProject: false` configuration in values.yaml

### E. Conclusion

**The Native Filter replacement should be addressed after resolving this fundamental design issue.**

In the current state, replacing Lua → Native is essentially "replacing broken routing logic with different broken routing logic."

Your review comments are valid regarding the Native Filter syntax, but even if we fix those issues, the underlying routing problem would remain unsolved.

krapie:
@raararaara Thank you for the information. That routing fallback logic existed before `ClusterService`, so it did not account for this inconsistency. I think your analysis and proposed solution are valid, and removing the default fallback logic while returning HTTP 400 for unmapped shard/API keys seems like a solid approach.

PR #1627, opened 2026-01-11
Fix to use b.Loop() in benchmark test introduced in Go 1.24

ggyuchive:
- Go 1.24 Release Notes: https://go.dev/doc/go1.24#new-benchmark-function
- b.Loop() Example: https://go.dev/blog/testing-b-loop

In benchmark test, we can run it like below before Go 1.24.
```go
func Benchmark(b *testing.B) {
  ... setup ...
  b.ResetTimer() // if setup may be expensive
  for range b.N {
    ... code to measure ...
    ... use sinks or accumulation to prevent dead-code elimination ...
  }
  b.StopTimer() // if cleanup or reporting may be expensive
  ... cleanup ...
  ... report ...
}
```

From Go 1.24, `b.Loop()` is newly introduced that automatically run `b.ResetTimer()` before start loop and `b.StopTimer()` after loop ended.
```go
func Benchmark(b *testing.B) {
  ... setup ...
  for b.Loop() {
    // optional timer control for in-loop setup/cleanup
    ... code to measure ...
  }
  ... cleanup ...
}
```
`b.Loop()` keep alive function call parameters and results, preventing the compiler from fully optimizing away the loop body. So in CI, **bench test runtime become much faster (from 9min. to 7min.)**

In case of setup and cleanup inside of loop is required, below is preferred.

1. Only setup inside of loop is needed
```go
func Benchmark(b *testing.B) {
  for b.Loop() {
    b.StopTimer()
    ... setup ...
    b.StartTimer()
    ... code to measure ...
  }
}
```

2. setup and teardown inside of loop is needed
- I kept to use `for range b.N` because `b.StopTimer()` should be called before cleanup.
- Run b.Loop() with stop state is invalid.
```go
func Benchmark(b *testing.B) {
  for range b.N {
    b.StopTimer()
    ... setup ...
    b.StartTimer()
    ... code to measure ...
    b.StopTimer()
    ... cleanup ...
  }
}
```

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:

- [X] Addressed and resolved all CodeRabbit review comments
- [X] Didn't break anything

## Summary by CodeRabbit

* **Tests**
  * Updated benchmark tests to use improved iteration patterns for more accurate performance measurements.
  * Refined timing control in benchmark tests for better measurement precision.

* **Chores**
  * Updated test helper utilities to support enhanced benchmark infrastructure.

### DIFF
--- a/test/bench/deactive_bench_test.go
+++ b/test/bench/deactive_bench_test.go
@@ -56,7 +56,6 @@ func benchmarkDeactivate(
 
 	b.StartTimer()
 	err = c.Deactivate(ctx)
-	b.StopTimer()
 	assert.NoError(b, err)
 }
 
@@ -74,13 +73,13 @@ func BenchmarkDeactivate(b *testing.B) {
 
 	ctx := context.Background()
 	b.Run("deactivate with 100 total 50 attached", func(b *testing.B) {
-		for range b.N {
+		for b.Loop() {
 			benchmarkDeactivate(b, ctx, svr, 100, 50)
 		}
 	})
 
 	b.Run("deactivate with 100 total 100 attached", func(b *testing.B) {
-		for range b.N {
+		for b.Loop() {
 			benchmarkDeactivate(b, ctx, svr, 100, 100)
 		}
 	})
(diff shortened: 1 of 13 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #1639, opened 2026-01-17
Fix SplayTree Find semantics for CRDT Array

ggyuchive:
Currently `Splay.Find` function is focused on `Text` type which returns cursor position(index, offset) in tree.
Also `Splay.Find` used in RGATreeList as well, but it's for `Array` type that should satisfy `offset < Value.Len()`.
RGATreeList manages removed node with length 0 and 1 otherwise. So when use same Find() function, it can return wrong result with below.
- Return removed node
- Can be `offset == 1` even maximum Len() is 1.

Below attached picture is current behavior what I understood.
image

After `Splay.Find`, `RGATreeList.Get` iterate next node while non-removed element found. It will be expensive after bulk deletion.

So I divided method to `Splay.FindForText` and `Splay.FindForArray`.
`FindForText` is same with previous `Find`, and `FindForArray` is used by RGATreeList.
It can ensure `Splay.FindForArray` not return removed element.

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:

- [X] Addressed and resolved all CodeRabbit review comments
- [X] Didn't break anything

## Summary by CodeRabbit

* **Refactor**
  * Improved lookup mechanisms for different access patterns (text cursor vs. array-like operations), enhancing performance and error handling specificity for each use case.

* **Tests**
  * Expanded test coverage to validate both text-based and array-based access patterns with additional test scenarios and edge case validation.

* **Documentation**
  * Corrected a documentation reference path.

### DIFF
--- a/pkg/document/crdt/rga_tree_list.go
+++ b/pkg/document/crdt/rga_tree_list.go
@@ -198,29 +198,11 @@ func (a *RGATreeList) InsertAfter(prevCreatedAt *time.Ticket, elem Element, exec
 
 // Get returns the element of the given index.
 func (a *RGATreeList) Get(idx int) (*RGATreeListNode, error) {
-	splayNode, offset, err := a.nodeMapByIndex.Find(idx)
+	splayNode, err := a.nodeMapByIndex.FindForArray(idx)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #1774, opened 2026-04-24
Support IntegerDedupCnt in YSON serialization with HLL round-trip

### LINKED ISSUE #1773: Dedup counter integration gaps in YSON: silent snapshot/revision errors and blocked compaction
## TL;DR

[#1733][pr1733] added `IntegerDedupCnt` to `pkg/document/crdt/counter.go`
but did not update the corresponding logic in `pkg/document/yson`. This
leaves two gaps on the background path launched at
`server/packs/pushpull.go:167`:

- **Part 1 — immediate bug, small fix.** `Counter.Marshal()`,
  `parseCounter()`, and the YSON grammar table don't handle dedup
  counters, so every pushpull against a dedup-containing document causes
  `storeRevision` to fail silently. Client API returns 200 OK while
  server logs fill with `ERROR ... marshal counter: unsupported element`.

- **Part 2 — design discussion, follow-up.** Even after Part 1 is
  addressed, YSON's `Counter` representation captures only the derived
  count (not the HLL register bitmap). This blocks `CompactDocument`'s
  rebuild-compare invariant (`content mismatch after rebuild`) and means
  revision restore cannot recover dedup history.

**A small PR is forthcoming for Part 1.** Part 2 needs design discussion
and can be split into a separate issue if that's preferred for tracking.

[pr1733]: https://github.com/yorkie-team/yorkie/pull/1733

## Environment

- Yorkie server **v0.7.5**
- Introducing commit: [`bcff432c`][bcff432c] (Counter dedup mode)
- Storage: MongoDB
- Reproduced both with the test-helper in-process server (single
  replica) and on a production cluster (20 `yorkie` Deployment
  replicas behind a gateway)

[bcff432c]: https://github.com/yorkie-team/yorkie/commit/bcff432c

---

### Reproduction (deterministic, `go test`)

Self-contained integration test. Relies on
`helper.SnapshotThreshold = 10` to force the background
snapshot/revision goroutine.

```go
//go:build integration

package integration

import (
	"context"
	"fmt"
	"testing"
	"time"

	"github.com/yorkie-team/yorkie/pkg/document/presence"
)

func TestDedupCounterSnapshotBug(t *testing.T) {

	ctx := context.Background()

	// 15 updates > SnapshotThreshold (=10) → forces background snapshot.
	const actors = 15
	for i := 0; i < actors; i++ {
		actor := fmt.Sprintf("user-%d", i)
		err := d1.Update(func(root *json.Object, p *presence.Presence) error {
			if i == 0 {
				root.SetNewDedupCounter("uv").Add(actor)
			} else {
				root.GetCounter("uv").Add(actor)
			}
			return nil
		}, fmt.Sprintf("add %s", actor))
		assert.NoError(t, err)
	}
	time.Sleep(2 * time.Second) // let the background goroutine run

	// A fresh client sees the correct client-visible state.
	assert.Equal(t, fmt.Sprintf(`{"uv":%d}`, actors), d2.Marshal())
}
```

Run:

```bash
go test -tags integration ./test/integration/ -run TestDedupCounterSnapshotBug -v
```

### Observed behavior

The test **passes** (client-visible state is correct), but the server
logs the following on every run:

```
INFO  b2  SNAP: 'testdedupcountersnapshotbug', serverSeq: 16
ERROR b2  store revision of Document (...): marshal counter: unsupported element
          /yorkie/server/packs/pushpull.go:167
          /yorkie/server/backend/background/background.go:91
--- PASS: TestDedupCounterSnapshotBug (2.06s)
```

The `SNAP:` line indicates the snapshot itself is stored (the protobuf
path handles dedup counters correctly via
`JSONElement_Counter.hll_registers`). The subsequent `storeRevision`
step — which uses YSON — is what fails.

### Expected behavior

Documents containing `IntegerDedupCnt` should complete background
snapshot and revision persistence without errors. The server log should
contain no `ERROR` lines for normal pushpulls.

### Root cause

`pkg/document/yson/yson.go` has three unreachable branches for dedup
counters:

| Site | What it does | Current gap |
|---|---|---|
| `Counter.Marshal()` (lines ~187-196) | CRDT → YSON text | `switch y.Type` covers `IntegerCnt`/`LongCnt` only → `default` returns `ErrUnsupported` |
| `parseCounter()` (lines ~429-450) | YSON JSON shape → `yson.Counter` | Same gap in the reverse direction |
| `preprocessTypeValues()` (lines ~530-545) | YSON text → JSON shape | No grammar entry for a dedup variant |

`Counter.Marshal()` is reached from `storeRevision` via
`server/packs/snapshot.go:283` (`yson.FromCRDT` → `ysonObj.Marshal()`)
which is triggered from the goroutine at `pushpull.go:167`.

### Proposed fix (Option A — PR forthcoming)

- Add `IntegerDedupCnt` cases to `Counter.Marshal()`, `parseCounter()`,
  and the YSON grammar table.
- Emit/accept `DedupCounter(Int(N))` as the YSON form. This preserves
  the dedup-type distinction for round-trip within the constraints of
  Option A (see Part 2 for what isn't round-tripped).
- Add unit regression coverage in `pkg/document/yson/yson_test.go`
  (alongside existing `IntegerCnt`/`LongCnt` cases) and an integration
  regression guard derived from the reproducer above.

`LongDedupCnt` does not exist in the crdt package today, so the fix
covers only `Int`. Extending to `Long` is trivial once the corresponding
`crdt.LongDedupCnt` is introduced.

Draft diff (full PR to follow):

```go
// yson.go — Counter.Marshal()
case crdt.IntegerDedupCnt:
    return fmt.Sprintf("DedupCounter(Int(%v))", y.Value), nil

// yson.go — parseTypedValue()
case "DedupCounter":
    return parseCounter(raw, "DedupCounter")

// yson.go — parseCounter(raw, outerType string)
dedup := outerType == "DedupCounter"
...
case "Int":
    if dedup {
        counter.Type = crdt.IntegerDedupCnt
    } else {
        counter.Type = crdt.IntegerCnt
    }

// yson.go — preprocessTypeValues grammar table
// (DedupCounter must precede Counter — substring shadowing)
{`DedupCounter(`, `{"type":"DedupCounter","value":`},
{`Counter(`,      `{"type":"Counter","value":`},
```

### Backward compatibility

Existing YSON outputs never contained `DedupCounter(...)` — under the
current code, any attempt to marshal a dedup counter fails before
producing output. Therefore no stored data uses the new token and
introducing it is non-breaking.

---

### Context

With Option A applied, YSON round-trip for a dedup counter preserves
`{Type, Value}` but **not** the internal HLL register bitmap (~16 KB).
Two observable consequences, both confirmed locally once Part 1 is
addressed:

### 2-a. Revision restore loses dedup history

After restoring to a revision, a previously-deduplicated actor
re-adding itself will be counted again, because the restored counter's
HLL is empty. Only the derived count value survives.

### 2-b. `CompactDocument` rebuild-compare fails

`server/packs/compaction.go:105` asserts `prevMarshalled ==
newMarshalled` after rebuilding the document through
`yson.FromCRDT → SetYSON → yson.FromCRDT`. For a dedup counter the
rebuild produces `Value=0`:

```go
// pkg/document/crdt/counter.go:85-91
case IntegerDedupCnt:
    return &Counter{
        valueType: valueType,
        value:     int32(0),   // passed-in value is intentionally ignored
        createdAt: createdAt,
        hll:       NewHLL(),
    }, nil
```

So after `SetYSON(DedupCounter(Int(15)))`, the new counter's derived
value is `0` (empty HLL). The invariant check fails with
`content mismatch after rebuild`, and housekeeping auto-compaction
(`server/server.go:269` → `documents.CompactDocuments`) silently skips
the document. This is **safe from silent corruption** but leaves the
`changes` collection growing unboundedly for dedup docs.

### Design options to discuss

- **A.** Extend `yson.Counter` with an optional `Registers []byte`
  field, encoded as base64 in YSON text form. `crdt.NewCounter` for
  dedup would accept the bytes and call `RestoreHLL`.
  Pros: full round-trip, compaction works, restore works.
  Cons: breaks YSON human-readability for dedup counters (~16 KB
  base64); requires coordinated changes across yson/json/crdt.

- **B.** Keep YSON minimal. Bypass YSON for dedup counters in the
  compaction rebuild check — use the protobuf snapshot format, which
  already preserves HLL via `JSONElement_Counter.hll_registers`.
  Pros: small targeted change; revision format stays readable.
  Cons: revision restore still loses dedup state; split serialization
  strategy (yson for revisions, proto for compaction check) is harder
  to reason about.

- **C.** Document the limitation. Compaction is not supported for
  dedup counters; operators rely on document TTL or sharding for
  long-term storage management.
  Pros: no code change.
  Cons: silent scalability cliff for long-lived dedup docs.

### Scope suggestion

Part 2 is independent of Part 1 in implementation (they don't share
changed code paths beyond the `yson.Counter` struct). Happy to split
Part 2 into a separate issue if the maintainers prefer that for
tracking — noted here together only to preserve the root-cause narrative
(both stem from the dedup integration in #1733).

---

## Production impact observed

k6 load test, 3000 concurrent VUs on a single analytics daily document
(`pv: IntegerCnt, uv: IntegerDedupCnt`), against a 20-replica cluster
at v0.7.5:

- Client `PushPullChanges` success rate: **100%**
- `yorkie` pod error log rate:
  **~200+ `marshal counter: unsupported element` per minute** across
  replicas (Part 1)
- Housekeeping `compaction` silently blocked on dedup docs (Part 2)
- Older dedup docs from before v0.7.5 stabilized have become
  un-attachable (`Attach` / `Admin GetDocument` return HTTP 500
  `"not applicable datatype"`); we suspect accumulated
  un-compacted/un-snapshotted state, though the precise failure mode on
  those specific docs was not isolated in this investigation.

---

## Attached / linked

- Reproducer: the Go integration test above is self-contained and
  runnable against a standard yorkie test environment.
- The k6 driver that produced the production load test output above is
  in a separate repo (not required for reproduction).
- A PR addressing **Part 1 only** is being prepared:
  `fix(yson): support IntegerDedupCnt in Counter marshal/parse/grammar`.

hackerwins:
## Summary

- Add `DedupCounter(Int(N),"base64...")` YSON text format for dedup counters
- Full HLL register round-trip: CRDT → YSON → CRDT preserves all 16KB of HLL state
- Fix `storeRevision` silent failure and `CompactDocument` rebuild-compare invariant violation for dedup documents

## Changes

| File | Change |
|------|--------|
| `pkg/document/yson/yson.go` | `Registers` field on Counter, Marshal/Parse/Grammar for DedupCounter |
| `pkg/document/yson/to_yson.go` | Extract `HLLBytes()` in `toCounter()` |
| `pkg/document/json/object.go` | Pass HLL registers through `setNewCounter` for SetYSON |
| `pkg/document/json/array.go` | Pass HLL registers through `AddNewCounter` for AddYSON |
| `pkg/document/json/counter.go` | Add `RestoreHLL` method |
| `pkg/document/yson/yson_test.go` | 6 new tests: marshal, unmarshal, round-trip, CRDT conversion, full round-trip |

## Design

Design doc: `docs/design/yson-dedup-counter.md`

Key decisions:
- **Regex pre-substitution** for DedupCounter in `preprocessTypeValues` — the existing global `)` → `}` replacement can't handle the compound structure
- **HLL restored inside creator function** (not after) — Yorkie's clone/execute architecture deep-copies counters inside `setInternal`, so restoration must happen before the copy
- **No `HLL()` wrapper** — second argument is always base64 HLL registers

Fixes #1773

## Test plan

- [x] DedupCounter marshal produces `DedupCounter(Int(N),"base64...")`
- [x] DedupCounter unmarshal parses back to correct Type/Value/Registers
- [x] Standalone and mixed-object round-trip tests
- [x] CRDT → YSON extracts HLL registers
- [x] Full round-trip (CRDT → YSON → new CRDT → YSON) produces identical output (compaction invariant)
- [x] All existing YSON tests pass (no regressions)
- [x] `make lint` passes (0 issues)
- [x] `make test` passes (full unit + integration suite)

## Summary by CodeRabbit

* **New Features**
  * Dedup counters now preserve state during serialization, maintaining accuracy across snapshots, revisions, and compaction operations.

* **Documentation**
  * Added "Serialization" section documenting dedup counter format, marshaling behavior, and round-trip guarantees in design documentation.

* **Tests**
  * Added YSON serialization tests, round-trip validation tests, and integration tests covering snapshot/revision behavior and concurrent scenarios.


test/integration/counter_dedup_snapshot_test.go
```diff
@@ -0,0 +1,134 @@
+//go:build integration
+
+/*
+ * Copyright 2025 The Yorkie Authors. All rights reserved.
```
hackerwins:
```suggestion
 * Copyright 2026 The Yorkie Authors. All rights reserved.
```

PR #1775, opened 2026-04-24
fix(yson): support IntegerDedupCnt in Counter marshal/parse/grammar

emplam27:
## What this PR does

Adds `crdt.IntegerDedupCnt` handling to `pkg/document/yson`. Before this
patch, any background code path that marshaled the document root via
YSON would fail silently for documents containing a dedup counter,
flooding server logs with `marshal counter: unsupported element`.

This addresses **Part 1** of #1773. See the issue for full
context, production impact, and the scope of Part 2 (not addressed
here — requires design discussion around HLL register preservation).

## Changes

Three YSON sites previously covered only `IntegerCnt` / `LongCnt`. All
three now also handle `IntegerDedupCnt`:

- **`Counter.Marshal()`** — emits `DedupCounter(Int(N))` for dedup.
- **`parseTypedValue()`** — dispatches `"DedupCounter"` type.
- **`parseCounter()`** — accepts `outerType string` to construct
  `IntegerDedupCnt` when the outer token is `DedupCounter`.
- **`preprocessTypeValues()`** — adds `DedupCounter(` to the YSON-text
  grammar table. It is placed **before** the existing `Counter(` entry
  because `DedupCounter` has `Counter` as a substring; reversing the
  order would cause `ReplaceAll` to rewrite the inner occurrence first
  and corrupt the grammar.

`LongDedupCnt` is not yet defined in the crdt package; this patch
covers only `Int`. Extending to `Long` is a single additional `switch`
case when `LongDedupCnt` lands.

## Tests

**Unit** — `pkg/document/yson/yson_test.go`:

- `TestYSONMarshal` object subtest: added `IntegerDedupCnt` entry and
  its YSON-text counterpart.
- `TestYSONMarshal` array subtest: added `IntegerDedupCnt` entry and
  its YSON-text counterpart.
- New `TestYSONMarshal` "dedup counter marshal/unmarshal test" subtest
  asserting the explicit round-trip form `DedupCounter(Int(42))` and
  that the decoded `Counter.Type` equals `crdt.IntegerDedupCnt`.

These fail without the fix with `ErrUnsupported` and pass after.

**Integration** — `test/integration/counter_dedup_snapshot_test.go`
(new file, build tag `integration`):

- `TestCounterDedupSnapshotReproSingleClient` — 15 adds against a
  dedup counter (> `helper.SnapshotThreshold = 10`) to force the
  background snapshot / revision goroutine. Asserts client-visible
  state is correct and a fresh client attaches successfully.
- `TestCounterDedupSnapshotReproConcurrent` — N concurrent clients each
  adding 8 distinct actors to the same doc. Exercises the background
  path under contention and verifies convergence for a verifier client.

Before this patch these tests still PASS at the client level but the
server logs `store revision of Document ... : marshal counter: unsupported element`
on every run. After this patch the log contains only the expected
`SNAP:` and `REV:` INFO lines.

(The concurrent test additionally surfaces `CompactDocument` warnings
— `content mismatch after rebuild` — which are caused by a separate
issue outside this PR's scope. See "Out of scope" below.)

## How to verify locally

```bash
# Unit
go test ./pkg/document/yson/...

# Integration (requires a MongoDB reachable at the test default URI)
go test -tags integration ./test/integration/ \
    -run "TestCounterDedup|TestYSONMarshal" -v -count=1
```

Expected: all tests PASS, no `ERROR` lines in the server log for the
new tests.

## Out of scope (tracked by Part 2 of #1773)

This patch preserves only the counter's derived `Value` across YSON
round-trip. The HLL register bitmap is not represented in YSON, which
means:

- **`CompactDocument` rebuild-compare fails for dedup docs.** The
  invariant check at `server/packs/compaction.go:105` fails because
  `crdt.NewCounter(IntegerDedupCnt, N, …)` ignores the passed value
  and initializes the counter with empty HLL state. Housekeeping
  auto-compaction silently skips these docs — safe, but changes keep
  accumulating.
- **Revision restore cannot recover dedup history.** The counter's
  count value is restored, but previously-deduplicated actors would be
  counted again on subsequent `Add`.

These require a YSON format change (add `Registers []byte` to
`yson.Counter`) or a strategy change (bypass YSON for dedup in the
compaction invariant). Rather than bundle them into this PR, the
discussion is kept in the linked issue under Part 2.

## Backward compatibility

Existing YSON outputs never contained `DedupCounter(...)` — under the
pre-patch code, any attempt to marshal a dedup counter returned
`ErrUnsupported` before producing output, so no stored revision in the
wild uses the new token. Introducing it is therefore non-breaking for
existing data.

The `parseCounter` signature changes from
`parseCounter(raw) → Counter` to
`parseCounter(raw, outerType string) → Counter`. This function is
unexported, so no external callers are affected.

## Closes

Issue #1773 — Part 1 only. Part 2 remains open for design
discussion.

## Release Notes

* **New Features**
  * Added support for serializing and deserializing deduplicated integer counters with proper type handling and round-trip conversion.

* **Tests**
  * Extended test coverage with regression and integration tests validating deduplicated counter behavior during snapshots and concurrent operations.

### DIFF
--- a/pkg/document/yson/yson.go
+++ b/pkg/document/yson/yson.go
@@ -190,6 +190,8 @@ func (y Counter) Marshal() (string, error) {
 		return fmt.Sprintf("Counter(Int(%v))", y.Value), nil
 	case crdt.LongCnt:
 		return fmt.Sprintf("Counter(Long(%v))", y.Value), nil
+	case crdt.IntegerDedupCnt:
+		return fmt.Sprintf("DedupCounter(Int(%v))", y.Value), nil
 	default:
 		return "", fmt.Errorf("marshal counter: %w", ErrUnsupported)
 	}
@@ -305,7 +307,9 @@ func Unmarshal(data string, elem Element) error {
 		}
 
 	case *Counter:
-		counter, err := parseCounter(raw.(map[string]interface{}))
+		rawMap := raw.(map[string]interface{})
+		outerType, _ := rawMap["type"].(string)
+		counter, err := parseCounter(rawMap, outerType)
 		if err != nil {
 			return err
 		}
@@ -343,7 +347,9 @@ func parseTypedValue(raw map[string]interface{}) (interface{}, error) {
 
 		return val, nil
 	case "Counter":
-		return parseCounter(raw)
+		return parseCounter(raw, "Counter")
+	case "DedupCounter":
+		return parseCounter(raw, "DedupCounter")
 	case "Tree":
 		if value, ok := raw["value"].(map[string]interface{}); ok {
 			return parseTree(value)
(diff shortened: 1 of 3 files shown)


hackerwins:
The Part 1 fix from this PR has been incorporated into #1774, which also addresses Part 2 (HLL register preservation across YSON round-trip). Specifically:

- The `DedupCounter(Int(N))` marshal/parse/grammar support is covered with a full HLL round-trip format: `DedupCounter(Int(N),"base64...")`
- The integration tests from this PR (`TestCounterDedupSnapshotSingleClient`, `TestCounterDedupSnapshotConcurrent`) have been adopted into #1774

Closing in favor of #1774. Thank you for the detailed issue report and reproduction — it made the design work much easier.

PR #1786, opened 2026-04-29
Reduce lock contention in channel manager for high-concurrency skew

raararaara:

Reduces lock contention in `channel.Manager` for high-concurrency single-channel (skew) workloads. The previous full per-channel `Mutex` serialized Attach and Detach for every session on the same channel, producing thousands of blocked goroutines under 20K skew. This PR replaces serialization with finer-grained synchronization across three call sites:

1. **Atomic `activeCount` + `RWMutex` on `Channel`**: Track session count via `atomic.Int64`. Attach holds `ch.mu` only as a **read lock** during the commit phase (trie re-check + `Sessions.Set` + reverse-index updates); concurrent Attaches share the read lock and do not contend with each other. Detach takes the **write lock** only when `activeCount` transitions to zero (channel deletion), so the write lock is rare in steady state. The decrement is gated on `Sessions.Delete`'s bool return value to prevent over-decrement when two detaches race for the same session ID.
2. **Get fast path in `ShardedPathTrie`**: `getOrCreateShard()` tries `Get` (lock-free) before falling back to `Upsert` (write lock). Existing shards avoid the write lock entirely.
3. **Atomic `updatedAt` in `Session`**: `Session.UpdatedAt time.Time` is replaced with `updatedAt atomic.Int64` (UnixNano). Refresh updates the timestamp via `atomic.Store` instead of `Sessions.Set`, making the Refresh path lock-free on the Sessions cmap.

Benchmark results (20K skew):

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| attach med | 1,170ms | 568ms | -51.5% |
| detach med | 1,300ms | 71ms | -94.5% |
| http_req p(95) | 1,990ms | 1,170ms | -41.2% |
| blocked goroutines | 8,994 | 1,608 | -82% |
| throughput | 14,206/s | 14,719/s | +3.6% |
| failure rate | 0.00% | 0.00% | — |

Fixes #

- The synchronization relies on two cooperating mechanisms: atomic `activeCount` to fast-path Detach's deletion decision, and the `RWMutex` to serialize Attach's commit with Detach's actual deletion. Either mechanism alone is insufficient — atomic ordering does not establish happens-before across the trie store, so the read/write lock is required to prevent a window in which Attach observes the channel as present and commits while Detach removes the channel from the trie.
- `Session.UpdatedAt` changed from exported `time.Time` to unexported `updatedAt atomic.Int64`. TTL expiration in `CleanupExpired` reads via `time.Unix(0, atomic.Load())`.
- `TestChannelManager_RaceConditions` covers the attach/detach deletion race, repeated full-detach cycles, massive single-channel contention, channel recreation after full detach, session-count accuracy under contention, and refresh/cleanup overlap with TTL expiration. The deletion-race subtest reliably reproduced the original bug within ~100 rounds and now passes 2,000 rounds under `-race`.

```release-note
NONE
```

**Additional documentation**:

```docs

```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

* **Performance**
  * Faster shard lookups to avoid unnecessary work when shards already exist.
  * More efficient session timestamp handling for lower overhead.

* **Bug Fixes**
  * Improved reliability of concurrent session attach/detach to prevent lost sessions.
  * Fixed race conditions in session expiration and cleanup under high concurrency.

* **Tests**
  * Added extensive concurrency tests to validate attach/detach, refresh, expiration, and cleanup behavior.

### DIFF
--- a/api/types/id.go
+++ b/api/types/id.go
@@ -21,6 +21,7 @@ package types
 import (
 	"encoding/hex"
 	"fmt"
+	"sync"
 
 	"go.mongodb.org/mongo-driver/v2/bson"
 
@@ -85,7 +86,15 @@ func IDFromBytes(bytes []byte) ID {
 	return ID(hex.EncodeToString(bytes))
 }
 
+var actorIDCache sync.Map // time.ActorID → ID
+
 // IDFromActorID returns ID represented by the encoded hexadecimal string from actor ID.
+// It caches results to avoid repeated hex encoding for the same actor.
 func IDFromActorID(actorID time.ActorID) ID {
-	return IDFromBytes(actorID.Bytes())
+	if v, ok := actorIDCache.Load(actorID); ok {
+		return v.(ID)
+	}
+	id := IDFromBytes(actorID.Bytes())
+	actorIDCache.Store(actorID, id)
+	return id
 }
--- a/api/types/id.go
+++ b/api/types/id.go
@@ -21,7 +21,6 @@ package types
 import (
 	"encoding/hex"
 	"fmt"
-	"sync"
 
 	"go.mongodb.org/mongo-driver/v2/bson"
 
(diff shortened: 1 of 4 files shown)


hackerwins:
### Code review

Found 1 issue:

1. `Detach` no longer atomically claims ownership of the `activeCount` decrement. The existence check (`ch.Sessions.Get(id)`) at line 336 and the decrement (`ch.activeCount.Add(-1)`) at line 357 are non-atomic. Two concurrent `Detach` calls for the same session ID — e.g., an RPC-driven detach racing with `CleanupExpired` for an expired session, or a client retrying Detach — can both pass the existence check, both call `ch.Sessions.Delete(id)` (the second is a no-op on the cmap shard but the code ignores its return value), and both decrement `activeCount`. The over-decrement leaves `activeCount < 0`, so the `== 0` re-check under lock at line 361 never fires and the channel is leaked from `m.channels` even after the last session leaves. A simple fix is to gate the decrement on the return value of `Delete`: `if ch.Sessions.Delete(id) { … decrement … }`, or use an atomic get-and-delete primitive.

https://github.com/yorkie-team/yorkie/blob/252d2bcc54448f5779c4291c9568a5ef4b6fd06c/server/backend/channel/manager.go#L335-L365

PR #1794, opened 2026-05-11
Cascade channel session detach on client deactivate

JOOHOJANG:
## Summary

- Clients that lose their browser without sending an explicit `DetachChannel` (refresh, tab close, network drop) stay in channel polling for up to ~60s, until `ChannelSessionTTL` expires and the cleanup ticker sweeps them. `clients.Deactivate` already cleans up documents via cluster RPC but leaves channel sessions untouched, and channels are sharded per node so no single server knows where a given actor's sessions live.
- Adds an actor-keyed cluster RPC `DetachActorFromChannels` and a backend helper `BroadcastDetachActorFromChannels` (mirrors `BroadcastChannelCount`) that fans the detach out to every cluster node and runs the existing `Detach` path locally for that actor's sessions.
- Calls the broadcast from `clients.Deactivate` as best-effort: per-node failures are logged but do not fail deactivation. `ChannelSessionTTL` remains the final safety net for nodes we couldn't reach.

## Why server-side cascade vs SDK fix

Cleaning up on `pagehide`/`beforeunload` in each SDK has structural limits: every SDK language needs the same fix, browsers cancel pending fetches on unload (sendBeacon partially mitigates), and any missed cleanup still falls through to the same TTL window. The server already produces a deactivation event for every client lifecycle; cascading from that one event covers all SDKs and unload paths in one place, with the TTL as a true last-resort backstop.

## Test plan

- [x] `go test ./server/backend/channel/` — new `DetachByActor` unit tests + existing channel suite pass
- [x] `go test -tags integration ./test/integration/ -run TestChannelIntegration` — full channel integration suite passes, including new `deactivate cascades channel session detach` test that asserts `SessionCount` drops within 10s (well below 60s TTL)
- [x] Manual verification with `./bin/yorkie server -l debug`: 10 page refreshes + 1 new tab produce immediate `Detach` events with no `CHAN: ... expires[]` cleanup lines (prior behavior had ~9 expires per cycle)

## Out of scope

- SDK changes: `DetachChannel` RPC API stays as-is for explicit per-channel unsubscribe.
- Optimizing fan-out under high housekeeping volume (N inactive clients × M cluster nodes RPCs per tick). Same amplification pattern exists in `BroadcastChannelList`; revisit only if operational data shows it's a problem.

## Summary by CodeRabbit

* **New Features**
  * Cluster-wide cascade to immediately detach channel sessions for an actor when a client deactivates.
  * New cluster RPC to request actor-detach across nodes.

* **Documentation**
  * Added task design and lessons learned documenting the cascade approach and rollout guidance.

* **Tests**
  * Added integration and unit tests verifying immediate detach behavior and manager-level detach logic.

[Review Change Stack](https://app.coderabbit.ai/change-stack/yorkie-team/yorkie/pull/1794)

### DIFF
--- a/api/docs/yorkie/v1/cluster.openapi.yaml
+++ b/api/docs/yorkie/v1/cluster.openapi.yaml
@@ -3,7 +3,7 @@ info:
   description: Yorkie is an open source document store for building collaborative
     editing applications.
   title: Yorkie
-  version: v0.7.5
+  version: v0.7.8
 servers:
 - description: Production server
   url: https://api.yorkie.dev
@@ -18,6 +18,18 @@ paths:
       responses:
         "200":
           $ref: '#/components/responses/yorkie.v1.ClusterService.CompactDocument.yorkie.v1.ClusterServiceCompactDocumentResponse'
+        default:
+          $ref: '#/components/responses/connect.error'
+      tags:
+      - yorkie.v1.ClusterService
+  /yorkie.v1.ClusterService/DetachActorFromChannels:
+    post:
+      description: ""
+      requestBody:
+        $ref: '#/components/requestBodies/yorkie.v1.ClusterService.DetachActorFromChannels.yorkie.v1.ClusterServiceDetachActorFromChannelsRequest'
+      responses:
+        "200":
+          $ref: '#/components/responses/yorkie.v1.ClusterService.DetachActorFromChannels.yorkie.v1.ClusterServiceDetachActorFromChannelsResponse'
         default:
           $ref: '#/components/responses/connect.error'
       tags:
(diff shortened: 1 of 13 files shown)


hackerwins:
Thanks for your contribution.

Light follow-up for the benchmark phase:

The current fan-out broadcasts to all N cluster nodes regardless of where the actor's sessions actually live. For a client attached to a single channel, that's N-1 no-op RPCs, plus sequential per-node rpcTimeout accumulating into Deactivate latency. Could matter in larger clusters or during deactivation storms.

If the benchmark surfaces this, one option: derive target nodes from clientInfo(e.g. clientInfo.Documents) via consistent hashing and only RPC the nodes that actually host the actor's sessions.

It would be good to keep this for a follow-up if benchmarks show it's a real issue.

hackerwins:
Closing due to performance concerns: the per-deactivate fan-out RPC to every cluster node amplifies housekeeping load (N inactive clients × M nodes per tick) and makes this approach hard to use in practice. The 60s TTL fallback remains acceptable for now; we'll revisit with a lower-overhead design (e.g., routing detach only to the node owning the actor's sessions, or batching) if user-visible latency becomes an issue.

PR #1802, opened 2026-05-12
Collapse channel lifecycle RPCs + reduce session TTL to 15s (in-memory, no Mongo read)

JOOHOJANG:
## Summary

Combines two of the in-flight channel experiments into one branch so the bench matrix gets a fifth row:

- **PR #1800** (`consolidate-channel-rpcs-in-memory`): collapses the 5-RPC lifecycle into a single `RefreshChannel` **and** removes the per-heartbeat `FindActiveClientInfo` Mongo lookup (liveness proven by the in-memory channel session map).
- **PR #1798** (`reduce-channel-session-ttl`): drops `DefaultChannelSessionTTL` from 60s to 15s, cutting the stale-membership window from 60–70s to 15–25s.

This branch = #1800 + the single TTL commit from #1798. No conflicts (the two PRs touch disjoint files).

## Bench matrix (with this branch added)

| Branch | Lifecycle RPCs | Per-heartbeat Mongo read | Session TTL |
|---|---|---|---|
| `main` | 5 | yes | 60s |
| #1798 | 5 | yes | **15s** |
| #1799 | 1 | yes | 60s |
| #1800 | 1 | **no** | 60s |
| (companion #1801) | 1 | yes | **15s** |
| **this** | **1** | **no** | **15s** |

All three axes (RPC collapse, Mongo read, TTL) varied independently so any latency / stale-window delta can be attributed to the right cause.

## Why this is the most aggressive variant

- Fewest RPCs per session lifecycle (1)
- Fewest Mongo round-trips per heartbeat (0)
- Shortest stale window on silent disconnect (15–25s)

If the benchmark suite favors this combination, it's the strongest candidate among the experiments. If it regresses, the matrix above tells us which axis caused it.

## SDK compatibility note (carried over from #1798)

Go client default `ChannelHeartbeatInterval` is 30s. A server with 15s TTL + an unmodified SDK will flap live sessions. Rollouts must either:
1. Override `channel-session-ttl` on the server to ≥ 3× the SDK heartbeat in use, or
2. Bump SDK defaults in lockstep.

## Test plan

- [x] `make build` passes
- [x] `go build ./...` clean
- [ ] Benchmark suite (oss.navercorp.com PR #347) re-run on this branch and compared against #1799 / #1800 / #1801 / main
- [ ] Verify TTL-vs-SDK-heartbeat safety margin holds when running the bench k6 script

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
[...]

PR #1803, opened 2026-05-13
Add read_only flag to channel attach (design + impl)

JOOHOJANG:
## Summary

Adds a `read_only` flag to channel attach so a client can register with a channel — receiving pubsub events, broadcasts, and TTL accounting — but be excluded from the reported `session_count`. This decouples count *display* from count *contribution*, enabling a UI pattern where a surrounding page shows a live count without inflating it (e.g. a "join writers (N writing)" floating button that, when clicked, navigates to a composer route that attaches as a participant).

Stacked on PR #1802 (`consolidate-channel-rpcs-in-memory-ttl15`).

## Changes

**proto** — `api/yorkie/v1/yorkie.proto`
- `bool read_only = 3` on `AttachChannelRequest`
- `bool read_only = 6` on `RefreshChannelRequest` (meaningful only on the first call)

**channel manager** — `server/backend/channel/manager.go`
- `Session.ReadOnly` field; immutable for the session's lifetime
- `Manager.Attach(ctx, key, clientID, readOnly)` signature gains the flag
- `SessionCount` filters via a new `countParticipants` helper (O(N) over Sessions, called only at attach / heartbeat / detach — not per broadcast)
- `Attach`/`Detach` publish the filtered count to pubsub so other clients see the participant count, not the total session count

**RPC handlers** — `server/rpc/yorkie_server.go`
- legacy `AttachChannel` and the collapsed `firstChannelRefresh` both forward `req.Msg.ReadOnly` into `Channel.Attach`
- heartbeat path is untouched; read-only-ness is a property of the existing session

**tests** — `server/backend/channel/manager_test.go`
- read-only attach does not bump the reported count
- mixed populations (read-only + participants) report only participants
- detaching a read-only session leaves the count unchanged
- the flag is immutable: re-attach with `readOnly=false` returns the existing read-only session (upgrades go through detach + new attach, typically a provider remount)
- read-only sessions are reaped by TTL like participants
- `SessionCount(key, includeSubPath=true)` filters across hierarchical channels

## Design

See [`docs/design/channel-readonly-attach.md`](docs/design/channel-readonly-attach.md) for the full rationale, alternatives considered (PeekChannel RPC, two-channel split, broadcast signaling, metadata flagging, Watch-without-attach), and explicit decisions:

- Flag on `Session` rather than a parallel "observers" map — single membership list keeps `activeCount`, TTL cleanup, and pubsub fan-out paths unchanged
- Filter at read-time, not write-time — avoids a second source of truth under concurrent attach/detach
- `read_only` is immutable per session — typical upgrade UX goes through route-change-driven re-attach
- Both modes share the `AttachChannel` ReadWrite permission — any client may read-only-attach (a stricter permission tier can be added later without breaking existing callers)
- Broadcast is **not** restricted by the flag — accounting and message-flow direction are orthogonal; a future `silent` flag is a better fit if "uncounted *and* send-disabled" is ever needed

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` clean
- [x] `TestChannelManager_ReadOnly/*` (7 sub-tests) pass
- [x] full channel package suite passes (115/115) — no regression in existing tests

## Companion

JS SDK + React + example: forthcoming PR on yorkie-team/yorkie-js-sdk.

### DIFF
--- a/api/yorkie/v1/yorkie.pb.go
+++ b/api/yorkie/v1/yorkie.pb.go
@@ -2378,8 +2378,13 @@ type RefreshChannelRequest struct {
 	// is empty). The server activates the client and attaches it to the channel
 	// before refreshing, collapsing ActivateClient + AttachChannel + RefreshChannel
 	// into a single round trip.
-	ClientKey     string            `protobuf:"bytes,4,opt,name=client_key,json=clientKey,proto3" json:"client_key,omitempty"`
-	Metadata      map[string]string `protobuf:"bytes,5,rep,name=metadata,proto3" json:"metadata,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
+	ClientKey string            `protobuf:"bytes,4,opt,name=client_key,json=clientKey,proto3" json:"client_key,omitempty"`
+	Metadata  map[string]string `protobuf:"bytes,5,rep,name=metadata,proto3" json:"metadata,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
+	// read_only is meaningful only on the first call (session_id empty). When
+	// true, the resulting session is attached to the channel but excluded from
+	// session_count reported to all clients. Read-only sessions still receive
+	// broadcasts and channel events; the flag only affects count accounting.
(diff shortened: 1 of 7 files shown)


JOOHOJANG:
Superseded by #1805 — single consolidated PR against main.

Issue #1816, opened 2026-05-26
Flaky TestClusterNodes leadership-revocation subtest due to tight Eventually


raararaara:
**What happened**:

`TestClusterNodes/Leadership_revocation_and_reacquisition_after_temporary_DB_disconnection_test`
in `test/integration/clusternodes_test.go:202` failed in CI on an
unrelated PR, while every other subtest in the same `TestClusterNodes`
group passed in the same run:

```
--- FAIL: TestClusterNodes (15.09s)
    --- PASS: TestClusterNodes/Identical_active_cluster_nodes_across_servers_test (2.69s)
    --- PASS: TestClusterNodes/Leadership_reacquisition_after_server_restart_test (2.64s)
    --- FAIL: TestClusterNodes/Leadership_revocation_and_reacquisition_after_temporary_DB_disconnection_test (1.95s)
    --- PASS: TestClusterNodes/Concurrent_connection_leadership_election_test (5.12s)
    --- PASS: TestClusterNodes/Leader_graceful_shutdown_and_succession_test (2.69s)
```

Observed in: <CI run / job link>

The PR whose CI exposed this failure does not touch `server/backend/membership`
or any package it depends on, so the failure does not appear to be
caused by changes under test.

**What you expected to happen**:

After `mockDB.SetDisconnected(true)`, the disconnected leader should
lose its lease and the cluster-nodes view should converge to zero
active nodes; after `mockDB.SetDisconnected(false)`, a new lease should
be acquired with a different `LeaseToken`. The subtest asserts all of
this should complete within 1s `Eventually` windows.

**How to reproduce it (as minimally and precisely as possible)**:

This is an integration test gated by `-tags integration` and requires
the sharded MongoDB fixture brought up by
`build/docker/sharding/docker-compose.yml`. The CI workflow that
exercises it is the `complex` / integration job in
`.github/workflows/ci.yml`:

```sh
docker compose -f build/docker/sharding/docker-compose.yml up -d --wait
# (plus the init-config / init-shard scripts as in CI)

go test -tags integration -race -v -count=1 \
  -run "TestClusterNodes/Leadership_revocation_and_reacquisition_after_temporary_DB_disconnection_test" \
  ./test/integration/...
```

Repeated runs are expected to fail intermittently; a single local run
is unlikely to be conclusive.

**Anything else we need to know?**:

Code reading suggests the failure is timing-sensitive rather than a
correctness bug. The subtest configures very tight membership timers:

```go
[...]

Issue #1815, opened 2026-05-26
Flaky TestLRUWithStats/hit_rate_calculation due to sharded LRU per-shard capacity


raararaara:
**What happened**:

`TestLRUWithStats/hit_rate_calculation` in `pkg/cache/cache_test.go`
fails intermittently with output like:

```
cache_test.go:69: Not equal: expected: 3 actual: 2   // Hits
cache_test.go:70: Not equal: expected: 2 actual: 3   // Misses
cache_test.go:72: Not equal: expected: 60 actual: 40 // HitRate
```

Measured locally on `origin/main` (commit fecea21a):

| Branch | Failures / 20 runs |
| --- | --- |
| `main` | 5 / 20 (25%) |

The same rate reproduces on unrelated feature branches, confirming this
is independent of any in-flight PR. The failing assertion always loses
exactly one of the three pre-loaded keys.

**What you expected to happen**:

After `c.Add("key1", 1)`, `c.Add("key2", 2)`, `c.Add("key3", 3)` into
an LRU of capacity 5, all three subsequent `c.Get("key1..3")` calls
should hit. The test asserts exactly that (`hits=3`, `misses=2`,
`HitRate=60`).

**How to reproduce it (as minimally and precisely as possible)**:

```sh
cd <yorkie>
for i in $(seq 1 20); do
  go test -race ./pkg/cache/... -run TestLRUWithStats -count=1 2>&1 \
    | tail -1
done | sort | uniq -c
```

Expect ~15% to 25% of runs to print `FAIL`. The failure rate is
independent of `-race` and of test ordering.

**Anything else we need to know?**:

Root cause is in `pkg/cache/lru_with_stats.go` introduced by #1783
("Replace single LRU with sharded LRU"):

```go
const numShards = 16
var hashSeed = maphash.MakeSeed()   // random per process start

func NewLRU[K comparable, V any](size int, name string) (*LRU[K, V], error) {
    perShard := size / numShards
    if perShard < 1 {
        perShard = 1
    }
    // ...creates 16 shards of capacity `perShard` each.
}
```

For the test call `NewLRU[string, int](5, "test-cache")`:

- `perShard = 5 / 16 = 0`, then clamped to `1`.
- The cache has 16 shards, each of capacity **1**, so total capacity
  is `16 * 1 = 16` keys — but **at most one key can live in any single
  shard at a time**.
- `c.shard(key)` distributes keys via `maphash.Comparable(hashSeed, key)`,
[...]

PR #1820, opened 2026-05-26
Fix flaky leadership revocation and LRU hit-rate tests

hackerwins:
## Summary

Fix two pre-existing flaky tests, both caused by tests relying on assumptions that only hold under specific timing or hashing luck. Production code unchanged in both cases.

### 1. `TestClusterNodes/Leadership_revocation...` (~25–35% flake)

After #1804 added a pre-check that counts the caller's own row as an active leader, a node that loses in-memory lease on DB disconnect intentionally waits for natural lease expiry before reacquiring. The test assumed immediate new-token reacquisition, so it could observe a resurfaced stale row (same `lease_token`, `is_leader=true`) and fail `assert.NotEqual(prvToken, currToken)`. Also observed on an unrelated dependabot PR.

**Fix:** strengthen the post-reconnect `Eventually` predicate with `LeaseToken != prvToken`, bump its timeout to 2s, and re-read the live leader immediately before `SetDisconnected(true)` so `prvToken` reflects MongoDB's latest renewal.

### 2. `TestLRUWithStats/hit_rate_calculation` (~18% flake)

`pkg/cache.LRU` is sharded across 16 fixed shards with per-shard capacity floored at 1, so the test's `size=5` cache had one slot per shard. The shard mapping uses `maphash` with a process-scoped random seed; three keys hashed into 16 shards collide with probability ~18%, and on a single-slot shard a collision evicts the older entry, observed as expected 3 hits/2 misses vs actual 2/3. Hidden locally by `go test` result caching (within one process the seed is fixed; iterations reuse the cached result).

**Fix:** bump the test cache size to 64 so per-shard capacity is 4 — enough to hold all three added keys even on worst-case shard collision.

## Test plan

- [x] `go test -tags integration -run 'TestClusterNodes/Leadership_revocation' -count=30 ./test/integration/` — 30/30 pass
- [x] `go test -tags integration -run 'TestClusterNodes' -count=3 ./test/integration/` — 18/18 pass (no regressions in other subtests)
- [x] LRU fix repro: 30 fresh processes with `go clean -testcache` between each — 30/30 pass (was 6/30 fail before fix)
- [x] `go test ./pkg/cache/` — 8/8 pass
- [x] `make lint` — clean
- [ ] CI green

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced intermittent CI flakes causing false leader reappearance and cache hit‑rate instability.

* **Tests**
  * Made post-disconnect reacquisition checks require a changed lease token and increased timeouts; increased test cache size to avoid shard-collision flakiness.

* **Documentation**
  * Added investigation notes, verification results, guidance on asserting state transitions, timing adjustments, cache sizing, and avoiding seed-dependent test caching.

[Review Change Stack](https://app.coderabbit.ai/change-stack/yorkie-team/yorkie/pull/1820?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/tasks/active/20260527-clusternodes-flaky-test-lessons.md
+++ b/docs/tasks/active/20260527-clusternodes-flaky-test-lessons.md
@@ -60,6 +60,48 @@ disconnect switch, narrowing the race window from
 when you compare a captured value against a live one, capture the
 captured value as close to the change-of-state as possible.
 
+## A test cache size below the shard count is a process-dependent flake
+
+`pkg/cache.LRU` is sharded across 16 fixed shards, with per-shard
+capacity computed as `size / numShards` and floored at 1. A test that
+constructs the cache with `size=5` therefore gets a 16-shard cache
+with one slot per shard — a 16-slot pool with the strict additional
+constraint that no shard may hold more than one entry.
+
+The shard mapping is `maphash.Comparable(hashSeed, key)`, and
+`hashSeed` is initialized once per process via `maphash.MakeSeed()`.
+That means: within a single process the shard assignment is stable
+(so `-count=N` cannot reproduce the flake), but across processes it
+is random. Three keys hashed into 16 shards with seed-determined
+positions have probability `1 − (16·15·14)/16³ ≈ 18%` of producing at
+least one collision, which on a single-slot shard is an eviction.
+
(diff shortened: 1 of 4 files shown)


hackerwins:
Fixes #1816 #1815

PR #1833, opened 2026-06-09
Self-prune leaked Subscription via consecutive Publish failures

raararaara:

Adds a self-prune fallback so `Subscription` entries are reclaimed
when their owning WatchDocument stream never delivers a clean
termination signal (half-closed TCP, hung client). A `Subscription`
counts consecutive `Publish` failures and self-closes once
`maxFailures` is reached; the `BatchPublisher` reaps dead entries on
the next cycle. Stream-lifecycle cleanup via the handler's `defer`
remains the primary path. The leak surfaces as log volume — one
`Publish to ... timeout or closed` line per dead sub per publish
window — and as publish-loop latency for live subs sharing the
publisher, not as memory pressure.

Fixes #

```release-note
NONE
```

Additional documentation:

- Design: pub-sub Risks and Mitigation: docs/design/pub-sub.md
- Task plan: docs/tasks/active/20260609-subscription-self-prune-todo.md

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

* **New Features**
  * Automatic self-pruning of stuck/abandoned subscriptions so pub/sub entries are detected and removed locally, preventing indefinite memory buildup and noisy retries.

* **Documentation**
  * Expanded design notes detailing risks, the self-prune mitigation, and its limitation to the owning pod (no cluster-wide reaping).

* **Tests**
  * Added unit and integration tests validating failure-counting, self-prune behavior, and cleanup of stuck consumers.

### DIFF
--- a/docs/design/pub-sub.md
+++ b/docs/design/pub-sub.md
@@ -187,4 +187,30 @@ func (s *yorkieServer) WatchDocument(
 
 ### Risks and Mitigation
 
-Currently, Subscription instances are managed in memory.
+Subscription instances are managed in memory on the pod that owns the
+stream. Two concrete risks follow from this.
+
+**1. Subscription leak when the stream never terminates cleanly.**
+The only cleanup path tied to a `Subscription` is the WatchDocument
+stream handler's `defer s.backend.PubSub.Unsubscribe(...)`. If the
+stream context never reaches `Done()` — half-closed TCP, dropped
+HTTP/2 stream, a client that crashes without an explicit close — the
+entry survives in `Subscriptions.internalMap` indefinitely. The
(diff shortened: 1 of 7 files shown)


hackerwins (commented):
Thanks for your contribution.

One concern before approving: the new tests verify the buffer-full symptom, but not the qa-cluster root cause.

Two asks:

**A.** Deploy this branch to the QA cluster and confirm the leak logs / memory actually drop. Most direct signal we can get.

**B.** Add a test that reproduces the QA scenario more directly if it is possible. It is also OK if the current integration test cover that.

hackerwins (approved):
Thanks for your contribution.

PR #1841, opened 2026-06-17
Add disable_presence Document option for presence-free docs

raararaara:

Adds a Document-scope `disable_presence` option so a document can
declare it never accepts or stores presence. Motivated by high-fan-out
Counter workloads where the presence map accumulates monotonically on
the server because clients reach end-of-life without sending `detach`.

The flag is declared at first attach via
`AttachDocumentRequest.disable_presence`, persisted on
`DocInfo.DisablePresence` via mongo `$setOnInsert` (immutable
thereafter), and read by the PushPull handler into
`PushPullOptions.DisablePresence`. The server then:
- Strips presence changes at PushPull entry (`stripPresenceChanges`).
- Serializes an empty presence map in `pullSnapshot` / `storeSnapshot`.
- Defensively drops any stray cached presence on the read path
  (`pullChangeInfos`).

Go SDK exposes `WithDisablePresence()`, sends the wire field on first
attach, reads the server-fixated value back from
`AttachDocumentResponse.disable_presence`, and gates its own
`Document.Update` so opt-in clients silently no-op on presence
updates. JS SDK ships as a follow-up PR.

Fixes #

- `client/client.go` calls `d.InternalDocument().ResetPresences()`
  after the attach response when the server-fixated value differs
  from the local opt. Without it, a client that attached without
  opting out but landed on an already-presenceless doc would keep a
  stale local entry. Covered by the
  `late_attacher_observes_the_persisted_value` integration test.
- The `stripPresenceChanges` benchmark lives at
  `server/packs/strip_bench_test.go` because the helper is
  package-private to `server/packs`. Build tag `bench`.
- `pkg/document` has no logger dependency, so the SDK-side
  silent-drop warning uses stdlib `log.Printf` with a `sync.Once` to
  fire exactly once per Document.

Yes.

```release-note
Add `disable_presence` Document option (Go SDK + proto). Documents
attached with `WithDisablePresence()` do not produce, consume, or
store presence. The option is fixated at first attach via mongo
`$setOnInsert` and immutable thereafter; subsequent attaches observe
the persisted value from `AttachDocumentResponse.disable_presence`.
```

**Additional documentation**:

- Task plan: docs/tasks/active/20260616-presenceless-document-option-todo.md

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

* **New Features**
  * Added document-scoped “presenceless” support via `disable_presence`/`WithDisablePresence()`: the server fixates the choice on first attach, the client gates presence locally, presence is stripped from updates, and snapshots return empty presence.
  * Added a new `PeekChannel` RPC endpoint with request/response schemas.
  * Extended attach/push-pull options with `disable_gc` where applicable.
* **Tests**
  * Added unit and integration coverage for presence stripping and cross-client server enforcement.
* **Documentation**
  * Added implementation planning documentation for the presenceless document option.

### DIFF
--- a/api/docs/yorkie/v1/yorkie.openapi.yaml
+++ b/api/docs/yorkie/v1/yorkie.openapi.yaml
@@ -130,6 +130,18 @@ paths:
           $ref: '#/components/responses/connect.error'
       tags:
       - yorkie.v1.YorkieService
+  /yorkie.v1.YorkieService/PeekChannel:
+    post:
+      description: ""
+      requestBody:
+        $ref: '#/components/requestBodies/yorkie.v1.YorkieService.PeekChannel.yorkie.v1.PeekChannelRequest'
+      responses:
+        "200":
+          $ref: '#/components/responses/yorkie.v1.YorkieService.PeekChannel.yorkie.v1.PeekChannelResponse'
+        default:
+          $ref: '#/components/responses/connect.error'
+      tags:
+      - yorkie.v1.YorkieService
   /yorkie.v1.YorkieService/PushPullChanges:
     post:
       description: ""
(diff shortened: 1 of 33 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #1849, opened 2026-07-01
Parallelize per-candidate Deactivate within a cycle

raararaara:

Dispatches per-candidate `Deactivate` calls across a bounded worker pool
(default 8) via a weighted semaphore instead of processing them
sequentially. Adds `Housekeeping.DeactivateConcurrency` and bumps the
default `CandidatesLimit` from 500 to 2000, raising the per-cycle cap
from `~17/s` to `~66/s`.

Each `Deactivate` operates on an independent `ClientRefKey` — its own
mongo read, its own optional cluster RPC fan-out, its own atomic
`$expr`-guarded update — so sequential dispatch is an implementation
choice, not a correctness requirement. Concurrent writes on distinct
`_id`s scale linearly under WiredTiger document-level locking; the
leader pod's CPU sits idle waiting on mongo I/O otherwise. The cluster
RPC path already fans out concurrently via `Backend.FanOut` — this
applies the same pattern to housekeeping.

Concurrency 8 balances the Go mongo driver's default max pool size
(100/host) against per-cycle wallclock reduction. Setting
`--housekeeping-deactivate-concurrency=1` restores the sequential loop;
`--housekeeping-candidates-limit=500` restores the previous batch size.

Fixes #

Alternatives considered:

| Alternative | Why not |
|---|---|
| Bulk `UpdateMany` for trivial candidates | Only covers the no-attached-docs subset; non-trivial candidates still need per-candidate cluster RPC fan-out. Orthogonal — can be layered later. |
| Cursor sharding into N ranges | Larger change with separate cursor state and range-filtered queries. Same net effect on throughput as bumping `CandidatesLimit`. |
| Reusing `Backend.FanOut` | Shaped around `map[K]V` with a server-wide cluster-RPC semaphore. Housekeeping dispatch has different bounds and benefits from its own knob. |

Observability: a registration-time log line lists `interval`,
`candidates_limit`, and `concurrency` so operators can verify the
dispatch mode without waiting for the first cycle. Each per-cycle
HSKP log now includes `cc=<N>` so sequential fallback is grep-able.

Rollback: `--housekeeping-deactivate-concurrency=1` restores the exact
sequential dispatch loop, and `--housekeeping-candidates-limit=500`
restores the previous batch size — both without a rebuild.

Yes.

```release-note
Housekeeping now dispatches per-candidate `Deactivate` calls across a
bounded worker pool. Introduces `--housekeeping-deactivate-concurrency`
(yaml: `Housekeeping.DeactivateConcurrency`, default 8). Bumps the
default `Housekeeping.CandidatesLimit` from 500 to 2000. Setting
`--housekeeping-deactivate-concurrency=0` or `1` falls back to the
exact sequential loop for legacy behavior.
```

**Additional documentation**:

```docs
```

**Checklist**:

- [ ] Didn't break anything

## Summary by CodeRabbit

* **New Features**
  * Added `Housekeeping.DeactivateConcurrency` and a CLI flag to cap the maximum concurrent client deactivations per housekeeping cycle (with `0`/`1` enabling sequential behavior).
  * Increased the default housekeeping candidate limit (500 → 2000) and introduced a default deactivation concurrency value (8).
* **Documentation**
  * Updated the sample configuration to include `DeactivateConcurrency`.
  * Startup/registration logs now include housekeeping interval, candidates limit, and deactivation concurrency.
* **Bug Fixes**
  * Negative concurrency values are now rejected; explicitly configured values (including `0`) are preserved.
* **Tests**
  * Added/extended tests for sequential vs parallel behavior, concurrency limits, cancellation behavior, and context handling.

### DIFF
--- a/cmd/yorkie/server.go
+++ b/cmd/yorkie/server.go
@@ -345,6 +345,12 @@ func init() {
 		"candidates limit for a single housekeeping run",
 	)
 	cmd.Flags().IntVar(
+		&conf.Housekeeping.DeactivateConcurrency,
+		"housekeeping-deactivate-concurrency",
+		server.DefaultHousekeepingDeactivateConcurrency,
+		"max concurrent client deactivations within a single housekeeping cycle (0 or 1 = sequential)",
+	)
+	cmd.Flags().IntVar(
 		&conf.Housekeeping.CompactionMinChanges,
 		"housekeeping-compaction-min-changes",
 		server.DefaultHousekeepingCompactionMinChanges,
--- a/docs/design/housekeeping.md
+++ b/docs/design/housekeeping.md
@@ -52,8 +52,12 @@ Housekeeping:
   # Interval between housekeeping runs (default: 30s)
   Interval: "30s"
 
-  # Maximum number of candidates to be returned in a single query (default: 500)
-  CandidatesLimit: 500
+  # Maximum number of candidates to be returned in a single query (default: 2000)
+  CandidatesLimit: 2000
+
+  # Maximum number of concurrent client deactivations within a single cycle.
+  # Values <= 1 fall back to sequential execution (default: 8)
+  DeactivateConcurrency: 8
 
   # Minimum number of changes required to compact a document (default: 1000)
   CompactionMinChanges: 1000
@@ -80,9 +84,19 @@ This approach is essential because:
 // Pseudocode for client deactivation
 for each project {
     candidates = FindInactiveClients(project, threshold: "24h")
+
+    // Dispatch per-candidate deactivations across up to
+    // DeactivateConcurrency goroutines within the cycle.
+    // Concurrency <= 1 falls back to sequential execution.
+    sem = NewWeighted(DeactivateConcurrency)
     for each candidate {
-        DeactivateClient(candidate)
+        sem.Acquire()
+        go {
+            defer sem.Release()
+            DeactivateClient(candidate)
+        }
     }
+    wait for all in-flight deactivations
 }
 ```
 
@@ -126,6 +140,7 @@ for each project {
 
 - **Interval**: Runs every 30 seconds by default
 - **Concurrency Control**: Uses distributed locking to prevent multiple instances from running simultaneously
+- **Parallel Deactivation**: Within a single cycle, up to `DeactivateConcurrency` client deactivations run in parallel (default: 8); values <= 1 fall back to sequential execution
 - **Project Cycling**: Processes projects in a round-robin fashion to distribute load
 - **Batching**: Limits candidates per project to prevent overwhelming the system
 
@@ -169,6 +184,7 @@ For detailed information about the garbage collection algorithm, refer to:
 Housekeeping:
   Interval: "10s"
   CandidatesLimit: 10
+  DeactivateConcurrency: 8
   CompactionMinChanges: 100
 ```
 
@@ -178,6 +194,7 @@ Housekeeping:
 Housekeeping:
   Interval: "1m"
   CandidatesLimit: 1000
+  DeactivateConcurrency: 8
   CompactionMinChanges: 5000
 ```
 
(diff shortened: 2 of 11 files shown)


hackerwins:
### Code review

Found 3 issues:

1. Negative `DeactivateConcurrency` in a YAML config is silently reset to 8, making the `Validate()` guard dead code for the file path. `ensureHouseKeepingDefaultValue` runs before `Validate()` (via `NewConfigFromFile`) and uses `< 0` as its sentinel, coercing e.g. `DeactivateConcurrency: -5` to 8 so the `< 0` check in `Validate()` never fires. Contrast `CandidatesLimit`, which uses `== 0` in ensure and `<= 0` in validate, so its validation actually works. Use `== 0` here (or drop the coercion and let `Validate()` own the rejection).

https://github.com/yorkie-team/yorkie/blob/d6a13ee3672594cde0600100cea575803967e201/server/config.go#L236-L242

Dead validation check:

https://github.com/yorkie-team/yorkie/blob/d6a13ee3672594cde0600100cea575803967e201/server/backend/housekeeping/config.go#L64-L69

2. The shutdown-safety cancel paths are unreachable in production. `dispatchDeactivate` breaks on `ctx.Err()` / `sem.Acquire(ctx, 1)` failure, and the doc comment claims "Parent ctx cancellation (typically a shutdown signal) still stops dispatching new candidates". But the scheduler task creates a fresh `ctx := context.Background()` per run, and `Stop()` (`StopJobs`/`Shutdown`) never cancels it — so the ctx is never cancelled and dispatch does not stop mid-cycle on shutdown. With `CandidatesLimit` raised to 2000, a cycle keeps dispatching regardless of a termination signal. Either wire a cancellable context into the task or drop the claim.

https://github.com/yorkie-team/yorkie/blob/d6a13ee3672594cde0600100cea575803967e201/server/backend/housekeeping/housekeeping.go#L59-L64

Comment asserting the behavior:

https://github.com/yorkie-team/yorkie/blob/d6a13ee3672594cde0600100cea575803967e201/server/clients/housekeeping.go#L95-L99

3. Behavior changed but the design doc was not updated (CLAUDE.md says design docs should be "keep updated when behavior changes"). `docs/design/housekeeping.md` still shows `CandidatesLimit: 500` (now 2000), has no mention of `DeactivateConcurrency` or parallel dispatch, and its pseudocode still shows a sequential `for each candidate { DeactivateClient(candidate) }` loop.

https://github.com/yorkie-team/yorkie/blob/d6a13ee3672594cde0600100cea575803967e201/docs/design/housekeeping.md#L53-L57

Outdated sequential pseudocode:

https://github.com/yorkie-team/yorkie/blob/d6a13ee3672594cde0600100cea575803967e201/docs/design/housekeeping.md#L81-L86

raararaara:
@hackerwins  Thanks for the review. Addressed all three points.

**1. Negative `DeactivateConcurrency` / dead `Validate()` check** — Dropped the coercion entirely (your second option). `newConfig()` pre-seeds the default before the YAML unmarshal, so an omitted field still resolves to the default without the ensure-coercion. Now an explicit `0` is preserved as a sequential opt-in and a negative value reaches `Validate()` and is rejected. (`== 0` would have coerced an explicit `0` back to the default and broken the sequential opt-in, which an existing test covers.) Added regression tests for the negative-rejection and zero-preservation paths.

**3. Design doc** — Updated `docs/design/housekeeping.md`: `CandidatesLimit` 500 → 2000, documented the `DeactivateConcurrency` knob, and replaced the sequential deactivation pseudocode with the parallel worker-pool dispatch.

**2. Unreachable shutdown-cancel path** — You're right; the task runs on `context.Background()` and `Stop()` never cancels it, so the cancel path can't fire. Rather than expand this PR, I dropped the claim: removed the `context.WithoutCancel` in-flight handling and the doc comment asserting graceful-shutdown behavior. The dispatch layer still honors a cancellable ctx (`ctx.Err` / `sem.Acquire`) as defensive, unit-tested code, so wiring a real cancellable context into the scheduler task is a clean follow-up — I left a `TODO` at that spot noting the wiring must also restore in-flight protection and audit the compaction / project-stats tasks that share the same path. Will track it separately.

hackerwins (approved):
Thanks for your contribution.

PR #1861, opened 2026-07-09
docs: add identity-preserving restore (GC-recreate) to undo-redo remaining options

harrykim8672:

Adds the identity-preserving restore approach to the Remaining Options table in undo-redo.md. The approach extends Approach B (selective un-tombstone) past its multi-client GC failure: the reverse op carries (createdAt, offset-range, content) spans, so purged nodes are recreated under their original IDs instead of resurrected in place. Overlapping undos then deduplicate by construction (idempotent restore).

## Summary by CodeRabbit

* **Documentation**
  * Added a new undo/redo option to the design notes, describing an identity-preserving restore approach for surviving and removed items.
  * Clarified the tradeoffs of this option, including GC safety and the need for convergence validation.


hackerwins (approved):
Thanks for your contribution.

PR #1234, opened 2025-04-16
Use Reactor ActorID as value

hackerwins:
**What this PR does / why we need it**:

Use Reactor ActorID as value

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:

- [ ] Added relevant tests or not required
- [ ] Addressed and resolved all CodeRabbit review comments
- [ ] Didn't break anything

PR #1864, opened 2026-07-11
Remove stale ActorID cache-priming comment in pubsub

cloneot:
`Publish()` had a comment and a no-op `event.Actor.String()` call explaining
that it primes ActorID's internal string cache to avoid a race on concurrent
access. That cache was removed when ActorID was refactored into a plain
`[12]byte` value type in #1234, so the call has been a no-op ever since —
it computes and discards a string with no side effect. Removing it as dead
code.

Fixes #

```release-note

```

**Additional documentation**:

```docs

```

**Checklist**:

- [X] Didn't break anything

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved event publishing reliability by removing unnecessary actor data access that could cause concurrency-related issues.

### DIFF
--- a/server/backend/pubsub/pubsub.go
+++ b/server/backend/pubsub/pubsub.go
@@ -150,10 +150,6 @@ func (m *PubSub) Publish(
 	publisherID time.ActorID,
 	event events.DocEvent,
 ) {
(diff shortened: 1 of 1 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #1874, opened 2026-07-15
Register GC pairs for pieces split off tombstoned nodes

### LINKED ISSUE #1873: Pieces split off a tombstoned tree node never register GC pairs and become unpurgeable
**What happened**:

Splitting an already-tombstoned text node — which happens when a concurrent delete's boundary lands inside another client's tombstone — creates pieces that are born dead: cloneText copies removedAt and split() registers the piece only into nodeMapByID via registerNode, but never into gcPairMap.

GC pairs are registered exclusively on live to dead transitions inside remove() (it returns true only for that transition). A node created dead never fires this trigger so the piece becomes a ghost tombstone node that incremental GC can neither purge nor count.

Identical logical state leads to identical purge results on both replicas. 

Running the following unit test confirms this issue:

1. Both replicas share `<p>hello</p>` (one text node).
2. Concurrently: d1 deletes "el" (splits the text while live at d1); d2 deletes the whole `<p>` (tombstones the text unsplit at d2).
3. Cross-sync; both replicas converge to <doc></doc>.
4. garbageCollect(maxVectorOf([A1, A2])) on both.

Observed:
``` 
AssertionError: asymmetric purge for identical state: d1=4 d2=2
```

**Environment**:
- Operating system: 
- Browser and version: 
- Yorkie version (use `yorkie version`): main
- Yorkie JS SDK version:

harrykim8672:

Fixes a garbage-collection leak around **tombstone splits**, the Go-server counterpart of the same fix in yorkie-js-sdk.

When a node is split, the new piece is cloned from the original — including `removedAt` (`SplitText`/`SplitElement` and `RGATreeSplitNode.split` all copy it). So a piece split off an already-tombstoned node is *born dead* without ever passing through `remove()`. GC pairs are only created on the live→dead transition, so no pair is ever registered for these pieces: they linger in the tree / RGA list forever, invisible to `GarbageLen()`, and a replica that splits a tombstone purges asymmetrically from one that split the node while it was still live.

The fix buffers a GC pair on the `Tree` / `RGATreeSplit` whenever a split produces a born-removed piece, and drains the buffer into the pairs returned by `Edit`/`Style`/`RemoveStyle` so operations register them with the root as usual. Covers both text and element splits (shared split path).

It also accounts `docSize` correctly. `RegisterGCPair` normally moves a child's size `Live`→`GC`, but a born-dead piece was never in `Live`, so the usual move drives `docSize.Live` negative and leaves `GC` residue after a full GC (measured: the split replica ended `Live.Data -8`, `GC.Data 8` vs `0/0` on its peer). `GCPair` gains an optional `GCOnlySize` holding the net-new size the split created; when set, `RegisterGCPair` adds it to `GC` and `AdjustDiffForGCPair` leaves `Live` untouched. Across the split's original node and its born-dead pieces the GC contributions telescope back to zero.

Unlike the JS SDK, Go already handles two related cases: `Tree.GCPairs` traverses tombstones (so the snapshot-load path was fine), and `RGATreeSplitNode.Remove` returns `false` on an LWW tombstone overwrite (so there is no toggle-unregistration bug). Those parts of the JS fix are not needed here.

#### Which issue(s) this PR fixes?

Fixes #1873

#### Does this PR introduce a user-facing change?

```release-note
Fix a garbage-collection leak where pieces split off a tombstoned Tree or Text node were never purged.
```

### Checklist
- [x] Didn't break anything

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed garbage-collection accounting for split pieces that were already tombstoned, including correct GC-only sizing and diff handling.
  * Ensured pending GC pairs are drained and propagated consistently for tree/text styling and split flows, including when operations return errors.
  * Reordered edit/style/remove-style execution so GC pair registration and diff application occur before error propagation.
* **Tests**
  * Added regression tests for concurrent tombstone + split GC behavior with replica synchronization (tree and text), verifying convergence and empty garbage state.
  * Added text edit error-path tests to confirm GC pairs are still returned and metadata diffs are preserved.

### DIFF
--- a/pkg/document/crdt/rga_tree_split.go
+++ b/pkg/document/crdt/rga_tree_split.go
@@ -321,6 +321,13 @@ type RGATreeSplit[V RGATreeSplitValue] struct {
 	initialHead *RGATreeSplitNode[V]
 	treeByIndex *splay.Tree[*RGATreeSplitNode[V]]
 	treeByID    *llrb.Tree[*RGATreeSplitNodeID, *RGATreeSplitNode[V]]
+
+	// pendingGCPairs buffers GC pairs for nodes that were created
+	// already-tombstoned by splitting a removed node. Such pieces inherit
+	// removedAt without ever passing through Remove(), so they would
+	// otherwise never be registered for GC. edit, Text.Style and
+	// Text.RemoveStyle drain this buffer into the GC pairs they return.
+	pendingGCPairs []GCPair
 }
 
 // NewRGATreeSplit creates a new instance of RGATreeSplit.
@@ -441,7 +448,32 @@ func (s *RGATreeSplit[V]) splitNode(
 	diff.Add(node.DataSize(), splitNode.DataSize())
 	diff.Sub(prevSize)
 
+	// NOTE: A piece split off an already-tombstoned node inherits
+	// removedAt without going through Remove(), so no GC pair is created
+	// for it in the normal deletion path. Buffer one here so it can be
+	// purged; otherwise it stays in the list forever. The piece was never
+	// live, so the net-new size the split created goes straight to
+	// docSize.GC when the pair is registered; return a zero diff to the
+	// caller (which accounts diffs to docSize.Live).
+	if splitNode.removedAt != nil {
+		gcSize := diff
+		s.pendingGCPairs = append(s.pendingGCPairs, GCPair{
+			Parent:     s,
+			Child:      splitNode,
+			GCOnlySize: &gcSize,
+		})
+		return splitNode, resource.DataSize{}, nil
+	}
+
 	return splitNode, diff, nil
+}
+
+// drainPendingGCPairs returns the GC pairs buffered for born-tombstoned
+// split pieces and clears the buffer.
+func (s *RGATreeSplit[V]) drainPendingGCPairs() []GCPair {
+	pairs := s.pendingGCPairs
+	s.pendingGCPairs = nil
+	return pairs
 }
 
 // InsertAfter inserts the given node after the given previous node.
@@ -541,6 +573,8 @@ func (s *RGATreeSplit[V]) edit(
 			Child:  removedNode,
 		})
 	}
+
+	pairs = append(pairs, s.drainPendingGCPairs()...)
 
 	return caretPos, pairs, diff, nil
 }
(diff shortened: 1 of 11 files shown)


harrykim8672:
Fixed CodeRabbit's comments in f74a391.

pkg/document/crdt/rga_tree_split.go
```diff
@@ -542,6 +574,8 @@ func (s *RGATreeSplit[V]) edit(
 		})
 	}
 
+	pairs = append(pairs, s.drainPendingGCPairs()...)
```
raararaara, line 578: If `findNodeWithSplit(from)` above returns an error, this drain is skipped — and the caller doesn't `RegisterGCPair` on the error path either. So a born-dead pair that the preceding `to` split already buffered never gets registered.

The split has already mutated the list, so that piece is left as an untracked tombstone — the same leak this PR fixes, re-surfacing whenever an operation errors after splitting a tombstone. Since the buffer is only drained on the success path, the cleanup is tied to the rest of the operation succeeding.

Registering the pair at split time (or returning it up the call chain instead of buffering it on the struct) would decouple the two. Same shape in `Text.Style/RemoveStyle` and the `Tree` paths.

harrykim8672: Good catch — fixed in ce4377c3. Went with the "return it up the call chain" option rather than registering at split time, since RGATreeSplit/Tree don't hold a *Root reference to register against directly.

Every error-return path in RGATreeSplit.edit, Text.Style/RemoveStyle, and Tree.Edit/Style/RemoveStyle now drains and returns whatever GC pairs already exist (including buffered split-buffer entries, and — in Tree.Edit specifically — any already-collected real remove-pairs from the delete phase, which had the same problem for mergeNodes/split/insert errors). The 4 callers in pkg/document/operations/ now register returned pairs before checking the error rather than after, so a partial-success pair list still gets registered even when the overall op fails.

Added a regression test in text_test.go that reproduces the exact scenario (to splits an already-tombstoned node and buffers a pair, from then fails) and confirms the pair is still returned — verified it fails without the fix.

hackerwins (approved):
Thanks for your contribution.

PR #1879, opened 2026-07-18
Set updated_at on project creation

### LINKED ISSUE #1848: `updated_at` is not set on project creation (remains zero value)
**What happened**:

When a project is created, `updated_at` is never initialized and stays at the Go zero value (`0001-01-01T00:00:00Z`) until the project is first updated via `UpdateProjectInfo`.

`NewProjectInfo` sets only `CreatedAt` (`server/backend/database/project_info.go`), and the MongoDB insert omits the `updated_at` key (`CreateProjectInfo` in `server/backend/database/mongo/client.go`).

The in-memory backend (`CreateProjectInfo` in `server/backend/database/memory/database.go`) inserts the same struct, so both backends are affected. Since `UpdatedAt` is serialized through `api/converter/to_pb.go` and `api/types/project.go` (`json:"updated_at"`), the bogus timestamp reaches the admin API and Dashboard.

Effect / Impact
- API/Dashboard shows a bogus timestamp: newly created projects report `updated_at = 0001-01-01T00:00:00Z` until they are edited once.
- Sorting by "recently updated" is broken: a brand-new project sorts as the oldest item (year 0001) instead of the newest, which is counterintuitive for any "last modified" ordering.
- Broken invariant: the common expectation `updated_at >= created_at` (and `updated_at == created_at` right after creation) does not hold. Any SDK/consumer relying on it can misbehave.
- Inconsistency: created_at is populated but updated_at is not, even though both fields exist on the model and are serialized.

On creation, `updated_at` should be set equal to `created_at`, consistent across both the MongoDB and in-memory backends.

1. Create a project (admin API `CreateProject` or `CreateProjectInfo` directly).
2. Read it back via the API / Dashboard, or inspect the stored document.
3. Observe `updated_at` is `0001-01-01T00:00:00Z` while `created_at` is set.

**Environment**:
- Operating system: N/A (server-side logic)
- Browser and version: N/A
- Yorkie version (use `yorkie version`): main (reproducible on current main)
- Yorkie JS SDK version: N/A

lemon0333:

When a project is created, `updated_at` was never initialized and stayed at the Go zero value (`0001-01-01T00:00:00Z`) until the project was first updated via `UpdateProjectInfo`. Admin API responses exposed this bogus timestamp for freshly created projects.

There were two separate causes:

1. `NewProjectInfo` (`server/backend/database/project_info.go`) only set `CreatedAt`.
2. The MongoDB insert in `CreateProjectInfo` (`server/backend/database/mongo/client.go`) omitted the `updated_at` key entirely, so the field was never persisted even if the constructor set it.

This PR initializes `CreatedAt` and `UpdatedAt` from the same timestamp so they are equal on creation (matching the existing doc-info convention), and adds the missing `updated_at` key to the MongoDB insert document.

Fixes #1848

- The memory backend stores the whole struct, so it only needs the constructor fix; MongoDB needs the additional insert key.
- The shared test suite (`RunUpdateProjectInfoTest`) now asserts `UpdatedAt` is non-zero right after creation, **including after re-fetching via `FindProjectInfoByID`** — the re-fetch assertion is what actually catches the missing MongoDB key. I verified it fails when the MongoDB insert fix is reverted.
- No migration is needed: existing rows keep their current value until the next update, same as before.

```release-note
Set updated_at when a project is created
```

**Additional documentation**:

```docs

```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

* **Bug Fixes**
  * Project records now reliably include an update timestamp when they’re first created.
  * Newly created projects set both creation and update timestamps to the same initial value.
  * Saved records preserve consistent timestamp values after reloading.
* **Tests**
  * Expanded update-project timestamp assertions to verify values are present and persist correctly.

### DIFF
--- a/server/backend/database/mongo/client.go
+++ b/server/backend/database/mongo/client.go
@@ -544,6 +544,7 @@ func (c *Client) CreateProjectInfo(
 		"public_key":                      info.PublicKey,
 		"secret_key":                      info.SecretKey,
 		"created_at":                      info.CreatedAt,
+		"updated_at":                      info.UpdatedAt,
 	})
 	if err != nil {
 		if mongo.IsDuplicateKeyError(err) {
--- a/server/backend/database/project_info.go
+++ b/server/backend/database/project_info.go
@@ -165,6 +165,7 @@ type ProjectInfo struct {
 
 // NewProjectInfo creates a new ProjectInfo of the given name.
 func NewProjectInfo(name string, owner types.ID) *ProjectInfo {
+	now := time.Now()
 	return &ProjectInfo{
 		Name:                        name,
 		Owner:                       owner,
@@ -187,7 +188,8 @@ func NewProjectInfo(name string, owner types.ID) *ProjectInfo {
 		AutoRevisionEnabled:         true,
 		PublicKey:                   shortuuid.New(),
 		SecretKey:                   shortuuid.New(),
-		CreatedAt:                   time.Now(),
+		CreatedAt:                   now,
+		UpdatedAt:                   now,
 	}
 }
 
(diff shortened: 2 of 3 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #1881, opened 2026-07-18
Migrate to sync.WaitGroup.Go for goroutine spawning

### LINKED ISSUE #1860: Migrate to `sync.WaitGroup.Go` for Goroutine Spawning

Replacement of the `Add(1) / go func() { defer Done() ... }()` idiom with Go 1.25's `sync.WaitGroup.Go` across the codebase, plus a lint note so new code follows the shorter form.

1. **Toolchain** — requires `go 1.25.0`+ in `go.mod`
2. **Mechanical rewrite** — file-by-file, no behavior change.
3. **Convention** — a one-liner in `CONTRIBUTING.md` pointing new goroutine spawns paired with a `WaitGroup` to `wg.Go`.

**How to migrate**:

Before:

```go
var wg sync.WaitGroup
for i := 0; i < n; i++ {
    wg.Add(1)
    go func(i int) {
        defer wg.Done()
        doWork(i)
    }(i)
}
wg.Wait()
```

After:

```go
var wg sync.WaitGroup
for i := 0; i < n; i++ {
    wg.Go(func() {
        doWork(i)
    })
}
wg.Wait()
```

- `wg.Go(f)` calls `wg.Add(1)`, spawns `f` in a new goroutine, and calls `wg.Done` when `f` returns.
- Removes the class of bugs where `Add` and `Done` are unbalanced (e.g., an early `return` before `Done` runs, or `Add` inside the goroutine).
- Candidate files can be from a grep of `sync.WaitGroup` in the tree.

**References**:
- [`sync.WaitGroup.Go` docs](https://pkg.go.dev/sync#WaitGroup.Go)
- [Go 1.25 release notes — `sync`](https://go.dev/doc/go1.25#sync)
- [Proposal: `sync.WaitGroup.Go`](https://github.com/golang/go/issues/63796)

mr8356:

Replaces the `wg.Add(1)` / `go func() { defer wg.Done(); ... }()` idiom with Go 1.25's `sync.WaitGroup.Go` across the codebase — a mechanical rewrite, no behavior change. `wg.Go(f)` calls `Add(1)`, spawns `f`, and calls `Done` when `f` returns, so it removes the class of bugs where `Add`/`Done` get unbalanced (an early return before `Done` runs, `Add` called inside the goroutine, etc.).

Also bumps `go.mod`, CI (`ci.yml`, `base-docker-publish.yml`), and the `Dockerfile` build image from Go 1.24 to 1.25, since `WaitGroup.Go` requires it — CI/Docker were still pinned to 1.24 and needed to move together with `go.mod`, not as a separate follow-up.

Reference: [`sync.WaitGroup.Go`](https://pkg.go.dev/sync#WaitGroup.Go) (added in go1.25.0) — *"Go calls f in a new goroutine and adds that task to the WaitGroup. When f returns, the task is removed from the WaitGroup."*

Fixes #1860

29 files touched (7 non-test, 22 test) across every `sync.WaitGroup` call site found by `grep -rl "sync.WaitGroup" --include="*.go" .`. Two variants came up beyond the plain swap:

- **Add/Done split across two functions** (e.g. a `Start()` method does `wg.Add(1); go x.loop(args)` while `loop` itself has `defer wg.Done()`): passed the method value straight to `wg.Go`, or wrapped in a closure, and dropped the now-orphaned `defer wg.Done()` at the far end. (`server/backend/membership/membership.go`, `pkg/limit/limiter.go`)
- **Closure took loop-var params** to dodge pre-1.22 capture bugs (e.g. `go func(gIdx int) {...}(g)`): identifiers are **never renamed** — both the outer loop var and the param name stay exactly as they were, and a single shadow-assignment line (`gIdx := g`) is added as the first statement of the new closure body. Only `wg.Add`, the param list, and `defer wg.Done()` are touched.
- **`Add(2)`/`Add(3)` immediately followed by 2 or 3 separate literal `go func(){ defer wg.Done(); ... }()` blocks** (each contributing exactly one `Done()`) converts cleanly to that many separate `wg.Go(...)` calls, dropping the batched `Add(n)`.

Deliberately **left untouched** (documented per-file in commit messages):
- `pkg/cache/cache_test.go` — both sites are `Add(n)` before a loop with a single loop-generated `Done()` call site, not N independent Add/Done units.
- `test/integration/doc_presence_test.go` — `Done()` fires conditionally from inside a `select` loop rather than via one deferred call at goroutine exit.
- Several `Add(n)`-before-a-loop sites in `pkg/cmap`, `pkg/trie`, `server/backend/channel` — same shape as the cache case above.

Not in scope: `errgroup.Group` usage (different type, per the issue).

```release-note
Bump minimum Go version to 1.25 (required for sync.WaitGroup.Go). Consumers building this module or its Go client package need Go 1.25 or later. action required
```

**Additional documentation**:

```docs
- [sync.WaitGroup.Go docs]: https://pkg.go.dev/sync#WaitGroup.Go
- [Go 1.25 release notes — sync]: https://go.dev/doc/go1.25#sync
```

**Checklist**:

- [x] Added relevant tests or not required — mechanical rewrite, existing tests exercise the same code paths; no new tests needed
- [x] Didn't break anything

## Test plan

Every commit was verified individually before being made (`gofmt -l`, `go build`, `go vet`, `golangci-lint run`, and `go test` scoped to the touched package/file — `-race` where the test suite made that practical) before moving to the next file/group. Where `golangci-lint` reported pre-existing issues unrelated to this change, a `git stash` A/B comparison confirmed zero new issues were introduced.

| Commit | Scope | Verification |
|---|---|---|
| `b8650560` | `go.mod`, `ci.yml`, `base-docker-publish.yml`, `Dockerfile` → Go 1.25 | `go build ./...`, `go vet ./...`, `make lint` (0 issues), `go test ./...` (full unit suite) |
| `2605156a` | `server/clients/housekeeping.go` | build/vet/gofmt, `go test ./server/clients/...` |
| `63d8348c` | `server/backend/background/background.go` | build/vet/gofmt (no test file for this package) |
| `79524bc3` | `server/backend/fanout.go` | build/vet/gofmt, `go test ./server/backend/...` |
| `dbb48bd4` | `server/backend/membership/membership.go` | build/vet/gofmt, `go test ./server/backend/membership/...` |
| `180cd10d` | `server/rpc/testcases/testcases.go` | build/vet `-tags integration`, `golangci-lint --build-tags integration` (0 issues) |
| `c4f04d99` | `client/client.go` | build/vet/gofmt, `golangci-lint`, `go test ./client/...` |
| `d95b6982` | `pkg/limit/limiter.go` | build/vet/gofmt, `golangci-lint`, `go test ./pkg/limit/...` |
| `1d594861` | `cluster/pool_test.go` | build/vet/gofmt, `golangci-lint` (0), `go test ./cluster/... -run TestClientPool` |
| `d2652c94` | `test/bench/*.go` (5 files) | build/vet/gofmt `-tags bench`, `golangci-lint --build-tags bench` (34 pre-existing issues, 0 new — stash-diff confirmed), `go test -tags bench -bench ...` incl. `-race` smoke run |
| `42bfb60b` | `server/backend/pubsub/pubsub_test.go` | build/vet/gofmt, `golangci-lint` (0), `go test ./server/backend/pubsub/... -v` |
| `cb0b9f7a` | `server/backend/membership/membership_test.go` | build/vet/gofmt, `golangci-lint` (0), `go test ./server/backend/membership/... -v` (incl. `TestLeadershipConcurrency`) |
| `24f76bf4` | `server/rpc/admin_server_unit_test.go` | build/vet/gofmt, `golangci-lint` (0), `go test ./server/rpc/... -run TestSemaphoreFanOutPattern -v` |
| `c8c25464` | `pkg/limit/limiter_test.go` | build/vet/gofmt, `golangci-lint` (0), `go test ./pkg/limit/... -v` |
| `70567710` | `pkg/locker/locker_test.go` | build/vet/gofmt, `golangci-lint` (0), `go test ./pkg/locker/... -v` |
| `e434bc83` | `pkg/cmap/cmap_test.go` | build/vet/gofmt, `golangci-lint` (0), `go test ./pkg/cmap/... -v` |
| `e26c5045` | `server/backend/channel/*.go` (2 files) | build/vet/gofmt, `golangci-lint` (0), `go test ./server/backend/channel/... -v` and `-race` |
| `3f5c2ee2` | `test/integration/*.go` (7 files) | build/vet `-tags integration`, `golangci-lint --build-tags integration` (70 pre-existing, 0 new — stash-diff confirmed), `go test -tags integration ./test/integration/... -v` (MongoDB up) |
| `fd7b0256` | `pkg/trie/*.go` (2 files) | build/vet/gofmt, `golangci-lint` (0), `go test ./pkg/trie/... -race` |
| `773ed7af` | `CONTRIBUTING.md` | docs-only |

Final pass across the full branch diff before opening this PR:
- `make lint` → **0 issues**
- `golangci-lint --build-tags integration ./...` and `--build-tags bench ./...` → same pre-existing counts as `main`, **0 new**
- `go build ./...`, `go vet ./...` → clean
- `go test ./...` → all packages pass

Not run locally: the full-repo `make test` (`go test -tags integration -race ./...`, no test-case filter). Every touched package was already exercised individually with `-tags integration`/`-race` per the table above; the untouched remainder of the suite is left to CI rather than re-running the whole thing here.

## Summary by CodeRabbit

* **Build & Compatibility**
  * Updated the required Go toolchain version to **1.25** for CI and Docker builds.
* **Reliability**
  * Standardized concurrent execution to use **`sync.WaitGroup.Go`** for safer, consistent goroutine lifecycle tracking across background processes, core components, and concurrency-heavy tests/benchmarks.
* **Documentation**
  * Added guidance to **CONTRIBUTING** and an archived task page describing the recommended **`sync.WaitGroup.Go`** migration approach and conventions.


test/integration/clusternodes_test.go
```diff
@@ -287,11 +287,10 @@ func TestClusterNodes(t *testing.T) {
...
 		var wg sync.WaitGroup
 		for _, svr := range svrs {
-			wg.Add(1)
-			go func(svr *server.Yorkie) {
-				defer wg.Done()
+			wg.Go(func() {
+				svr := svr
 				require.NoError(t, svr.Start())
-			}(svr)
+			})
```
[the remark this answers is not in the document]
mr8356: Good catch on both points — thanks for the review.

On the svr := svr shadow line: this is intentional, not an oversight. This PR is a mechanical rewrite across 22 test files replacing the wg.Add(1) / go func(param) { defer wg.Done(); ... }(arg) idiom with wg.Go. To keep every site a pure, reviewable diff with zero identifier renaming, the convention used throughout is: when the closure took a parameter to capture a loop variable (the old pre-1.22 capture-safety idiom), drop the parameter but shadow-assign it as the first line of the new closure body (svr := svr) instead of renaming the outer loop variable or the body's references to the param. You're right that it's redundant under Go 1.22+'s per-iteration loop semantics — but removing it here would make this one site inconsistent with the other ~20 sites using the same pattern in this PR, so I'd rather leave it as-is for consistency within this PR. Happy to do a follow-up cleanup pass removing these shadow lines repo-wide if that's useful, but treating it as out of scope here.

On the require.NoError inside the goroutine — good catch, that's a real pre-existing issue (not introduced by this PR, the original code already called require.NoError inside the spawned goroutine; this rewrite only replaced Add/go func/defer Done() with wg.Go). Fixing it in a follow-up commit on this PR since it's a trivial, adjacent safety fix.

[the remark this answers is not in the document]
ggyuchive: Hi @mr8356  , I didn't get it. Why did you add `svr := svr`?
Please check https://github.com/yorkie-team/yorkie/issues/1860 code example again.

pkg/locker/locker_test.go
```diff
@@ -283,8 +281,8 @@ func TestRWLockerConcurrency(t *testing.T) {
 
 	var wg sync.WaitGroup
 	for i := 0; i <= 1000; i++ {
-		wg.Add(1)
-		go func(i int) {
+		wg.Go(func() {
+			i := i
```
ggyuchive: Same in here.

docs/tasks/archive/2026/07/20260718-waitgroup-go-migration-todo.md
ggyuchive, line 1: Please move to archive if task is done.

pkg/locker/locker_test.go
mr8356: It's to prevent a loop variable capture bug (race condition).

wg.Go uses a parameterless closure, so it captures the reference of i. Without i := i, all goroutines share the exact same i variable and read the final iteration value. i := i creates a fresh variable for each iteration so every goroutine gets its own value.

(Note: Since we are on Go 1.22+, I kept this as defensive programming, but loop variables are scoped per iteration by default now. Let me know if you prefer to remove it!)

ggyuchive: Thanks for checking @mr8356 !
Since it's unnecessary code and not the best practice of Go 1.25, so it will be great to delete - also for good convention of our repository.

```go
// Part of https://pkg.go.dev/sync#example-WaitGroup.Go
func main() {
	var wg sync.WaitGroup
	var urls = []string{
		"http://www.golang.org/",
		"http://www.google.com/",
		"http://www.example.com/",
	}
	for _, url := range urls {
		// Launch a goroutine to fetch the URL.
		wg.Go(func() {
			// Fetch the URL.
			http.Get(url)
		})
	}
	// Wait for all HTTP fetches to complete.
	wg.Wait()
}
```

I checked 2 of `x := x` pattern but there are many `x := y` pattern from your code. It can be also changed.
It will be great to push a commit to this branch for removing them.

mr8356: I removed all 61 loop-var shadow copies from the wg.Go closures and re-adopted the outer loop variable directly. thanks for the review! @ggyuchive

test/integration/channel_test.go
```diff
@@ -996,15 +996,13 @@ func TestChannelIntegration(t *testing.T) {
...
+			wg.Go(func() {
 				var err error
-				channels[idx], err = channel.New(key.Key(channelKeys[idx]))
+				channels[i], err = channel.New(key.Key(channelKeys[i]))
 				assert.NoError(t, err)
-				err = clients[idx].Attach(ctx, channels[idx])
+				err = clients[i].Attach(ctx, channels[i])
 				assert.NoError(t, err)
-			}(i)
+			})
```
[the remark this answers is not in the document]
ggyuchive: We've discussed earlier with official go docs.

Since it's unnecessary code and not the best practice of Go 1.25, so it will be great to delete - also for good convention of our repository.
``` go
// Part of https://pkg.go.dev/sync#example-WaitGroup.Go
func main() {
	var wg sync.WaitGroup
	var urls = []string{
		"http://www.golang.org/",
		"http://www.google.com/",
		"http://www.example.com/",
	}
	for _, url := range urls {
		// Launch a goroutine to fetch the URL.
		wg.Go(func() {
			// Fetch the URL.
			http.Get(url)
		})
	}
	// Wait for all HTTP fetches to complete.
	wg.Wait()
}
```

ggyuchive (approved):
Thanks for your contribution. It looks good to me.
Hi @hackerwins , this PR uses Go 1.25's `wg.Go(f func())` instead of manual
`wg.Add()` / `defer wg.Done()`, reducing line count and preventing goroutine
mistakes (Add/Done count mismatch, early return before `defer`, ...) later.

Is it okay to merge?

hackerwins (approved):
Thanks for your contribution.

Issue #1858, opened 2026-07-07
Add Fuzz Test Infrastructure in Yorkie


ggyuchive:
**What would you like to be added**:

A fuzzing suite for Yorkie's decoders, with three layers:

1. **Fuzz targets**: In a single `test/fuzz/` package (like `test/bench`: `//go:build fuzz`, `package fuzz`). Each target feeds random bytes to a decoder and asserts it never panics.
2. **Regression gate in CI**: The saved corpus runs on every PR (`go test -tags fuzz ./test/fuzz/`), so a fixed crash can never come back.
3. **Nightly continuous fuzzing**: A scheduled job runs `-fuzz` for a time budget; a new crash fails the job and is reported as an issue.

**How to add a fuzz target**

```go
//go:build fuzz

package fuzz

func FuzzBytesToObject(f *testing.F) {
    f.Add([]byte{})                 // seed corpus: a few example inputs
    f.Fuzz(func(t *testing.T, data []byte) {
        // Returning an error is fine; a panic is a bug and fails the run.
        _, _ = converter.BytesToObject(data)
    })
}
```

- Target any function that parses **untrusted input** (bytes/strings a client can send). Assert the invariant "never panics"; add stronger properties where they hold (e.g. round-trip `decode(encode(x)) == x`, or "no empty output").
- Run one target locally: `go test -tags fuzz -run=^$ -fuzz=FuzzBytesToObject ./test/fuzz/`
- If it finds a crash, Go writes the input to `test/fuzz/testdata/fuzz/<Target>/`.

**Candidate targets**
- [X] https://github.com/yorkie-team/yorkie/pull/1890
- [X] https://github.com/yorkie-team/yorkie/pull/1892
- [X] https://github.com/yorkie-team/yorkie/pull/1900
- [x] https://github.com/yorkie-team/yorkie/pull/1912
- [ ] https://github.com/yorkie-team/yorkie/pull/1920
- [ ] `converter.BytesToObject` / `BytesToArray` / `BytesToTree` / `BytesToSnapshot`
- [ ] `crdt.ValueFromBytes`
- [ ] `time.VersionVectorFromBytes`, `time.ActorIDFromHex`
- [ ] `channel.ParseKeyPath`

**Fuzz Run Infra in yorkie**
- [ ] Run corpus tests in CI
- [ ] Run continuous Fuzz tests in CI

**Why is this needed**:

[...]

PR #1890, opened 2026-07-21
Add fuzz coverage for ChangePack deserialization

snghyn28:

This PR adds fuzz coverage for `converter.FromChangePack` and includes regression corpus for malformed ChangePack payloads that previously caused panics during conversion.

Before adding the broader fuzz infrastructure, I started by running fuzz tests locally and splitting the discovered issues by fuzz target. This PR is intentionally scoped to the `FuzzFromChangePack` target so the test coverage, failing corpus, and corresponding server-side fix can be reviewed together.

The fix checks for missing fields inside ChangePack messages and returns errors instead of panicking on malformed client input.

Refs #1858

This PR is scoped to `FuzzFromChangePack` only. Other fuzz targets and CI/nightly fuzz workflow changes will be split into follow-up PRs.

```release-note
NONE
```

**Additional documentation**:

```docs

```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation during protobuf-to-internal conversion for changes, operations, and document structures.
  * Conversion now fails fast with descriptive errors on missing/invalid data (including nil inputs) rather than risking crashes or using empty defaults.
  * Presence conversion now reports errors when required payload content is missing.

* **Tests**
  * Added a fuzz target for change-pack conversion with new seeds and corpus entries.
  * Updated tests to assert conversion error handling explicitly.


hackerwins (approved):
Thanks for your contribution.

PR #1905, opened 2026-08-02
Move tombstones with merge to preserve tree RGA anchors

### LINKED ISSUE #1904: Concurrent `Tree.Edit` insertion into a removed range diverges between replicas
**What happened**:

Two replicas of the same document ended up in different states after a full sync, when one client inserted an element into a range that the other client concurrently removed.

- client A: `Tree.Edit(6, 6, <p>, 0)` - inserts an empty `<p>` between `c` and `d`
- client B: `Tree.Edit(0, 6, nil, 0)` - concurrently removes the first paragraph, the second paragraph's opening token, and `c`

Final states after both clients synced in both directions:
- client A: `{"type":"r","children":[{"type":"p","children":[]},{"type":"text","value":"d"}]}` - its inserted `<p>` survives, hoisted to the root
- client B: `{"type":"r","children":[{"type":"text","value":"d"}]}` - the inserted `<p>` is dropped

Both replicas converge to the same tree regardless of the order in which the concurrent operations are delivered.

```go
func TestTreeDivergenceRepro(t *testing.T) {

      t.Run("concurrent element insert into removed range", func(t *testing.T) {
              ctx := context.Background()

                              Type: "r",
                              Children: []json.TreeNode{
                              },
                      })
                      return nil
              }))

              // Concurrent: d1 inserts an empty <p> between c|d, d2 removes [0, 6).
                      root.GetTree("t").Edit(6, 6, &json.TreeNode{Type: "p", Children: []json.TreeNode{}}, 0)
                      return nil
              }))
                      root.GetTree("t").Edit(0, 6, nil, 0)
                      return nil
              }))

      })
}
```

- Originally found by the property-based Tree test for yorkie-team/yorkie-js-sdk#1298; the JS SDK diverges identically (yorkie-team/yorkie-js-sdk#1302), so this appears to live in the shared `Tree.Edit` semantics.
- Related to #1862 - property-based convergence tests would catch this class of bug on the Go side.

**Environment**:

- Browser and version: N/A 
- Yorkie version (use `yorkie version`): 0.7.13
- Yorkie JS SDK version: 0.7.13 - same divergence (yorkie-team/yorkie-js-sdk#1302)

hackerwins:
## Summary

A concurrent `Tree.Edit` that inserts into a range concurrently removed by a
merge diverged between replicas (yorkie-team/yorkie-js-sdk#1302).

Starting from `<r><p>ab</p><p>cd</p></r>`, with client A inserting an empty
`<p>` at index 6 and client B removing `[0,6)`:

- A (insert then remove): `<r><p></p>d</r>`
- B (remove then insert): `<r>d</r>` — the inserted node is dropped

**Root cause:** the merge moved only the *live* children of the merged
paragraph to the target and left the insert's RGA anchor (`c`) behind as a
tombstone in the removed parent. A late concurrent insert then resolved
against the removed parent and was tombstoned by the insert guard. With
several clients inserting at the same anchor, even the survivors ordered
inconsistently, because the merging replica never ran the RGA `CreatedAt`
tie-break for them.

**Fix (in the merge, not the resolver):** move tombstoned children with the
merge too, in child order, so the source's full RGA sequence is preserved in
the target. A late insert then resolves in the target through the normal
path and is ordered by the existing step-04 RGA tie-break, converging with
the replica that applied the insert before the merge — including the
multi-client same-anchor case.

The relocation uses a new `index.Node.MoveChild` that keeps both length
dimensions correct on both parents: a tombstone occupies no visible index
positions, so only `TotalLength` moves for a removed node and `VisibleLength`
is left untouched on source and target alike. (Reusing the alive-node
`DetachChild`/`Append` pair double-counts a tombstone's visible length.)
`FindTreeNodesWithSplitText` is unchanged (comment only); the shared position
resolver is not touched.

Design: `docs/design/concurrent-merge-split.md` §6.1 (Fix 19).
Fixes #1904

## Test plan

- New complex convergence tests (all converge):
  - `TestTreeConcurrencyInsertIntoRemovedRange` — the filed 2-client case
  - `TestTreeConcurrencyInsertIntoRemovedRangeMultiInsert` — three clients,
    two inserts at the same anchor into the removed range
- New unit test `TestTreeEdit/merge moves an element tombstone with correct
  length accounting` — guards the element-padding accounting of the move
- Regression (all green, no new skips/failures):
  - `go test -tags complex ./test/complex/ -run TestTreeConcurrency` — 1599 pass
  - `go test -tags integration ./test/integration/ -run 'Tree|GC|Snapshot'` — 169 pass
  - `go test ./...` — whole-repo unit tests pass
  - `gofmt` clean, `golangci-lint` 0 issues

## Follow-up

- Port the same merge change to `yorkie-js-sdk`
  (`packages/sdk/src/document/crdt/tree.ts`), which mirrors this structure.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved convergence for concurrent inserts into paragraphs removed during merges.
  * Preserved ordering and placement of nested content, including deleted content, during paragraph merges.
  * Corrected visible and structural length tracking when moving content between merged paragraphs.

* **Tests**
  * Added regression coverage for two- and three-replica concurrent editing scenarios.
  * Added coverage for preserving deleted nested elements during merges.

* **Documentation**
  * Documented merge behavior, ordering guarantees, troubleshooting lessons, and follow-up tasks.

### DIFF
--- a/docs/design/concurrent-merge-split.md
+++ b/docs/design/concurrent-merge-split.md
@@ -74,6 +74,11 @@ execution and rebuilt from the persisted `MergedFrom` field on
 snapshot load via `rebuildMergeState`. This enables a fast nil-check
 on the hot path without persisting a separate field.
 
+This redirect fires only for a *left-most* position, whose anchor is
+the tombstoned parent itself. A position with a left-sibling anchor
+resolves through the normal path, because the merge moves the sibling
+(live or tombstone) into the target — see §6.1.
+
 ### §1.2 Inverted Range Guard
 
 **Function**: `traverseInPosRange`
@@ -186,6 +191,26 @@ merge-target (`fromParent`). Each moved child receives:
   merge timestamp must be stable for version-vector checks.
 - **`mergedInto`** (runtime cache): set on the source node, pointing
   to the target. Rebuilt from `MergedFrom` on snapshot load.
+
+Both live and **tombstoned** children are moved, in child order
+(`collectBetween` collects them with `Children(true)`). Moving the
+tombstones preserves the source's full RGA sequence in the target, so
+a concurrent insert anchored on a removed sibling (e.g. `c` in a
+`cd → c,d` split whose `c` was deleted) resolves in the target through
+the normal path and is ordered against the other concurrent inserts by
+the RGA `CreatedAt` tie-break (`FindTreeNodesWithSplitText` step 04).
+This is what makes the merging replica converge with a replica that
+applied the insert *before* the merge, including when several clients
+insert at the same anchor.
+
+The relocation uses `index.Node.MoveChild`, which keeps both length
+dimensions correct on **both** parents. A tombstone occupies no visible
+index positions — `remove()` already debited its `VisibleLength` from
+its ancestors — so for a removed node `MoveChild` relocates only the
+include-removed `TotalLength` and leaves `VisibleLength` untouched on
+source and target alike. (Reusing the alive-node `DetachChild`/`Append`
+pair would double-count the tombstone's visible length on the target and
+under-count it on the source.)
 
 ### §6.2 Delete Propagation to Merge-Moved Children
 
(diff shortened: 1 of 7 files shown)


hackerwins:
@coderabbitai review this.

PR #1909, opened 2026-08-03
Keep style ranges from crossing a concurrent merge anchor

### LINKED ISSUE #1908: Concurrent `Tree.Style` over a removed range styles a concurrently inserted node on only one replica
**What happened**:

After #1905 and #1906, the structural divergence from yorkie-team/yorkie-js-sdk#1302 is fixed, but the same op family now diverges on attributes.

- client A: `Tree.Edit(8, 8, <p>, 0)` - inserts an empty `<p>` at the end, then `Tree.Style(0, 5, {"bold": "x"})` - styles the first paragraph and the second paragraph's opening region
- client B: `Tree.Edit(0, 5, nil, 0)` - concurrently removes the first paragraph and the second paragraph's opening token (paragraph-merging path)

Final states after both clients synced in both directions - the structure converges on both replicas, but:

- client A: `{"type":"p","children":[]}` - the surviving inserted `<p>` has no attributes
- client B: `{"type":"p","children":[],"attributes":{"bold":"x"}}` - the same `<p>` is styled

Both replicas converge to the same tree, including attributes, regardless of the order in which the concurrent operations are delivered.

```go
func TestTreeStyleDivergenceRepro(t *testing.T) {

	t.Run("concurrent style range overlapping merge-removed range", func(t *testing.T) {
		ctx := context.Background()

				Type: "r",
				Children: []json.TreeNode{
				},
			})
			return nil
		}))

		// d1: insert an empty <p> at the end, then style [0, 5).
			root.GetTree("t").Edit(8, 8, &json.TreeNode{Type: "p", Children: []json.TreeNode{}}, 0)
			return nil
		}))
			root.GetTree("t").Style(0, 5, map[string]string{"bold": "x"})
			return nil
		}))
		// d2: concurrently remove [0, 5) - merges across the paragraphs.
			return nil
		}))

	})
}
```

- Follow-up to yorkie-team/yorkie-js-sdk#1302 in the same op family: found by the property-based Tree test for yorkie-team/yorkie-js-sdk#1298 on the first run after the fixes landed; the JS SDK diverges identically (https://github.com/yorkie-team/yorkie-js-sdk/issues/1308).
- The style operation resolves to different node sets on the two replicas: client A styles the two (later removed) paragraphs, while on client B the concurrently inserted `<p>` ends up inside the style's anchor range after the merge moved tombstones, so the remote style also covers it.
- `RemoveStyle` diverges the same way, leaving an empty `attributes: {}` on one replica only.
- Without the concurrent insert, the same style/delete combination converges.

**Environment**:
- Browser and version: N/A
- Yorkie version (use `yorkie version`): main (includes #1905 and #1906)
- Yorkie JS SDK version: main - same divergence (yorkie-team/yorkie-js-sdk#1308)

Nahee-Park:

Fixes a Tree attribute-divergence: a concurrent `Tree.Style`/`Tree.RemoveStyle` whose range ends inside an element removed by a merge styled a node concurrently inserted at the merged anchor on only one replica.

From `<r><p>ab</p><p>cd</p></r>`, with client A inserting an empty `<p>` at index 8 and then styling `[0,5)` with `bold=x`, and client B concurrently removing `[0,5)`:

- A: the surviving inserted `<p>` has no attributes
- B: the same `<p>` carries `bold="x"` - the replicas diverge on attributes (the structure itself converges since #1905,#1906)

This happens because `FindTreeNodesWithSplitText` redirects a position inside a merged-away parent to the insertion boundary in the merge target, which sits before the first moved child. For an inserting edit this is intentional (the RGA `CreatedAt` tie-break needs that position), but for a style range end it extends the range over nodes concurrently inserted between the merge-source tombstone and the moved children - nodes the styling client saw outside its range, after the then-live parent.

This PR adds an optional boundary mode to `FindTreeNodesWithSplitText`. The default `BoundaryInsert` keeps the existing behavior for edits; `BoundaryRange`, used by `Style`/`RemoveStyle`, resolves the position right after the merge-source tombstone - before both the concurrent inserts and the moved children in visible coordinates, so a range end excludes them and a range start includes them, matching the styling replica in both directions. Ranges that genuinely cover the merged content still style it.

Tests:

- `test/complex/tree_concurrency_test.go` (new): `TestTreeConcurrencyStyleAcrossMergedAnchor` (fails without the fix), the `RemoveStyle` variant, a positive case where the range genuinely covers the merged content, and a chained-merge case asserting the exact converged value
- `pkg/document/crdt/tree_test.go` (new case): pins the boundary indexes for both modes
- `go test ./pkg/document/...`, `-tags complex` `TestTreeConcurrency*`, `-tags integration` `TestTree*` all green; `golangci-lint run ./...` reports 0 issues

Fixes #1908

- Found by the property-based Tree test for yorkie-team/yorkie-js-sdk#1298 as the pinned seed's next shrunk counterexample after #1905/#1906 landed; the JS SDK diverges identically (yorkie-team/yorkie-js-sdk#1308). A mirror PR for yorkie-js-sdk follows.
- Design: `docs/design/concurrent-merge-split.md` §9.3 (Fix 21), added in this PR.
- A range end anchored after a moved child (a non-left-most position in the merged parent) does not go through the redirect and remains a separate follow-up; the property-based test shrinks to it once this fix lands.

```release-note
Fix Tree.Style and Tree.RemoveStyle applying to a node concurrently inserted into a merge-removed range, which diverged replicas on attributes.
```

**Additional documentation**:

```docs
- [Design]: docs/design/concurrent-merge-split.md §9.3 (Fix 21)
```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrent styling and style removal across merged tree content.
  * Prevented incorrectly applying styles to concurrently inserted or moved content.
  * Added support for chained merge scenarios while preserving consistent replica results.

* **Documentation**
  * Documented merge-boundary behavior, known limitations, and recommended handling.

* **Tests**
  * Added regression and concurrency coverage for styling, style removal, and merged content boundaries.

### DIFF
--- a/docs/design/concurrent-merge-split.md
+++ b/docs/design/concurrent-merge-split.md
@@ -419,6 +419,30 @@ attribute deep-copy in §7.2 — causing divergence.
 Helper: `ticketKnown(vv, ticket)` — reused for the unknown-sibling
 check.
 
+### §9.3 Range Boundary at Merged-Away Anchors
+
+A style position anchored at the left-most position inside a parent
+that a concurrent merge removed must not resolve to the insertion
+boundary of §1.1. The insertion boundary sits before the first child
+moved by the merge, so nodes concurrently inserted between the
+merge-source tombstone and the moved children fall inside the styled
+range on the applying replica — while the styling replica saw them
+outside its range, after the then-live parent. Replicas then diverge
+on attributes.
+
+`FindTreeNodesWithSplitText` takes a boundary mode: `BoundaryInsert`
+(default, §1.1 behavior for edits) and `BoundaryRange`, used by
+`Style`/`RemoveStyle`, which resolves the position to right after the
+merge-source tombstone. In visible coordinates this lands before both
+the concurrent inserts and the moved children, so a range end excludes
+them and a range start includes them — matching the styling replica in
+both directions. Ranges that genuinely cover the merged content
+(anchored outside the merged parent) are unaffected.
+
+A range end anchored after a moved child (a non-left-most position in
+the merged parent) does not go through the §1.1 redirect and can still
+diverge; this remains a follow-up.
+
 ## Key Design Decisions
 
 | Decision | Reason |
@@ -496,3 +520,4 @@ For traceability from git history (commit messages reference Fix N).
 | Fix 18 | §3 | Cross-parent range narrowing |
 | Fix 19 | §6.1 | Move tombstones with merge to preserve RGA anchors |
 | Fix 20 | §6.3 | Flatten chained merge (P→Q→R) for redirect + snapshot consistency |
+| Fix 21 | §9.3 | Style range boundary right after merge-source tombstone |
--- a/docs/tasks/active/20260803-tree-style-merge-anchor-lessons.md
+++ b/docs/tasks/active/20260803-tree-style-merge-anchor-lessons.md
@@ -1,0 +1,17 @@
+# Lessons: tree style merge-anchor boundary
+
+**Created**: 2026-08-03
+
+- The §1.1 insertion-boundary redirect is correct for inserts (RGA
+  tie-break needs the position before the first moved child) but wrong
+  for range boundaries: one position semantics cannot serve both, so
+  the resolver takes an explicit boundary mode.
+- The "right after the merge-source tombstone" position fixes both
+  range directions at once: in visible coordinates it lands before the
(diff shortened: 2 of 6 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #1910, opened 2026-08-04
Validate Checkpoint in ChangePack for PushPull API requests

### LINKED ISSUE #805: Validate Checkpoint in ChangePack for PushPull API requests

YBSeok:

This PR adds request-level validation to `PushPull` before any database mutation. It rejects malformed checkpoint state when the request carries a future `ServerSeq`, and it also rejects non-sequential `ClientSeq` values within the request `Changes`.

Duplicate retries are still handled idempotently by skipping already-known client sequences during the push path, so retried requests do not re-apply the same change. For all other malformed request cases, the RPC now surfaces `INVALID_ARGUMENT` cleanly.

Fixes #805

This change keeps the validation at the `PushPull` request boundary so the server rejects tampered or out-of-order checkpoints before they reach the write path. The existing `pushPack` behavior remains responsible for ignoring duplicate client sequences already present in the client's checkpoint.

NONE

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid synchronization requests are now rejected when client sequence numbers are non-contiguous or server checkpoints are ahead of the current state.
  * These validation failures now return consistent invalid-request errors without advancing checkpoints.
  * Existing detailed connection errors are preserved instead of being replaced by generic errors.
  * Deactivating a client with an attached document now correctly detaches it and records the presence change.

* **Tests**
  * Added coverage for invalid sequence numbers, future server checkpoints, and client deactivation with attached documents.


hackerwins (approved):
Thanks for your contribution.

PR #1912, opened 2026-08-05
Add fuzz coverage for JSON object and array decoding

### LINKED ISSUE #1911: Add fuzz coverage for converter decoders (BytesToObject/Array/Tree/Snapshot)
**Parent roadmap**: #1858

Add fuzz coverage for the `api/converter` decoders, one of the remaining unchecked targets in #1858:

- `converter.BytesToObject`
- `converter.BytesToArray`
- `converter.BytesToTree`
- `converter.BytesToSnapshot`

These decode client-supplied protobuf (the most central untrusted-input path on the server), so they're the highest-value fuzz targets. Each target feeds random bytes to the decoder and asserts it never panics (returning an error is fine).

**Plan** (following the maintainer's guidance of 1–2 targets per PR):
- PR 1: `FuzzBytesToObject`, `FuzzBytesToArray`
- PR 2: `FuzzBytesToTree`, `FuzzBytesToSnapshot`

All targets go into the `test/fuzz/` package (`//go:build fuzz`, `package fuzz`) as described in #1858.

I'd like to take this — could you please assign it to me? I'll align the CI/workflow parts on Discord if needed before touching them.

lemon0333:

Adds fuzz coverage for `converter.BytesToObject` and `converter.BytesToArray`, which decode client-controlled snapshot bytes.

`proto.Unmarshal` succeeds on many malformed inputs, leaving a message with nil or incomplete fields that the converter then dereferences and panics on. Fuzzing surfaced three such nil-pointer panics, so this PR guards the decode path to return errors instead of panicking:

- `fromJSONElement`: nil element
- `fromJSONObject` / `fromJSONArray`: nil oneof body
- reject object members / array elements whose `createdAt` is nil (the ticket is later dereferenced when keying the container)

Includes a regression corpus for each guarded case, in the same `go test fuzz v1` format as the existing targets. As with the previous fuzz PRs, this only touches `test/fuzz/` and the source fix (`//go:build fuzz` keeps the targets out of the normal `go test ./...` run); the CI/nightly fuzz workflow is left for a follow-up.

Fixes #1911

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation when converting serialized data into objects and arrays.
  * Invalid null values, malformed structures, and missing creation timestamps are now rejected with clear errors.

* **Tests**
  * Added fuzz testing for object and array deserialization.
  * Added coverage for malformed payloads and missing element metadata.


lemon0333:
Thanks @coderabbitai — good catch. Addressed in cee1c7f5.

Rather than asserting inside the fuzz callbacks (which only run under `-tags fuzz`), I added deterministic error-contract tests in `api/converter/from_bytes_test.go` that run in the normal test suite. `TestBytesToObjectMalformed` / `TestBytesToArrayMalformed` assert that `BytesToObject`/`BytesToArray` return a non-nil error (and a nil result) for each guarded case — mismatched oneof body, nil member element, and nil `createdAt` — so a future change that silently accepts them fails CI. The fuzz targets stay as arbitrary-input no-panic coverage.

lemon0333:
Thanks @coderabbitai — added in the latest commit. `TestBytesToObjectMalformed` now also covers a member whose `Element` is an empty `&api.JSONElement{}` (nil oneof body), and `TestBytesToArrayMalformed` covers both an `RGANode{Element: nil}` (no position timestamps) and an element with a nil oneof body. Each asserts a non-nil error and a nil result, so every guard on the object/array decode path is exercised.

PR #1913, opened 2026-08-06
Recreate purged text runs in order on restore, not reversed

harrykim8672:
## Summary

Identity-preserving Text restore reverses a deleted run when undoing **after the run's tombstones were GC-purged**. Reported via wafflebase notes undo: type "hello my name is", delete "my name", undo → "hello **eman ym** is". A page refresh showed the correct text because the server still had the tombstones and un-tombstoned them, while the client had already GC'd and took the recreate path.

**Root cause.** Typing character-by-character makes each character its own single-character insertion (distinct `createdAt`, offset 0). When the run is purged, `restore` must recreate each fragment. In `findRestoreAnchor` the same-insertion rungs (a/b/c) never fire — no character shares another's `createdAt`, and `gapStart` is always 0 — so every fragment fell through to the same fixed fallback anchor, and `InsertAfter(sameLeft, node)` repeatedly *prepended*, rebuilding the run backwards. Un-tombstoning in place (no GC) always preserved order, which is why it only surfaced post-GC.

## Changes

- Thread a **chain anchor** — the fragment placed just before this one in the same restore (spans arrive in document order) — and anchor after it when no same-insertion piece survives, so a purged run is rebuilt left-to-right.
- Thread the op's **`from` position** as the first fragment's fallback anchor. `findRestoreAnchor` previously fell back to `initialHead`, which placed a fully-purged run at the front (`my namehello  is`). This matches the JS SDK's `fallbackAnchor` so server snapshots converge with clients.
- Rung order is now: (a/b/c) same-insertion identity → (d) chain anchor → (e) `from` position → (f) head.

## Test plan

- `TestTextRestoreExecuteAfterGC` used a single `"0123456789"` insertion (one recreate) and never exercised multi-fragment ordering — added `TestTextRestoreAfterGCKeepsOrderAcrossInsertions` covering the char-by-char purge-then-undo case.
- `go test ./pkg/document/... ./api/converter/` green; `golangci-lint` 0 issues.
- Verified end-to-end against the JS SDK companion fix: `history_text` + text/gc integration suites pass against a server built from this branch.

Companion PR: yorkie-team/yorkie-js-sdk (client-side fix).
Fixes the SDK side of wafflebase/wafflebase#629.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restoring text after deleted content has been garbage-collected now preserves the original left-to-right character order.
  * Restoration more reliably places content near its original position, even when surrounding fragments no longer exist.

* **Tests**
  * Added coverage for restoring multiple insertions after garbage collection while preserving document content and structure.

### DIFF
--- a/pkg/document/crdt/rga_tree_split.go
+++ b/pkg/document/crdt/rga_tree_split.go
@@ -733,7 +733,17 @@ func subValue[V RGATreeSplitValue](full V, from, to int) V {
 //     register them before un-registering (keeps GC accounting balanced)
 func (s *RGATreeSplit[V]) restore(
 	spans []restoreSpanValue[V],
+	fromPos *RGATreeSplitNodePos,
+	executedAt *time.Ticket,
 ) (untombstoned, recreated []*RGATreeSplitNode[V], stillTombstoned []GCPair) {
+	// chainAnchor is the node last placed at the current cursor (un-tombstoned
+	// or recreated), in document order. When a recreated fragment has no
+	// surviving same-insertion anchor, chaining after this keeps a purged
+	// multi-fragment run in left-to-right order instead of each fragment
+	// prepending at the same fixed fallback (initialHead), which would rebuild
+	// the run reversed. Spans arrive in document order, so this is always the
+	// recreated fragment's left neighbour.
+	var chainAnchor *RGATreeSplitNode[V]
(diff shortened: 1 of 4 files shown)


hackerwins (approved):
Thanks for your contribution.

Issue #1862, opened 2026-07-10
Add property based CRDT convergence tests to yorkie


ggyuchive:
**What would you like to be added**:

yorkie's integration tests are all hand-written deterministic scenarios. CRDT convergence correctness,
so every replica ending in the same state regardless of the order in which it receives a given set of ops is hard to cover with hand-written cases.

Add **property-based tests** that randomly generate op sequences across multiple clients through the real server sync path and assert every replica's final marshal matches.

Scope of CRDT types and op families:

| CRDT | Ops |
|------|-----|
| Tree | `Edit`, `Style`, `RemoveStyle`, `EditBulk` (with random splitLevel) |
| Text | `Edit`, `Style` |
| Array | `Add*`, `Delete`, `MoveBefore`, `MoveFront`, `MoveLast`, `MoveAfterByIndex` |
| Object | `Set*`, `Delete` |
| Counter | `Increase` |

PBT Target
- [ ] Json.Tree
- [ ] Json.Text
- [ ] Json.Array
- [ ] [Json.Object](https://github.com/yorkie-team/yorkie/issues/1917)
- [ ] [Json.Counter](https://github.com/yorkie-team/yorkie/issues/1894)
- [ ] Running PBT into CI - for CRDT convergence regression test (must be deterministic)

Each PBT drives the real server sync path:
```
Attach N clients → random per-client op sequence → Sync all → assert every client's Marshal() is identical.
```

**Why is this needed**:

Deterministic integration tests only cover scenarios the author imagined.
CRDT bugs usually live in edge cases whose reproduction depends on concurrent op ordering; enumerating them by hand does not scale.

We can use Go PBT built on [pgregory.net/rapid](https://github.com/flyingmutant/rapid).

Related to https://github.com/yorkie-team/yorkie-js-sdk/issues/1288

hiSandog:
The generator should include operation delivery with duplication and reordering, plus replicas that go offline and later receive a compacted history. Persisting the random seed and minimized operation sequence on failure would turn a convergence counterexample into a deterministic regression fixture.

ggyuchive:
It's Roadmap issue, so when you like to take this issue, please open sub-issue of each target and request to assign.

g-hyeong:
@ggyuchive I opened #1894 for the `Json.Counter` PBT target. Could you add it as a sub-issue of this issue and assign it to me?

PR #1918, opened 2026-08-08
Add property-based convergence tests for Json.Object

### LINKED ISSUE #1917: Add property based convergence tests for Json.Object

Part of #1862.

Add a property-based integration test for `Json.Object`.

The generated input is a sequence of Object operations. Each client
applies a seeded random sequence drawn from:

- `SetString` / `SetInteger` / `SetLong` / `SetDouble` / `SetBool` /  `SetNull` on a bounded key range
- `Delete` on the same key range

The key range is kept intentionally narrow so that concurrent writes and
delete/set races on the same key occur frequently.

All clients start from the same empty document attached through the real
server path. After the sequences finish, all clients are synced and the
test verifies that every replica has identical `Marshal()` output.

Each sequence is derived from a fixed seed embedded in the subtest name,
so runs are deterministic (as required for CI in #1862) and a failing
case is reproducible from the test name alone.

The scope is limited to top-level keys with primitive values and a small
number of clients and operations. Nested objects/arrays and other CRDT
types are out of scope.

**Why is this needed**:

The existing Object integration tests
(`test/integration/object_test.go`) validate selected concurrent
set/delete scenarios. They do not systematically explore client
orderings, value types, and delete/set races.

Property-based generation can cover a broader range of operation
sequences through the real server path while checking that all replicas
converge to the same state.

moaventurer:
Convergence coverage for Object relies on hand-written deterministic scenarios, which cannot enumerate the orderings that concurrent operations can produce. Add a seeded random operation generator that applies Object Set/Delete operations from three clients and asserts that every replica converges to the same state, reusing the existing syncClientsThenAssertEqual helper. Each sequence is derived from a fixed seed, so a failing case can be reproduced from the subtest name alone.

Adds a property-based convergence test for `Json.Object`, as part of the plan in #1862.

Three clients each apply a seeded random sequence of Object operations (`SetString`/`SetInteger`/`SetLong`/`SetDouble`/`SetBool`/`SetNull`/`Delete`) over a narrow key range to the same document, then all replicas are synced and asserted to converge to the identical state via the existing `syncClientsThenAssertEqual` helper in `test/integration/main_test.go`. Ten seeded sequences run per test.
Every sequence is reproducible from the seed embedded in the subtest name, so runs are deterministic and a failing case can be minimized into a simplified scenario for a separate report.

Fixes #1917 

The generator uses seeded `math/rand` (with `//nolint:gosec` and a reason) so that sequences stay reproducible.

```release-note
NONE
```

**Additional documentation**:

```docs

```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage for concurrent object-property updates across multiple clients.
  * Verifies replicas converge consistently after mixed value assignments, null updates, and deletions.
  * Runs reproducible randomized scenarios across multiple seeds.

### DIFF
--- a/test/integration/object_property_test.go
+++ b/test/integration/object_property_test.go
@@ -1,0 +1,113 @@
+//go:build integration
+
+/*
+ * Copyright 2026 The Yorkie Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package integration
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/yorkie-team/yorkie/pkg/document"
+	"github.com/yorkie-team/yorkie/pkg/document/json"
+	"github.com/yorkie-team/yorkie/pkg/document/presence"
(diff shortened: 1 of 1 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #1920, opened 2026-08-09
Add fuzz coverage for tree and snapshot decoding

### LINKED ISSUE #1915: Add fuzz coverage for BytesToTree and BytesToSnapshot
**Parent roadmap**: #1858

Follow-up to #1911 / #1912 (which added fuzz coverage for `BytesToObject` and `BytesToArray`). This covers the remaining `api/converter` snapshot decoders from #1858:

- `converter.BytesToTree`
- `converter.BytesToSnapshot`

Both decode client-controlled snapshot bytes, so they're high-value fuzz targets. Each target feeds random bytes to the decoder and asserts it never panics (returning an error is fine).

**Plan** (same approach as #1912):
- Add `FuzzBytesToTree` and `FuzzBytesToSnapshot` to `test/fuzz/fuzz_test.go`.
- Guard the nil-pointer panics fuzzing surfaces (e.g. nil `Tree` body / empty tree nodes for `BytesToTree`; nil `Root` and a non-Object root for `BytesToSnapshot`) so the decoders return errors instead of panicking.
- Add deterministic error-contract tests + a regression corpus for each guarded case.

I'd like to take this — could you please assign it to me?

lemon0333:

Follow-up to #1912. Adds fuzz coverage for `converter.BytesToTree` and `converter.BytesToSnapshot`, the remaining `api/converter` snapshot decoders from #1858.

`proto.Unmarshal` succeeds on many malformed inputs, leaving a message with nil or incomplete fields that the converter then dereferences and panics on. Fuzzing surfaced several such panics, so this PR guards the decode path to return errors instead of panicking:

- `BytesToSnapshot`: a non-object root (previously an unchecked type assertion)
- `fromJSONTree`: nil tree body, and an empty tree with no root node (`NewTree(nil)`)
- `fromTreeNodeID` / `fromTextNodeID`: a tree/text node id whose `createdAt` is nil (later dereferenced when indexing)
- `fromTextNode`: a nil text node id

Includes deterministic error-contract tests and a regression corpus for each guarded case, same as #1912. Only `test/fuzz/` and the source fix are touched (`//go:build fuzz` keeps the targets out of the normal `go test ./...` run).

Fixes #1915

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation when decoding serialized trees and snapshots.
  * Malformed or incomplete data now returns descriptive errors instead of causing panics or creating invalid structures.
  * Added checks for invalid roots, missing nodes or timestamps, and incomplete text node metadata.

* **Tests**
  * Added coverage for malformed inputs and fuzz testing across tree and snapshot decoding scenarios.


hackerwins (approved):
Thanks for your contribution.

Issue #1922, opened 2026-08-11
Add fuzz coverage for VersionVectorFromBytes, ActorIDFromHex, and ParseKeyPath


lemon0333:
**Parent roadmap**: #1858

**What would you like to be added**:

Follow-up to #1912 / #1920 (converter decoders). This covers the remaining decoder targets from #1858's checklist:

- `time.VersionVectorFromBytes` (`pkg/document/time/version_vector.go`)
- `time.ActorIDFromHex` (`pkg/document/time/actor_id.go`)
- `channel.ParseKeyPath` (`pkg/channel/channel.go`)

All three parse client-controllable input, so they're good fuzz targets. Each feeds random bytes/strings to the decoder and asserts it never panics (returning an error is fine).

**Plan** (same approach as #1912 / #1920):
- Add `FuzzVersionVectorFromBytes`, `FuzzActorIDFromHex`, and `FuzzParseKeyPath` to `test/fuzz/fuzz_test.go`.
- Guard any nil/bounds panics fuzzing surfaces so the decoders return errors instead of panicking.
- Add deterministic error-contract tests + a regression corpus for each guarded case.

I'd like to take this — could you please assign it to me?

lemon0333:
This is implemented and merged in #1923 (all three targets — VersionVectorFromBytes, ActorIDFromHex, ParseKeyPath). I referenced it with `Refs` rather than `Fixes`, so it didn't auto-close. This can be closed. 🙏

PR #1923, opened 2026-08-11
Add fuzz coverage for VersionVectorFromBytes, ActorIDFromHex, and ParseKeyPath

lemon0333:

Follow-up to #1912 / #1920. Adds fuzz coverage for the remaining `#1858` decoder targets: `time.VersionVectorFromBytes`, `time.ActorIDFromHex`, and `channel.ParseKeyPath`. Each target feeds arbitrary bytes/strings to the decoder and asserts it never panics.

These three decoders are already guarded (length / hex / key validation checks), so **no production changes are needed** — this only adds the fuzz targets plus a seed corpus for each.

Refs #1922

No production code changes; the decoders already handle malformed input. `//go:build fuzz` keeps these targets out of the normal `go test ./...` run. Verified each target locally with `-fuzz` (1M+ execs, no crash).

```release-note
NONE
```

**Additional documentation**:

```docs

```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

- **Tests**
  - Added fuzz testing for version-vector decoding, actor ID parsing, and key-path parsing.
  - Expanded coverage with valid, empty, malformed, truncated, and edge-case inputs.
  - Added seed cases for invalid hexadecimal values, incorrect lengths, and unusual key-path separators.


hackerwins (approved):
Thanks for your contribution.

PR #1924, opened 2026-08-12
Isolate spans in Tree restore so overlapping undo converges

harrykim8672:
## Summary

Concurrent multi-user Tree undo can diverge and corrupt a document. Two clients that concurrently delete **overlapping** text and both undo diverge once GC has purged the tombstones (so restore takes the recreate path). Each replica rebuilds its run as a monolithic node with misaligned boundaries, and restore's straddle guard `break`s on a piece that crosses a span boundary — dropping the rest of the run. Reported in wafflebase docs multi-user undo.

The existing `history_tree` reconcile cases undo *before* GC runs, so they never exercise this recreate path — which is why it slipped through.

## Change

`Restore` and `Retombstone` now **isolate** the exact `[start, end)` sub-range out of every overlapping piece, splitting at the span boundaries (live or removed), so all replicas converge on the same text-node segmentation — the tree analogue of `RGATreeSplit.isolateRange`. Kept **byte-parallel with the JS SDK** so server snapshots match clients. The live-split metadata overhead and the pending GC pairs from splitting a removed straddler are threaded back through `Restore`/`Retombstone` so the caller's docSize/GC accounting stays exact.

## Test plan

- `go test ./pkg/document/... ./api/converter/` green; `golangci-lint` 0 issues.
- Verified end-to-end against the SDK companion PR: a new `history_tree_concurrent_test.ts` covering every overlap relationship (contained_by / contains / overlap_start / overlap_end / identical / adjacent) under undo and undo+redo after GC passes against a server built from this branch, and **all 135 existing `history_tree` tests still pass**.

## Known limitation

Deleting a **whole element** concurrently with a text edit *inside* it, then both undo after GC, converges on visible content but can still differ in internal text-node segmentation. A child sub-restore is skipped while its parent is transiently purged, so the replicas keep different split points and the element's monolithic span cannot re-introduce them. Tracked as two skipped tests on the SDK side. A sound fix needs the child restore's split points to survive a transiently-purged parent (e.g. undo-stack-aware GC). Merge-normalizing segmentation was tried and rejected as non-commutative (it broke GC/tombstone symmetry after redo).

Companion PR: yorkie-team/yorkie-js-sdk (client-side fix + tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved restoration of text and document content, including partially overlapping ranges.
  * Correctly revives deleted content and recreates content that was previously removed.
  * Improved handling of cleanup and data-size changes during restore and retombstone operations.
  * Errors encountered during these operations are now handled and reported more reliably.

### DIFF
--- a/pkg/document/crdt/tree.go
+++ b/pkg/document/crdt/tree.go
@@ -887,6 +887,16 @@ func (t *Tree) Purge(child GCChild) error {
 func (t *Tree) Restore(spans []*TreeRestoreSpan) (
 	untombstoned, recreated []*TreeNode, pairs []GCPair, diff resource.DataSize, err error,
 ) {
+	// Splitting a removed straddler buffers born-removed remainders as pending
+	// GC pairs (see TreeNode.Split). Drain them into the returned pairs on EVERY
+	// path — including a mid-restore error — so they can never linger to
+	// contaminate the next operation's drain. The caller registers these BEFORE
+	// un-registering the un-tombstoned targets, so a split-born target is walked
+	// gc->live correctly (mirrors the Text path).
+	defer func() {
+		pairs = append(pairs, t.drainPendingGCPairs()...)
+	}()
+
 	for _, span := range spans {
 		if !span.IsText {
 			node := t.findFloorNode(span.ID)
@@ -899,7 +909,7 @@ func (t *Tree) Restore(spans []*TreeRestoreSpan) (
 			}
 			created, cErr := t.recreateFromSpan(span, span.ID.Offset, span.Length)
 			if cErr != nil {
-				return nil, nil, nil, resource.DataSize{}, cErr
+				return nil, nil, nil, diff, cErr
 			}
 			if created != nil {
 				recreated = append(recreated, created)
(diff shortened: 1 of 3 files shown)


harrykim8672:
The failing `build` check is an unrelated flaky test: `TestRequestTimeout/request succeed after timeout` in `pkg/webhook/client_test.go`, which races on a ~5ms timing margin (10ms delay vs 15ms `RequestTimeout`) under CI load. This PR touches only `pkg/document/crdt` and `pkg/document/operations` — no webhook code — and `go test ./pkg/document/... ./api/converter/` + `golangci-lint` are green locally. Re-triggering CI.

harrykim8672:
Re-triggered; the previous re-run flaked on the same unrelated test again (`TestRequestTimeout/request succeed after timeout`, `pkg/webhook/client_test.go:215`). Note this test both fails and **panics** on a timeout race because it uses `assert.NotNil(resp)` (non-halting) and then dereferences `resp.Greeting` — so a slow runner trips the ~5ms margin and nil-panics. It's green on `main` and this PR touches no webhook code. If it flakes again, a maintainer re-run would help (I can't re-run from a fork).

hackerwins (approved):
Thanks for your contribution.

PR #1928, opened 2026-08-14
Stamp merge-bound inserts and filter style-range interlopers

### LINKED ISSUE #1916: Concurrent `Tree.Style` range ending after a merge-moved child styles a concurrently inserted node on only one replica
**What happened**:

This is the follow-up variant noted in #1909: a style range end anchored after a moved child still diverges.

- client A: `Tree.Edit(8, 8, <p>, 0)`, then `Tree.Style(0, 6, {"bold": "x"})`. The range end is anchored after `c`, a non-left-most position inside the second paragraph.
- client B: `Tree.Edit(0, 5, nil, 0)`. Concurrently removes the first paragraph and the second paragraph's opening token.

After a full sync, the structure converges but the attributes do not:

- client A: `{"type":"p","children":[]}`
- client B: `{"type":"p","children":[],"attributes":{"bold":"x"}}`

Both replicas converge to the same tree, including attributes.

```go
func TestTreeNonLeftmostStyleDivergence(t *testing.T) {

	t.Run("style range ending after a moved child", func(t *testing.T) {
		ctx := context.Background()

				Type: "r",
				Children: []json.TreeNode{
				},
			})
			return nil
		}))

			root.GetTree("t").Edit(8, 8, &json.TreeNode{Type: "p", Children: []json.TreeNode{}}, 0)
			return nil
		}))
			root.GetTree("t").Style(0, 6, map[string]string{"bold": "x"})
			return nil
		}))
			return nil
		}))

	})
}
```

- The range end resolves through the moved child (`c`) directly, so it never reaches the §9.3 boundary redirect from #1909. On client B the concurrently inserted `<p>` sits between the merge-source tombstone and `c`, inside the resolved range; on client A it was outside.
- A boundary adjustment cannot cover this case: the range must include `c` but exclude the inserted `<p>`, which sits before `c` on the merging replica. The traversal likely needs per-node filtering instead.
- The `RemoveStyle` variant passes the Marshal oracle here because `Marshal` omits empty attribute sets, but the same one-sided state is materialized (`RemoveAttr` creates an empty RHT on the applying replica only). The JS SDK exposes it via `toSortedJSON` and diverges (yorkie-team/yorkie-js-sdk#1311).
- The JS SDK diverges identically for the style variant (yorkie-team/yorkie-js-sdk#1311).

**Environment**:
- Browser and version:  N/A
- Yorkie version (use `yorkie version`):  main 
- Yorkie JS SDK version: main

Nahee-Park:
A concurrent `Tree.Style`/`Tree.RemoveStyle` whose range end is anchored after a merge-moved child styles a node concurrently inserted at the merged anchor on the applying replica only. 
From `<r><p>ab</p><p>cd</p></r>`: 
- A inserts an empty `<p>` at index 8 and styles `[0,6)`
- B concurrently removes `[0,5)` 

After a full sync, the `<p>` ends up `bold="x"` on B only.

The anchor resolves through the moved child, so the §9.3 redirect from #1909 never fires, and the `canStyle` version-vector check passes because the interloper is the styling client's own insert.

This PR fixes the issue in two steps:
1. It stamps inserts declared inside a merged-away parent with `MergedFrom` = declared parent.
2. During style traversal, it skips nodes under the merge target that appear after the merge-source tombstone and have no merge stamp. Each node is judged by its highest ancestor under the target, so an interloper's descendants are skipped with it.

Nodes carrying any `MergedFrom` fail open because an earlier merge leaves the original source in the stamp (first-move rule), so stamp equality cannot prove the node was outside the range.
`SplitElement` copies `MergedFrom`/`MergedAt` to the split product (`SplitText` parity) so the judgment is application-order independent.
The skip checks are shared between `Style` and `RemoveStyle` via a single predicate with a precomputed after-tombstone set.

Tests include six new complex cases:
- the reported case, which fails without this fix
- a RemoveStyle variant that also verifies no attribute entry is created on the interloper, which Marshal alone cannot detect
- interloper descendants
- an own insert inside the style range
- a sibling before the tombstone
- a child brought in by an earlier synced merge that should remain styled

Unit tests also verify the merge stamp and that SplitElement copies the stamp correctly.

All tree test lanes and golangci-lint pass.

Fixes #1916

- Found by the property-based Tree test for yorkie-team/yorkie-js-sdk#1298 right after #1909 landed; the JS SDK diverges identically (yorkie-team/yorkie-js-sdk#1311) and a mirror PR follows.
- Design: `docs/design/concurrent-merge-split.md` §9.4 (Fix 22), added here, including a known-limitations list (relay-only merged parents, the `MergedAt` fallback, stamped inserts in the delete cascade and the §1.1 boundary scan).
- Two variants remain and reproduce on main without this change (a range *start* anchored after a moved child; an edit-only divergence); both are next-cycle issues.

```release-note
Fix Tree.Style and Tree.RemoveStyle styling a node concurrently inserted at a merge-removed anchor on only one replica.
```

**Additional documentation**:

```docs
- [Design]: docs/design/concurrent-merge-split.md §9.4 (Fix 22)
```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved concurrent tree editing when content is merged, split, moved, or inserted into previously merged locations.
  - Prevented incorrect styling and style removal around moved anchors, nested descendants, and inserted content.
  - Preserved merge information across split operations for more consistent results across replicas.

- **Documentation**
  - Added guidance on merge provenance, moved-anchor filtering, known limitations, and follow-up work.

- **Tests**
  - Added regression and concurrency coverage validating document structure, styling, and replica consistency.

### DIFF
--- a/docs/design/concurrent-merge-split.md
+++ b/docs/design/concurrent-merge-split.md
@@ -440,8 +440,79 @@ both directions. Ranges that genuinely cover the merged content
 (anchored outside the merged parent) are unaffected.
 
 A range end anchored after a moved child (a non-left-most position in
-the merged parent) does not go through the §1.1 redirect and can still
-diverge; this remains a follow-up.
+the merged parent) does not go through the §1.1 redirect; §9.4 covers
+it with a per-node filter.
+
+### §9.4 Per-Node Filter at Moved Anchors
+
+A range end anchored after a moved child resolves through the moved
+child directly into the merge target. Nodes that sit between the
+merge-source tombstone and the moved children — concurrent inserts
+anchored at the tombstone — then fall inside the traversal, although
+the styling client saw them outside its range, after the then-live
+parent. The version-vector check in `canStyle` does not help: the
+interloper may be the styling client's own insert, causally known but
+positionally excluded.
+
+Two mechanisms close this:
+
+1. **Intended-parent stamp.** An insert whose position was declared
+   inside a merged-away parent physically lands in the merge target.
+   `intendedMergeParent` stamps such content with
+   `MergedFrom = declared parent` (merge ticket taken from a moved
+   sibling), making it indistinguishable from merge-moved children —
+   which is what it logically is. This also aligns merge-delete
+   propagation: when the declared parent is later fully deleted, the
+   insert dies with it, matching the replica that applied the insert
+   into the live parent and cascaded the delete.
+2. **Interloper filter.** `mergedAnchorInterloperGuard` activates when
+   the range-end position's declared parent is a merged-away tombstone
+   sitting directly under the merge target and the merge is unknown to
+   the styling client's version vector. During the style traversal it
+   skips element nodes under the target, positioned after the
+   tombstone, that carry **no merge stamp at all** — only such nodes
+   (concurrent inserts anchored at the tombstone) are positively
+   identifiable as interlopers. Any node with a `MergedFrom` fails
+   open: a child that reached the declared parent via an earlier merge
+   keeps the *original* source in `MergedFrom` (first-move stamp rule,
+   Fix 20), and path compression on `mergedInto` makes "was it inside
+   the declared parent" unrecoverable afterwards, so stamped nodes
+   fall back to pre-§9.4 behavior rather than risk skipping a node the
+   styler legitimately covered. Each traversed node is judged by its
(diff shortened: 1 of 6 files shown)


hackerwins (approved):
Thanks for your contribution.

PR #1933, opened 2026-08-16
Archive seven task records, index docs/ and scripts/, gate doc links

hackerwins:
Four commits: archive the finished task records, give `docs/` and
`scripts/` an index, add the design doc that index was missing, and gate
the whole graph so none of it rots.

## Archive the finished task records

Moves seven pairs from `docs/tasks/active/` to `docs/tasks/archive/2026/08/`
and regenerates the indexes with `scripts/tasks-archive.sh` +
`scripts/tasks-index.sh`. Each one's change is already on `main`:

| Task | Merged as |
|---|---|
| `20260802-tree-insert-into-merged-parent` | #1905 (JS: yorkie-js-sdk#1303) |
| `20260803-tree-chained-merge-convergence` | #1906 (JS: yorkie-js-sdk#1305) |
| `20260803-tree-style-merge-anchor` | #1909 (JS: yorkie-js-sdk#1309) |
| `20260808-tree-style-moved-anchor` | #1928 (JS: yorkie-js-sdk#1317) |
| `20260813-rpc-start-bind-sync` | landed with #1925 |
| `20260814-duplicate-tree-node-id` | #1927, #1929 |
| `20260815-undo-redo-go-port` | #1932 |

Three plans still carried open boxes for work that had in fact landed:
"Open PR" and "Port to yorkie-js-sdk" on the two tree tasks, and the
whole per-step list on the undo/redo port. Those boxes are now closed
with the merged PR named beside them.

The undo/redo port also gets the `## Review` section it left blank. It
records what shipped, and that the four defects the port surfaced were
split out and are still open as the 2026-08-16 tasks.

`docs/tasks/active/` keeps only those four.

## Index docs/ and scripts/

`docs/README.md` and `scripts/README.md` are new. The root README linked
`docs/design/` and `docs/tasks/` individually — so a third subdirectory
would have had to be added in two places to be discoverable — and
`scripts/` was not reachable from anywhere, meaning a reader had to run
`ls` to learn `tasks-archive.sh` exists. Each directory now owns its
table and the root README links the directory instead of restating its
contents.

`docs/design/disable-presence.md` had been on disk with no row in the
design index. Found by running wafflebase's `verify-doc-index.mjs`
coverage check against this repository; it was the only gap.

## Gate the graph

An index is a promise to a reader, and an unverified one rots into a
list of dead ends — so the indexes above arrive with the check that
keeps them honest. `scripts/verify-doc-links.mjs` walks from `CLAUDE.md`,
`AGENTS.md`, and `README.md` and fails on a target that does not exist.

Four categories are deliberately not findings, each measured against
real false positives before the rule was added:

| Skipped | Why |
|---|---|
| Archived task records (reached, not walked) | A finished task's citations record what was true then; gating them means editing history to keep a checker quiet |
| Extensionless targets → try `.md`, `index.md` | A docs site links a sibling as `./formulas`; demanding the extension rewrites working navigation |
| Site-absolute `/foo` targets | Resolve against a site root the script cannot locate |
| Inline code and fenced blocks | A document explaining this check writes `[board](board/TYPO.md)`, and nobody can click it |

It runs in its own `Docs` workflow rather than as a step in `ci.yml`,
because that workflow sets `paths-ignore: "**/*.md"` — a step added
there would be skipped on precisely the pull requests this polices.

The last commit closes the one gap it found in this repository's own
indexes: the generated tasks index named `active/` and `archive/` in
code spans instead of linking them, leaving both READMEs unreachable —
`active/README.md` worst of all, since it is hand-written and nothing
else points at it.

The same script lands in yorkie-js-sdk#1321 and wafflebase#861.

## Test plan

- `./scripts/tasks-archive.sh` moved exactly the seven pairs and left
  the four open tasks alone; `./scripts/tasks-index.sh` regenerated both
  indexes, archived count 20 → 27.
- `node scripts/verify-doc-links.mjs` — green, no findings.
- wafflebase's `collectFindings()` (coverage, the complementary
  direction) run against this tree — no design-doc or scripts-index gaps.
- Every `.md` in the repository is now reachable from a root document;
  it was 44 short before this branch.
- Docs and one new script — no Go code, no build or test impact.

## Summary by CodeRabbit

* **Documentation**
  * Improved project, design, task, archive, and script documentation.
  * Added detailed records of completed fixes, lessons learned, and follow-up items.
  * Updated task indexes and links for easier navigation.

* **New Features**
  * Added automated documentation-link verification with support for Markdown files, directories, images, and imports.
  * Added a documentation quality workflow for pull requests and pushes.

* **Tests**
  * Added comprehensive coverage for link validation, traversal, filtering, missing targets, and edge cases.


hackerwins:
Thanks — three of the six are applied in e70368b7, and I'm declining the other three. Reasoning for each.

## Applied

**Path filter on `.github/workflows/docs.yml`.** You're right, and the hole is worse than the review states: a link breaks when its *target* moves, and targets are not only markdown. `packages/cli`-style READMEs cite source files, so deleting a cited `.ts` breaks a link without touching a single `.md` — and the gate would have sat out that PR entirely. Both triggers now run unfiltered. The job installs Node and reads text, so the filter was buying seconds against a hole in the one check that watches links.

**Workflow permissions.** Dropped to `contents: read` with `persist-credentials: false`. No step here talks to the API.

**Tests for `collectFindings` / `linkTargets`.** Fair hit, and on my own stated design — the header says the functions are exported "so the suite can point it at a planted tree," and there was no suite. `scripts/test/verify-doc-links.test.mjs` now has 23 cases covering the traversal and every exclusion the header argues for: frozen archive records, fenced and inline code, extensionless and site-absolute targets, directory indexes, `@path` imports, cycles, and the no-root case.

**`scripts/README.md` on `tasks-archive.sh`.** Correct — a todo without a parseable `**Created**` date is skipped too, so eligibility is not unchecked boxes alone. Reworded.

**Trailing space in the `grep -c "it("` span.** Fixed.

## Declined

**`(nil, nil)` vs `ErrOperationSkipped` (`20260815-undo-redo-go-port-todo.md:743-778`)** and **the `27 PASS, 2 SKIP` count (`1365-1381`).**

Both are real divergences between what the plan predicted and what shipped, and both are already recorded — in the design doc, which is where that record belongs. `docs/design/undo-redo-go-port.md:661` states it directly:

> A skip is a sentinel error, not `(nil, nil)` — `(nil, nil)` cannot be told apart from "executed, no reverse". JS distinguishes them structurally (`undefined` result vs. `{reverseOp: undefined}`); Go needs `ErrOperationSkipped` to say the same thing

with the binding contract at `:144`, and the whole §"What Shipped vs. What This Plan Predicted (Task 21)" devoted to exactly this class of drift. The `## Review` section this PR adds to the archived plan points at that section by name.

Editing the plan's step text to match the outcome would destroy the one thing the artifact is good for. A completed plan is a record of what was believed *before* the work; the value of "the plan said `(nil, nil)`, the port shipped a sentinel, here is why" is entirely in the gap being visible. Overwrite the prediction and the design doc's divergence section becomes a list of differences from a document that no longer differs.

This is the same principle the gate in this PR encodes: archived task records are reached but never walked, precisely so a checker cannot pressure anyone into rewriting finished history.

**`can legitimately claim one id` (`20260814-duplicate-tree-node-id-todo.md:75-82`).**

I don't think the document contradicts itself. Line 75-82 says that *because split delimiters are simulated rather than replayed*, two nodes in one change can share an id today — and that the first version of the fix broke `TestTree/edit its content with path when multi tree nodes passed` by treating every collision as corruption. Lines 120-122, in the same file's follow-up list, say that this "deserves its own fix: normal editing should not be able to produce two nodes under one identity."

So: it happens, the fix has to tolerate it, and it shouldn't happen. That is one coherent position, not two conflicting ones. "Legitimately" is carrying "this is a real non-undo source, not corruption" rather than "this is correct," which I'd grant is a weaker word than it could be — but the follow-up it names shipped as #1929, *Stop ordinary editing from putting two nodes under one ID*. The concern is closed in the code, and rewording an archived record after the fact would only blur which of the two states the author was describing.

**`keeps every document loadable` (`20260814-duplicate-tree-node-id-lessons.md:51-52`).**

I read this as a claim about the mechanism, not an extrapolation from the three replays. The sentence contrasts two strategies: reject the malformed change and the document is unopenable at a later change; drop the duplicated content and apply the rest, and the replay cannot fail — so the document loads. That holds by construction, and the clause that follows ("at the cost of an undo that no longer restores its text") is the price of the same property. The three datasets in the next section are evidence for *choosing* the strategy, not the basis of the loadability claim.

PR #1934, opened 2026-08-17
Count a removed container's descendants in the GC total

hackerwins:
## Summary

`Root` kept three size-accounting sites that were individually plausible and jointly inconsistent. `RegisterElement` books a container **and every descendant** into `Live`; `deregisterElement` subtracted **both** from `GC`; but `RegisterRemovedElementPair` moved **only the container itself**. Deleting a non-empty object therefore stranded every descendant's cost in `Live` and pushed `GC` below zero once the tombstone was collected:

```
before GC: {Live:{Data:2 Meta:72} GC:{Data:0 Meta:72}}
after GC:  {Live:{Data:2 Meta:72} GC:{Data:-2 Meta:-48}}
```

`Document.Update` gates `MaxSizeLimit` on `DocSize.Total()`, which is `Live.Total() + GC.Total()`, so a negative `GC` silently discounts a document below its real cost and lets through updates that should be rejected.

Four routes reach the same corruption. Three of them turned up only by running scenarios, not by reading the diff:

| # | route | effect |
|---|---|---|
| 1 | remove a non-empty container | descendants stranded in `Live`, `GC` negative |
| 2 | two replicas remove the same container concurrently | **replicas report different sizes** for a document whose content converged |
| 3 | a member removed remotely inside an already-removed container | its size moves into `GC` twice |
| 4 | a remote `Set` lands inside a container this replica already tombstoned | booked into `Live`, subtracted from `GC` |

(2) is the worst: `DocSize` gates `MaxSizeLimit`, so one replica can accept an update another rejects, permanently.

## Approach

Guarding each route needed a third guard, which was the signal that the rule the three sites collectively maintain had never been written down. So it is written down, and each site holds to it:

> Every registered element's `DataSize` is counted in exactly one of `Live` or `GC`, and `Root.sizeInGC` records which — and, for GC, the exact amount charged.

- `moveSizeToGC` is the only way a size enters `GC`, and it is idempotent. That covers (2) and (3).
- `RegisterRemovedElementPair` applies it to the element and, for a `Container`, to every descendant.
- `deregisterElement` subtracts from `GC` only what `sizeInGC` says was charged, and from `Live` anything not listed there. That covers (4), which no descendant walk can reach — the element does not exist when the walk runs.

`sizeInGC` stores the charged `DataSize` rather than a flag because **`DataSize()` is not stable over an element's lifetime**: it grows by one ticket the moment `removedAt` is set, which can happen *after* the size has moved. Recording the amount keeps the two sides symmetric regardless of when either reads `DataSize()`.

The alternative from the original filing — shrinking `deregisterElement` to the container's own size — was rejected: `Live` is populated by the descendant-walking `RegisterElement`, so descendants have to leave `Live` somewhere.

Only `Object` and `Array` implement `Container`. `Text` and `Tree` are leaf `Element`s whose `DataSize()` already covers their whole structure, and were never affected — verified empirically alongside the object/array cases.

## Note for reviewers: this reverses a call made on #1932

The second commit changes `Set.Execute` to deregister the element actually registered under that `createdAt` rather than the incoming deep copy. CodeRabbit raised exactly this on #1932 and I replied that it was a line-for-line faithful port of JS and would be filed as a cross-SDK divergence rather than fixed one-sidedly.

That is being reversed deliberately. The reasoning that made "file, don't fix" right for the tree/undo divergences does not hold here: undo/redo stacks and reverse operations are behaviour a peer observes, whereas `DocSize` is local bookkeeping that never reaches the wire, a snapshot, or a change log. Fixing it in Go alone cannot desynchronise replicas — it can only stop `MaxSizeLimit` from being enforced against a corrupted number. The design doc's own precedent (`undo-redo-go-port.md`) already draws the same line, treating a `DocSize` divergence as Critical-and-fix rather than file-and-defer, because `DocSize` gates `MaxSizeLimit`.

The two differ in practice whenever the tombstone moved on after the reverse captured its copy, which a peer writing into the removed container is enough to cause. `@coderabbitai` has a stored learning from #1932 saying not to flag this line — please re-evaluate it rather than relying on that.

## Not fixed here

`NewRoot` seeds `Live` from an already-tombstoned element's post-removal size and then refunds a ticket that `Live` did hold, over-crediting `Live.Meta` by one `TicketSize` per uncollected tombstone. It is byte-identical before and after this branch (`+24/+48/+72/+96` for one through four sibling tombstones, on both). Filed as `docs/tasks/active/20260817-docsize-snapshot-rebuild-drift-todo.md` — it matters because `MaxSizeLimit` is checked against `cloneRoot.DocSize()`, which is a rebuild, while `Document.DocSize()` reports the incremental figure.

The JS SDK reproduces the container bug with byte-identical figures (`registerRemovedElement`, `crdt/root.ts:257`, has the same missing walk; `set_operation.ts:98-104` the same wrong deregister target). Filed as `docs/tasks/active/20260817-docsize-js-sdk-parity-todo.md`.

## Test plan

Six new subtests in `TestDocumentSize`, each asserting the document's size returns exactly to the empty-document baseline after remove-then-GC, and — where two replicas are involved — that they agree:

- `removing a non-empty container test` (object, array, nested object)
- `removing a container holding an earlier tombstone test`
- `concurrently removing the same container test`
- `removing a member inside an already removed container test`
- `restoring a container over a diverged tombstone test`
- `undoing the removal of an array container test`

All six fail on `main` and pass here; the seven pre-existing `TestDocumentSize` subtests are unaffected.

- `go test ./...` — pass
- `make lint` — 0 issues
- `make test` (integration, MongoDB) — pass
- `make test-complex` — pass except `TestClientWithShardedDB/FindDocInfoByRefKey_with_duplicate_ID_test`, which fails identically on `main` (needs a sharded MongoDB; the local container is single-node)

## Also here: the pkg/webhook test flakes that blocked this PR's CI

Unrelated to the accounting change, carried here rather than in its own PR because it failed this branch's CI twice and nothing else on the branch reaches `pkg/webhook`. It is a separate commit.

Stressing the package turned up **four** instances of one shape: a budget tight enough that scheduler latency, not the behaviour under test, decides the outcome.

`newDelayServer` was the root of the worst. It waited on a context derived from `r.Context()`, so the wait ended not only after `delayTime` but also the moment the client gave up — and then wrote 200. The subtest asserting a timeout was racing a response it was never supposed to receive, and lost roughly one run in two hundred. It now waits for the delay to actually elapse and drops a cancelled request without answering.

The rest are budgets:

| test | why it could fail | fix |
|---|---|---|
| `TestRequestTimeout` success case | 10ms server delay vs 15ms budget — 5ms to cover connect, response, scheduling | budget of seconds (free: the server answers at 10ms) |
| `TestRequestTimeout` timeout case | shared the 10ms server, plus the bug above | its own 200ms server vs a 5ms budget |
| `TestBackoff` ×2 | about retry exhaustion (`ErrWebhookTimeout` means the retries ran out, not that a request timed out), but a 10ms per-request budget could turn that into a bare `context deadline exceeded` | budget of seconds (free) |
| `TestRequestFailure` ×2 | 10ms context / 15ms budget while asserting a 503 and a JSON-decode error against servers that answer immediately | budget of seconds (free) |

Widening a budget is free wherever the server answers first, which is everywhere except the one case that needs the server to be slow — there `httptest`'s `Close` waits for the outstanding handler, so the delay is the price of every run and 200ms buys a 40x margin without it. The package still runs in under a second.

The success case also took down the whole package binary when it failed: `assert.NoError` does not halt, so execution reached `resp.Greeting` with `resp` nil and panicked. The two checks the following lines dereference are now `require`.

One thing flagged but not changed: `request fails with context done test` asserts a 503, which is retry exhaustion — if its context won, `retry` would return `(0, ctx.Err())` and the assertion would fail. So despite the name it verifies the path where the context does *not* fire. Widening the deadline makes that deterministic; making the context genuinely decide the outcome would mean asserting something different, which does not belong in this PR.

Measured with `-count=100 -cpu=2` over twelve rounds: 1200 iterations, no failures. `main` fails within the first couple of hundred.

https://claude.ai/code/session_01BBBFtLJjQ8WZCXj1d22stF

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected document-size accounting for nested containers, removals, garbage collection, restoration, and undo/redo operations.
  * Prevented duplicate or incorrect size charges during concurrent deletions and tombstone reuse.
  * Improved consistency of live and garbage-collected size tracking across replicas.

* **Tests**
  * Added regression coverage for nested-container removal, restoration, garbage collection, concurrent deletion, and undo scenarios.

* **Documentation**
  * Added task tracking and lessons for document-size accounting, snapshot rebuild discrepancies, and JavaScript SDK parity.


hackerwins:
JS SDK mirror: yorkie-team/yorkie-js-sdk#1322 — same invariant, same six regression tests, same expected constants.

PR #1940, opened 2026-08-18
Add property based convergence tests for Json.Text

### LINKED ISSUE #1931: Add property based convergence tests for Json.Text
**Parent roadmap**: #1862

Add property-based convergence tests for `Json.Text`, following the pattern established for `Json.Object` in #1918 (`test/integration/object_property_test.go`) — and mirroring the sibling roadmap targets (`Json.Counter` #1894).

**Plan**:
- Add `test/integration/text_property_test.go` (`//go:build integration`) with:
  - `applyRandomTextOps(...)` — a deterministic, seed-driven sequence of `Text` operations (`Edit` insert/delete over a narrow index range, plus `Style`), so a failing case is reproducible from its seed.
  - `TestTextPropertyConvergence` — N replicas (e.g. 3) each apply a random op sequence, then synchronize, and the test asserts every replica converges to the same `Text` state.
- Keep the ranges intentionally narrow (like the Object test's small key space) so replicas touch the same region often, which is where text convergence is most at risk.

I'd like to take this — could you please assign it to me? (I recently completed the fuzz-coverage roadmap #1858 for the converter/time/channel decoders, so this is a natural follow-on.)

lemon0333:

Adds a property-based convergence test for `Json.Text`, following the pattern established for `Json.Object` in #1918.

Each of N replicas applies a deterministic, seed-driven sequence of `Edit` (insert/replace/delete over a narrow range) and `Style` operations, then the replicas synchronize and the test asserts they all converge to the same state. The shared `Text` is created on one replica and propagated first, so every replica edits the same CRDT. Ranges are kept narrow so replicas touch the same region often, which is where text convergence is most at risk; a failing case is reproducible from its seed.

Fixes #1931

Verified locally with `go test -tags integration -race -run TestTextPropertyConvergence ./test/integration/` (all 10 seeds converge, no data races).

```release-note
NONE
```

**Additional documentation**:

```docs

```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage for reproducible random text edits and styling operations across multiple replicas.
  * Verifies that synchronized replicas consistently converge to the same final text state.

### DIFF
--- a/test/integration/text_property_test.go
+++ b/test/integration/text_property_test.go
@@ -1,0 +1,137 @@
+//go:build integration
+
+/*
+ * Copyright 2026 The Yorkie Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package integration
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/yorkie-team/yorkie/pkg/document"
+	"github.com/yorkie-team/yorkie/pkg/document/json"
+	"github.com/yorkie-team/yorkie/pkg/document/presence"
+	"github.com/yorkie-team/yorkie/test/helper"
+)
+
+const (
+	// textPropertyClientCount is the number of replicas that edit the
+	// document concurrently.
+	textPropertyClientCount = 3
+
+	// textPropertySeedCount is the number of independent random sequences
+	// to run.
+	textPropertySeedCount = 10
+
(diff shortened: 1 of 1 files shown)


lemon0333:
@coderabbitai Thanks for the review. Dispositions:

**1. Span width (line 81) — Major — Fixed** ✅ (commit 168b91a)
Both the edit and style branches could select all remaining local text, which contradicts the narrow-overlapping-range goal. Added a `textPropertyMaxSpan = 3` constant and capped each span in both branches via `min(textPropertyMaxSpan, length-from)`, mirroring the narrow `objectPropertyKeyRange` convention from the merged Object test.

**2. Package comment (line 19) — Minor — Skipped.**
The `integration` package already has its package comment in `subscription_leak_test.go` ("// Package integration contains end-to-end tests..."). Go allows only one package comment per package, so adding a second here would be a duplicate. The merged sibling `object_property_test.go` also (correctly) has none.

**3. Detach after each seed subtest (line 115) — Minor — Skipped.**
This matches the established convention in the merged `object_property_test.go`, which attaches per seed and relies on `deactivateAndCloseClients` (Deactivate detaches all docs) for teardown. Each seed subtest uses a distinct doc key (`helper.TestKey(t)` derives from the subtest name), so there's no cross-seed stale-attachment interference. Kept the three property tests consistent; if per-subtest detach is desired it'd be better applied uniformly across all of them in a follow-up.

Verified: `go test -tags integration -race -run TestTextPropertyConvergence ./test/integration/ -count=1` passes (10 seeds); `golangci-lint run --build-tags integration` reports no issues in `text_property_test.go`.

hackerwins (approved):
Thanks for your contribution.

PR #1954, opened 2026-08-26
Recover style ranges collapsed by a merge at the from anchor

### LINKED ISSUE #1942: Concurrent Tree.RemoveStyle range starting after a merge-moved child materializes a one-sided attribute entry that Marshal hides
**What happened**:

This is the from-side variant of #1916. When a `RemoveStyle` range starts after a merge-moved child, the two replicas resolve the range differently. `Marshal` still converges because it omits empty attribute containers, but the internal attribute state diverges.

- client A: `Tree.Edit(8, 8, <p>, 0)`, then `Tree.RemoveStyle(6, 9, ["bold"])`. The range starts after `c` and ends inside A's new `<p>`, so A's local application touches the new `<p>`.
- client B: `Tree.Edit(0, 5, nil, 0)`. Concurrently removes the first paragraph and the second paragraph's opening token.

After a full sync both replicas marshal to `<r><p></p>cd</r>`, but client A's surviving `<p>` holds one attribute RHT entry (the removal tombstone that `RemoveAttr` records for a missing key) while client B's holds none. The asymmetry is real state that round-trips through snapshots; the marshal-level oracle cannot see it.

Both replicas hold the same CRDT state, including attribute RHT entries.

The following test fails on every run (the marshal assertion passes; the internal-state assertion fails):

```go
func TestTreeRemoveStyleFromSideAsymmetry(t *testing.T) {

	ctx := context.Background()

			Type: "r",
			Children: []json.TreeNode{
			},
		})
		return nil
	}))

		root.GetTree("t").Edit(8, 8, &json.TreeNode{Type: "p"}, 0)
		return nil
	}))
		root.GetTree("t").RemoveStyle(6, 9, []string{"bold"})
		return nil
	}))
		return nil
	}))

	// The surviving <p> must carry no attribute entries on either replica.
	for i, pair := range []clientAndDocPair{{c1, d1}, {c2, d2}} {
		tree, ok := pair.doc.RootObject().Members()["t"].(*crdt.Tree)
		assert.True(t, ok)
		var survivor *crdt.TreeNode
		for _, child := range tree.Root().Children(true) {
			if child.Type() == "p" && !child.IsRemoved() {
				survivor = child
			}
		}
		assert.NotNil(t, survivor, "d%d: live <p> not found", i+1)
		if survivor.Attrs != nil {
			assert.Empty(t, survivor.Attrs.Nodes(),
				"d%d: attribute entry materialized", i+1)
		}
	}
}
```

- Same range anchoring problem as the JS report (yorkie-team/yorkie-js-sdk#1324), where the divergence is visible in `toSortedJSON` as a one-sided `attributes: {}`.
- The per-node filter from #1916's fix cannot apply because the applier's traversal never reaches the node.
- Found by the property-based Tree test for yorkie-team/yorkie-js-sdk#1298.

**Environment**:

- Operating system: macOS 15.3
- Browser and version: N/A
- Yorkie version (use `yorkie version`): 0.7.17
- Yorkie JS SDK version: main

Nahee-Park:

A `Tree.Style`/`Tree.RemoveStyle` whose range starts after a merge-moved child styles nothing on the applying replica. From `<r><p>ab</p><p>cd</p></r>`:

- client A inserts an empty `<p>` at index 8, then calls `RemoveStyle(6, 9)` (from after `c`, to inside the insert)
- client B concurrently removes `[0,5)`, merging the paragraphs

`Marshal` hides the asymmetry (it omits empty attribute containers), but only A's insert carries the removal entry in its attribute RHT, and the state round-trips through snapshots. The JS SDK shows it as a one-sided `attributes: {}`.

**Root cause:** the merge moves the anchor child behind the insert, so the resolved range inverts (start past end) and the traversal is empty. The per-node filter from #1916 cannot help: it only skips visited nodes, and nothing is visited.

**Fix:** `reversedFromAnchorRecovery`, the from-side counterpart of that filter. When the range start matches the same merged-anchor shape and the resolved range actually collapsed, it re-anchors the traversal start just after the last live sibling before the merge-source tombstone; the shared skip predicate then styles only nodes positively identified as interlopers, and version-vector checks keep unknown inserts unstyled. An ordered range that moved with the merge is left untouched.

The removal tombstone now materializes on both replicas: even for a missing key it arbitrates a concurrent `SetAttr` with an earlier ticket by LWW. The regression test pins the converged entry/entry state directly, since Marshal alone cannot detect it.

Tests: three complex cases (Style reversal with the converged XML pinned, RemoveStyle with the symmetric-entry pin, ordered-range control) and two unit tests at the CRDT layer. `stylePrevAttrs` and `styleClientLamportAt` are extracted from the duplicated Style/RemoveStyle blocks (gocyclo).

Fixes #1942

- The from-side variant documented as a §9.4 known limitation when #1928 landed; the JS SDK diverges identically (yorkie-team/yorkie-js-sdk#1324, mirror PR follows). Found by the Tree PBT for yorkie-team/yorkie-js-sdk#1298.
- Design: `docs/design/concurrent-merge-split.md` §9.4 "From-side recovery (Fix 23)" plus a rewritten known-limitations list; the residual shapes listed there all reproduce on main without this change.
- Verified with the unit, integration `TestTree*`, and complex `TestTreeConcurrency*` lanes plus golangci-lint.

```release-note
Fix Tree.Style and Tree.RemoveStyle ranges starting after a merge-moved child styling only the writer's replica.
```

**Additional documentation**:

```docs
- [Design]: docs/design/concurrent-merge-split.md §9.4 (Fix 23)
```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved styling and style removal when concurrent paragraph merges move or collapse range anchors.
  - Preserved formatting on affected inserted content without incorrectly expanding ranges.
  - Ensured style removals are recorded consistently, including when attributes are missing.
- **Tests**
  - Added regression coverage for concurrent merge scenarios, range recovery, and synchronized style removal.
- **Documentation**
  - Documented the fix, recovery behavior, testing lessons, and known limitations.

### DIFF
--- a/docs/design/concurrent-merge-split.md
+++ b/docs/design/concurrent-merge-split.md
@@ -489,12 +489,52 @@ carries the same provenance under either merge-then-split or
 split-then-merge application order and is judged identically by the
 filter.
 
+**From-side recovery (Fix 23).** A range *start* anchored after a moved
+child is the mirror image: the merge moves the anchor child behind the
+interlopers, so on the applying replica the resolved range collapses
+(start past end) and the traversal misses nodes the styling client
+covered — the writer keeps a one-sided attribute entry. A skip filter
+cannot compensate (the traversal covers nothing), so
+`reversedFromAnchorRecovery` re-anchors the traversal start just after
+the last live sibling before the merge-source tombstone, but only when
+the same §9.4 shape holds for the *from* position **and** the resolved
+range actually collapsed; an ordered range that moved with the merge is
+left untouched. The recovered traversal styles only nodes the interloper
+predicate positively identifies (stamp-free children between the
+tombstone and the moved children); version-vector checks keep inserts
+unknown to the styler unstyled. For `RemoveStyle` this materializes the
+removal tombstone on both replicas — it must be recorded even for a
+missing key, because it arbitrates a concurrent `SetAttr` with an
+earlier ticket.
+
 **Known limitations** (tracked as follow-ups):
 
-- A range *start* anchored after a moved child shrinks the traversal
-  on the applying replica instead of growing it (the interlopers sit
-  before the start), which a skip filter cannot compensate; likewise
-  an edit-only divergence independent of styling.
+- An edit-only divergence independent of styling (concurrent unwrap
+  versus merge-delete of the same paragraph) remains open.
+- Within a collapsed from-side range, any *stamped* node fails open in
+  the other direction: a merge-moved element child at or after the
+  original from anchor (probe shape `<p>c<b/></p>`), an insert declared
+  inside the merged-away parent, or a child moved into the target by an
+  *earlier* merge the styler knew (and styled inside its range) all
+  carry a merge stamp, so the recovery cannot tell them from moved
+  content before the anchor and skips them — the writer styles them,
+  the applying replica does not. The pre-fix behavior styled nothing at
+  all there, so the recovery strictly narrows the divergence but does
+  not close this part.
+- A merge into a *sibling* (not the common parent) pulls the from
+  anchor inside that sibling, whose End token newly enters the range
+  and is styled on the merged replica only; the guard's direct-child
+  shape restriction deliberately excludes this shape (reproduces at
+  base, before Fix 23).
+- A range *end* whose left sibling is the merge-source tombstone under
+  a surviving parent resolves past the merge target's End token on the
+  applying replica and styles the target itself one-sidedly
+  (PBT counterexample: `RemoveStyle(6,8)` against `Edit(1,5)`).
+- The recovery inherits the guard's direct-child shape restriction, so
+  a from-anchor collapsed by a *chained* merge (declared parent's
+  tombstone left under an intermediate source) gets no recovery and
+  keeps the pre-fix behavior of styling nothing on the applying
+  replica.
 - A node moved into the target by an *earlier, different* merge that
   the styler knew (and therefore saw outside its range) also fails
   open and is styled on the applying replica — the pre-§9.4 behavior;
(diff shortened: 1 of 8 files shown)


hackerwins:
Maintainer review pass — addressed the outstanding CodeRabbit items in
b630da22.

- The `require.Len` item was already fixed in aa5e6bbf.
- "Compare the removal tombstones directly" was still open and was a
  real gap: the loop asserted the shape *per replica*, which passes even
  when the two entries carry different identities or removal tickets —
  the divergence class #1942 is about. Each replica's attribute entries
  are now rendered as sorted descriptors (key, value, `updatedAt`,
  removal flag) and the two lists are compared to each other. The
  per-replica assertions stay, since "identical on both replicas" alone
  would also be satisfied by both being empty.
- Same commit turns the remaining `assert` calls guarding a dereference
  into `require`, for the reason the length check already did.

Verified locally: `TestTreeConcurrencyStyleFromSideMovedAnchor`,
`TestTreeConcurrencyRemoveStyleFromSideMovedAnchor` and
`TestTreeConcurrencyStyleFromSideOrderedRange` pass, the full
`-tags complex -race -run TestTree` lane is green, and the RemoveStyle
test still fails cleanly on `d2: removal entry count` when
`pkg/document/crdt/tree.go` is reverted to `main` — so it remains a
valid regression oracle.

Also checked the Go/JS implementation parity across this PR and
yorkie-team/yorkie-js-sdk#1329 (guard extension, collapse check, anchor
selection over tombstone-inclusive children, predicate ordering): they
agree. The one shape difference — Go reassigning `fromParent`/`fromLeft`
in place where the JS uses separate `traverseFrom*` locals — is
behaviorally identical, since neither variable is read after the
traversal.

hackerwins (approved):
Reviewed the CRDT change and its Go/JS parity against yorkie-team/yorkie-js-sdk#1329; addressed the outstanding review items in b630da22. complex-test, build, load-test and doc-links are green on that commit.

Issue #1382, opened 2025-07-14
Introduce Operation Reorder Engine for non-commutative Operations


hackerwins:
**What would you like to be added**:

### Background

In Yorkie's CRDT system, most operations can leverage commutativity for convergence without requiring explicit reordering. This includes: `Object.Set`, `Object.Remove`, `Array.Insert`, `Array.Remove`, Text.Edit, Tree.Edit

However, certain position-sensitive operations are inherently non-commutative, and their final state depends on the order of application. These include: `Tree.Split`, `Tree.Merge`, `Array.Move`, `Object.Move`

These operations often require causal reordering or full replay to ensure convergence. Moreover, schema constraints (e.g., "this array can have only two items") can prevent concurrent operations from being applied directly, requiring validation and reconciliation.

To address these challenges, I propose a Reorder/Replay engine that guarantees convergence for all operation types through VersionVector-aware reordering.

### 🎯 Goals

- Introduce a Replay engine that can reorder non-commutative operations based on causal history
- Ensure 100% convergence across all CRDT operation types
- Maintain acceptable performance and memory overhead
- Integrate with existing PushPull sync model

### 🧩 Operation History Structure

We propose maintaining three components for operation replay and compaction:

- Start Snapshot: The document state at the minVersionVector received from the server.
- Linear Operation Array: A causally ordered array of operations(sorted by VersionVector). This forms the basis for replay.
- Latest Snapshot (Optional): Used for client-side caching and rendering, updated periodically but not required for correctness.

### Benefits

- Simplifies replay logic with a linear log
- Easy integration with PushPull API
- Enables GC via minVersionVector
- Avoids the complexity of DAGs or tree-based histories
- Supports schema validation and rollback during replay

### 📐 Design & Implementation Plan

A. Algorithm Design & Validation
- [ ] Define operation classification (commutative vs non-commutative)
- [ ] Design replay strategy based on VersionVector ordering
- [ ] Explore hybrid approaches (commutativity + replay fallback)
- [ ] Define schema validation & conflict resolution policy
- [ ] Build prototype and benchmark replay performance

B. Engine Implementation
- [ ] Implement VersionVector-based OperationHistory
- [ ] Integrate with PushPull response (start snapshot, minVersionVector, operations)
- [ ] Build snapshot-based replay logic
- [ ] Optimize in-memory operation storage
- [ ] Handle schema validation during replay

C. Optimization & Integration

- [ ] Analyze time/space complexity and optimize
[...]

PR #1958, opened 2026-08-30
fix(crdt): anchor Array.Add on position identity to fix snapshot divergence

### LINKED ISSUE #1948: Array concurrent Move does not converge under random concurrency (property-test divergence)
**What happened**:

While adding a property-based convergence test for `Json.Array` (#1946), concurrent Array `Move` operations (`MoveBefore` / `MoveAfterByIndex`) were found to leave replicas **permanently divergent** under randomized concurrency, so they were excluded from the randomized generator. This issue records a **minimal, deterministic reproduction** so the divergence is tracked as a real convergence gap rather than only a code comment.

The hand-written scenarios in `test/integration/array_test.go` (`TestConcurrentArrayOperations`, `TestComplicatedArrayConcurrency`) still converge — the specific cases fixed by #987 / #1423 / #1416 hold. The gap appears only under broader concurrency: **≥ 3 replicas**, an initially empty array built up by concurrent `Add`s, and a client that issues **two `Move`s followed by trailing `Add`s** concurrently with another client's inserts. Both `MoveAfterByIndex` and `MoveBefore` reproduce it.

After every replica exchanges all changes, all replicas hold the same element order. Every operation in the reproduction is individually valid, the operation set is identical on all replicas, and the element multiset is identical after merge — only the order differs, so convergence is the correct expectation.

Deterministic 3-replica integration test (drop into `test/integration/`, run with mongo up):

```go
//go:build integration

package integration

import (
	"context"
	"testing"

	"github.com/stretchr/testify/require"

	"github.com/yorkie-team/yorkie/pkg/document/presence"
)

func TestArrayConcurrentMoveConvergence(t *testing.T) {
	clients := activeClients(t, 3)
	ctx := context.Background()

	// Attach three replicas to an empty shared array.
	var pairs []clientAndDocPair
	for _, c := range clients {
		doc := document.New(helper.TestKey(t))
		require.NoError(t, c.Attach(ctx, doc))
		pairs = append(pairs, clientAndDocPair{c, doc})
	}
	require.NoError(t, pairs[0].doc.Update(func(root *json.Object, p *presence.Presence) error {
		root.SetNewArray("a")
		return nil
	}))
	for _, pr := range pairs {
		require.NoError(t, pr.cli.Sync(ctx))
	}
	require.NoError(t, pairs[0].cli.Sync(ctx))
	require.NoError(t, pairs[1].cli.Sync(ctx))

	// Each mutation is its own Update (its own change), applied concurrently.

	// Client 0: build [14,15], move index 0 after 1, then move it back,
	//           then append 26 and 66. Net local order: [14,15,26,66].
	c0 := pairs[0].doc
	require.NoError(t, c0.Update(func(r *json.Object, p *presence.Presence) error { r.GetArray("a").AddInteger(14); return nil }))
	require.NoError(t, c0.Update(func(r *json.Object, p *presence.Presence) error { r.GetArray("a").InsertIntegerAfter(0, 15); return nil }))
	require.NoError(t, c0.Update(func(r *json.Object, p *presence.Presence) error { r.GetArray("a").MoveAfterByIndex(1, 0); return nil }))
	require.NoError(t, c0.Update(func(r *json.Object, p *presence.Presence) error { r.GetArray("a").MoveAfterByIndex(1, 0); return nil }))
	require.NoError(t, c0.Update(func(r *json.Object, p *presence.Presence) error { r.GetArray("a").AddInteger(26); return nil }))
	require.NoError(t, c0.Update(func(r *json.Object, p *presence.Presence) error { r.GetArray("a").AddInteger(66); return nil }))

	// Client 1 (concurrent): build [18,89,56,94] with two Adds and two Inserts.
	c1 := pairs[1].doc
	require.NoError(t, c1.Update(func(r *json.Object, p *presence.Presence) error { r.GetArray("a").AddInteger(18); return nil }))
	require.NoError(t, c1.Update(func(r *json.Object, p *presence.Presence) error { r.GetArray("a").InsertIntegerAfter(0, 56); return nil }))
	require.NoError(t, c1.Update(func(r *json.Object, p *presence.Presence) error { r.GetArray("a").AddInteger(94); return nil }))
	require.NoError(t, c1.Update(func(r *json.Object, p *presence.Presence) error { r.GetArray("a").InsertIntegerAfter(0, 89); return nil }))

	// Client 2: no local edits (passive replica).

	// Exchange all changes (extra rounds guarantee full propagation).
	for round := 0; round < 3; round++ {
		for _, pr := range pairs {
			require.NoError(t, pr.cli.Sync(ctx))
		}
	}

	got0 := pairs[0].doc.Root().GetArray("a").Marshal()
	got1 := pairs[1].doc.Root().GetArray("a").Marshal()
	got2 := pairs[2].doc.Root().GetArray("a").Marshal()
	t.Logf("d0 = %s", got0)
	t.Logf("d1 = %s", got1)
	t.Logf("d2 = %s", got2)
	assert.Equal(t, got0, got1)
	assert.Equal(t, got0, got2) // FAILS: replica 2 diverges
}
```

Run:

```
go test -tags integration -race -run TestArrayConcurrentMoveConvergence ./test/integration/ -count=1 -v
```

**Exact divergent output** (stable — re-syncing 8 additional rounds does not change it, so this is a permanent divergence, not under-propagation):

```
d0 = [18,89,56,94,14,15,26,66]
d1 = [18,89,56,94,14,15,26,66]
d2 = [18,89,56,94,66,26,14,15]   <-- diverges
```

Replicas 0 and 1 agree; the passive replica 2 reorders client 0's entire moved-and-appended block from `14,15,26,66` to `66,26,14,15`. The structure (length, element multiset) is preserved — only the order diverges, matching the failure mode described in #987.

**`MoveBefore` is affected too.** The same randomized 3-client generator, restricted to `MoveBefore`, first diverges at seed 6:

```
c0 moves: MoveBefore(next=0,target=1), MoveBefore(next=0,target=2)
d0 = [79,14,65,62,48,70,23,69,42,16]
d1 = [79,14,65,62,70,23,48,69,42,16]   <-- diverges
d2 = [79,14,65,62,48,70,23,69,42,16]
```

The divergence is sensitive: reducing to a single `Move`, dropping a trailing `Add`, shrinking the concurrent client's sequence, or using only 2 replicas all converge. It requires enough concurrent structure to expose the order-dependent (non-commutative) merge — which is expected for this class of bug.

**Classification.** This is an instance of the **general non-commutative `Array.Move` limitation tracked by #1382** (Operation Reorder/Replay Engine), not a regression of the specific cases already fixed:

- #676 / #1416 (reference element moved → structure kept, values diverge) and #1423 (`remove`/`set` vs `move` via the `removedAt` condition) were the concrete cases addressed by #987 / #1395. Their hand-written scenarios still converge here, so those fixes hold.
- #987 ("Implement CRDT Move Convergence Mechanism", closed COMPLETED) implemented `movedAt` last-writer-wins per moved element. That resolves same-element concurrent moves, but the merged order still depends on the order in which concurrent remote moves are replayed against each replica's live list, so it does not guarantee convergence in the general case.
- #1382 (open) explicitly lists `Array.Move` among the inherently non-commutative operations that "do not yet converge in the general case" and proposes reorder/replay as the fix. The reproduction above is a concrete data point for it.

This issue is filed to mirror the #1730 precedent of documenting property-test divergences with a minimal reproducer instead of leaving them as a code comment. Concurrent `Move` was scoped out of the randomized `Json.Array` generator in #1946 for this reason; the exclusion is justified until #1382 lands.

**Environment**:
- Operating system: macOS (darwin)
- Yorkie version (use `yorkie version`): v0.7.17 (commit `2a4eeff6`, `v0.7.17-2-g2a4eeff6`), MongoDB backend
- Reproduced with the Go SDK integration harness (`test/integration`)

lemon0333:

Concurrent Array `Move` mixed with `Add`/`Insert` left replicas permanently divergent under random concurrency: a passive replica reordered another client's moved-and-appended block (e.g. `[...,14,15,26,66]` vs `[...,66,26,14,15]`).

**Root cause — a reconstruction bug, not a CRDT-ordering bug.**

`RGATreeList.Add` anchored the appended element on the last node's *element* `createdAt` (`a.last.CreatedAt()`) instead of its *position-node* `createdAt`. For an element that was moved into its slot, the two differ, and the element `createdAt` resolves via `nodeMapByCreatedAt` to the element's now-dead *original* position node (dead position nodes are intentionally kept in that map as stable anchors). So the append anchored at that stale dead slot early in the list, and the forward-skip in `insertAfter` then placed each successive `Add` *before* the previous one — reversing the appended run.

The live insert path was never affected: the `json` layer already converts to position identity via `LastCreatedAt()` before issuing operations. The bug surfaced only when the list is rebuilt through `Add`, i.e. **snapshot restore** (`fromJSONArray`) and **`Array.DeepCopy`**. That is exactly why the passive replica in the repro diverged only *after* the server took a snapshot, while every pure operation-replay ordering converges (verified by exhaustively enumerating all interleavings).

**Fix:** anchor `Add` on `LastCreatedAt()` (the last node's position identity). For never-moved elements position-createdAt == element-createdAt, so behavior is unchanged; it only differs for a moved last element.

Scope note: this is a scoped serialization/reconstruction fix. It is **not** the fundamental non-commutative-move redesign discussed in #1382 — the RGA LWW position register (#1762) is otherwise sound here.

Fixes #1948

Verification:
- New minimal, no-server regression `TestArrayMoveThenAddSnapshotStable` (pure CRDT `DeepCopy`) — fails without the fix, passes with it.
- Re-enabled concurrent `Move` (`MoveBefore` + `MoveAfterByIndex`) in the randomized property test `TestArrayPropertyConvergence`, which had excluded it. Confirmed deterministically green, including an out-of-band stress run at 200 seeds × 12 ops (x2) before restoring CI-friendly counts.
- Existing suites pass: `TestArray`, `TestArrayConcurrency` (49-case table), `TestComplicatedArrayConcurrency`, `TestArraySetByIndex`, plus `go build ./...`, `go vet ./pkg/document/...`, and the `crdt`/`document`/`converter`/`operations` unit tests.
- Additional confidence: a 100-seed randomized 3-client Move fuzz on a real server fails without the fix and passes with it.

Yes — concurrent Array `Move` now converges across replicas when reconstructed from a snapshot.

```release-note
Fix Array convergence divergence where a moved-then-appended element block was reversed after snapshot restore / DeepCopy (concurrent Move).
```

**Checklist**:

- [x] Didn't break anything

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where adding items after moving the last array element could reverse the order of newly appended items.
  * Ensured snapshots and copied arrays preserve the correct item order.

* **Tests**
  * Expanded array testing to cover move operations and verify consistent results across reconstruction and concurrent scenarios.

### DIFF
--- a/pkg/document/crdt/array_move_snapshot_test.go
+++ b/pkg/document/crdt/array_move_snapshot_test.go
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2026 The Yorkie Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package crdt_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/yorkie-team/yorkie/pkg/document/crdt"
+	"github.com/yorkie-team/yorkie/test/helper"
+)
+
+// TestArrayMoveThenAddSnapshotStable is the minimal deterministic regression
+// for yorkie#1948: when the last element of an array was moved into its slot,
+// appending more elements and then rebuilding the array through Add (which is
+// how DeepCopy and snapshot restore reconstruct the list) must preserve order.
+//
+// Before the fix, RGATreeList.Add anchored on the last node's ELEMENT createdAt
+// instead of its POSITION createdAt. For a moved last element those differ, and
+// the element createdAt resolves to the element's now-dead original position
+// node, so each Add landed before the previous one — reversing the appended
+// run. This surfaced as a passive replica diverging after a snapshot restore.
+func TestArrayMoveThenAddSnapshotStable(t *testing.T) {
+	root := helper.TestRoot()
+	ctx := helper.TextChangeContext(root)
+
+	list := crdt.NewRGATreeList()
+
+	add := func(v int) {
+		p, err := crdt.NewPrimitive(v, ctx.IssueTimeTicket())
+		require.NoError(t, err)
+		require.NoError(t, list.Add(p))
+	}
+
+	// Build [14,15].
+	p14, err := crdt.NewPrimitive(14, ctx.IssueTimeTicket())
+	require.NoError(t, err)
+	require.NoError(t, list.Add(p14))
+	p15, err := crdt.NewPrimitive(15, ctx.IssueTimeTicket())
+	require.NoError(t, err)
+	pos14, err := list.PosCreatedAt(p14.CreatedAt())
+	require.NoError(t, err)
+	require.NoError(t, list.InsertAfter(pos14, p15, ctx.IssueTimeTicket()))
+	assert.Equal(t, "[14,15]", list.Marshal())
+
+	// Move 14 after 15, then move 15 after 14 (net order [14,15], two moves that
+	// leave two dead position nodes and a moved last element).
+	pos15, err := list.PosCreatedAt(p15.CreatedAt())
+	require.NoError(t, err)
+	_, err = list.MoveAfter(pos15, p14.CreatedAt(), ctx.IssueTimeTicket())
+	require.NoError(t, err)
+	pos14b, err := list.PosCreatedAt(p14.CreatedAt())
+	require.NoError(t, err)
+	_, err = list.MoveAfter(pos14b, p15.CreatedAt(), ctx.IssueTimeTicket())
+	require.NoError(t, err)
+	assert.Equal(t, "[14,15]", list.Marshal())
+
+	// Append 26 and 66 after the moved last element.
+	add(26)
+	add(66)
+	arr := crdt.NewArray(list, ctx.IssueTimeTicket())
+	assert.Equal(t, "[14,15,26,66]", arr.Marshal())
+
+	// DeepCopy rebuilds the list through Add — it must not reorder.
+	cp, err := arr.DeepCopy()
+	require.NoError(t, err)
+	assert.Equal(t, "[14,15,26,66]", cp.Marshal(),
+		"DeepCopy of a moved-then-appended array must preserve order (yorkie#1948)")
+}
--- a/pkg/document/crdt/rga_tree_list.go
+++ b/pkg/document/crdt/rga_tree_list.go
@@ -241,8 +241,17 @@ func (a *RGATreeList) Marshal() string {
 }
 
 // Add adds the given element at the last.
+//
+// The anchor must be the last node's POSITION identity (LastCreatedAt), not its
+// element identity (last.CreatedAt). When the last element was moved here, its
+// element createdAt maps in nodeMapByCreatedAt to the element's now-dead
+// original position node, so anchoring on element identity would insert after
+// that stale dead slot instead of the tail — and the forward-skip in
+// insertAfter would then place each successive Add before the previous one,
+// reversing the appended run (yorkie#1948, seen only via snapshot restore and
+// DeepCopy, which rebuild the list through Add).
 func (a *RGATreeList) Add(elem Element) error {
-	return a.InsertAfter(a.last.CreatedAt(), elem, nil)
+	return a.InsertAfter(a.LastCreatedAt(), elem, nil)
 }
 
 // AddDeadPosition appends a dead position node during snapshot restoration.
--- a/test/integration/array_property_test.go
+++ b/test/integration/array_property_test.go
@@ -52,22 +52,19 @@ const (
 // read the replica's local length so the randomly picked indices always stay in
 // bounds.
 //
-// NOTE(convergence scope): The generated ops are limited to Add, Insert, Delete
-// and Set, which are the Array operations that converge under arbitrary
-// concurrency. Concurrent Move (MoveBefore/MoveAfterByIndex) is a non-commutative
-// operation that does not yet converge in the general case — a known limitation
-// tracked by #1382 (Operation Reorder/Replay Engine for non-commutative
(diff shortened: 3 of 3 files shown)


hackerwins:
### Code review

Found 1 issue:

1. The identical defect is still live in yorkie-js-sdk, so this Go-only fix converts a uniform cross-SDK bug into a permanent server-vs-client divergence. `RGATreeList.insert()` in the JS SDK still anchors on element identity (`this.last.getCreatedAt()`) even though `getLastCreatedAt()` — the position-identity accessor — already exists, and `insert()` is the snapshot-restore path via `fromArray`. After this PR a Go server rebuilds a moved-then-appended array as `[14,15,26,66]` while a JS client decoding the same snapshot still yields `[66,26,14,15]`. Note JS `CRDTArray.deepcopy` already uses `getLastCreatedAt()`, so the surviving JS defect is precisely the snapshot path #1948 reported. This is the case ruled out on #1932: "a one-sided Go fix would turn a uniform cross-SDK bug into a permanent server-vs-client divergence, since Go is also the server" (https://github.com/yorkie-team/yorkie/pull/1932#discussion_r3791889503). Suggest fixing both SDKs together, or filing the JS counterpart with file:line citations before merging.

https://github.com/yorkie-team/yorkie/blob/0a153121c112bc63987c1694d55c0ed07dd062be/pkg/document/crdt/rga_tree_list.go#L253-L255

JS counterpart still on element identity:

https://github.com/yorkie-team/yorkie-js-sdk/blob/dce5ef728c5a0fb969eb97a1e222bf00fef57f38/packages/sdk/src/document/crdt/rga_tree_list.ts#L485-L489

Called from the snapshot-restore path:

https://github.com/yorkie-team/yorkie-js-sdk/blob/dce5ef728c5a0fb969eb97a1e222bf00fef57f38/packages/sdk/src/api/converter.ts#L1706-L1711

The one-line Go fix itself is correct and well-diagnosed. It restores the intent of #1762, which moved `LastCreatedAt()` to `PositionCreatedAt()` "to avoid element/position identity confusion" but left `Add` as a missed call site. Verified red-to-green on the new unit test, an `ObjectToBytes`/`BytesToObject` round-trip, and the re-enabled property test.

hackerwins:
The JS counterpart is up: yorkie-team/yorkie-js-sdk#1332. It anchors `RGATreeList.insert` on `getLastCreatedAt()` and adds a snapshot round-trip regression test, so the cross-SDK divergence raised above is resolved once both land. Merging this first, then the SDK side.

PR #1990, opened 2026-09-15
Book an already-tombstoned element into gc when registering it

hackerwins:
Mirrors yorkie-team/yorkie-js-sdk#1350. `DocSize` gates `MaxSizeLimit` in both
SDKs and the server reports it, so the two implementations have to agree. Two
commits.

## 1. `Charge docSize.GC by element, not by createdAt`

`sizeInGC` records what GC is holding for an element. It was keyed by that
element's `createdAt`, with the element carried alongside so every read could
check it was the right one — but a `createdAt` names one element for most of a
document's life, not all of it. Undo restores a `DeepCopy` under the original's
identity while the original is still a tombstone, and undoing an *array* removal
reissues a ticket for the restored container alone, so its members come back
aliasing the tombstoned ones outright.

One slot cannot hold both. Whichever is charged last displaces the other, and
deregistering the displaced one then finds no charge of its own and falls
through to subtracting its size from Live, which was never holding it:

```
k = [{a:"1", b:"2"}] → delete k[0].b → delete k[0] → Undo() → delete k[0] → GC

before   Live {Data:-4, Meta:-48}   GC {Data:4, Meta:120}   `{"k":[]}`
after    Live {Data: 0, Meta: 72}   GC {Data:0, Meta:  0}   `{"k":[]}`
```

Keying the map by the element fixes it. Every read already wanted the charge for
one specific element, so the identity checks that guarded the key lookups are
what the map now does on its own — `gcCharge` and all three checks go away.

## 2. `Book an already-tombstoned element into gc when registering it`

`RegisterElement` charges `docSize.Live` and leaves the move to GC to a later
removal, and `RegisterRemovedElementPair` refunds one `TimeTicket` on that move
because Live was charged before the `removedAt` ticket existed. An element can
also be registered *after* it is already removed, and then Live did hold the
ticket and the refund over-credits it. Three routes reach that state:

1. **An undo re-sets the `DeepCopy` its reverse captured**, which keeps the
   members tombstoned before the container was. Nothing books those copies into
   GC, so they sit in Live and no collection entry names them.
2. **`ElementRHT` marks the losing side of a concurrent `Set` removed** before
   `Set.Execute` registers it, so the replica that applies the loser reports one
   ticket more than the one that does not, for identical content.
3. **`NewRoot` reads tombstones out of a snapshot** the same way — the drift the
   NOTE in `RegisterRemovedElementPair` already admitted and
   `docs/tasks/active/20260817-docsize-snapshot-rebuild-drift-todo.md` tracks.

Measured, byte-identical to what the JS SDK reported:

| | before | after |
|---|---|---|
| undo of a container holding a tombstone | collects 0, `Live {4,192}` | collects 1, `Live {2,120}` |
| replicas converged on `{"k":"2"}` | `Live.Meta` 72 vs **96** | 72 vs 72 |
| doc with one tombstone, `DeepCopy` | `Live.Meta` 120 vs **144** | 120 vs 120 |

Decide the side once, at the one place every route passes through: an element
that already carries a `removedAt` when it is registered is moved to GC and
entered for collection, and the refund is left to the elements that were
registered live and removed afterwards. `RegisterElement` takes the parent it
needs for that entry; every caller already had it in hand, and
`RegisterRemovedElementPair` next to it already took one.

## Notes for review

- `RegisterElement` now has a side effect on `docSize.GC` and
  `gcElementPairMap`. Callers are `NewRoot`, `Set`, `Add`, `ArraySet` and
  `change.Context`.
- `NewRoot`'s `if elem.RemovedAt() != nil { RegisterRemovedElementPair(...) }`
  scan is removed: `RegisterElement(root, nil)` above it already walks the same
  descendants and books every tombstone.
- Adopting runs as a second pass over the subtree rather than during the walk
  that charges Live. Doing it inline would move descendants Live has not been
  charged for yet, driving Live negative — and collecting them into a slice
  instead captures it in the `Descendants` closure, forcing it onto the heap for
  one extra allocation per registration.
- The array id aliasing itself is untouched. Commit 1 makes the accounting
  survive it; removing it means reissuing ids for the whole restored subtree in
  the `Add` undo branch, which is a separate change (yorkie-js-sdk#1351).
- **Behaviour change worth flagging.** The size limit is enforced against the
  clone's total, and the clone is built the way commit 2 fixes, so a document
  holding uncollected tombstones is now enforced one ticket (24 bytes) per
  tombstone smaller than before. The figure `DocSize()` reports and the one the
  limit is enforced on now agree; they differed by exactly that amount.
- Both SDKs carry a note at the identity check in `deregisterElement` saying
  that one reads identity from its own gc entry and the other from the pair map,
  because of what each element map holds -- same rule, one resolver each. Giving
  the JS worklist an entry of its own is yorkie-js-sdk#1351.

## Verification

- `go build ./...`, `go vet ./...` clean; `golangci-lint run ./pkg/document/...`
  reports 0 issues
- `go test -count=1 ./pkg/... ./api/... ./client/...` all pass; each commit
  builds and passes on its own
- Every added test fails on `main` and passes here (all five)
- 500 random undo / remove / restore / collect sequences: `docSize` went
  negative on 21 of 500 seeds on `main` and on 0 of 500 here, with no stale gc
  worklist entry on either. Measured in the JS port; the two implementations run
  the same accounting
- A 7-scenario parity probe (object restore, array restore, array cycle,
  snapshot rebuild, repeated restore/remove, 3-level nesting) reports
  byte-identical `DocSize`, garbage length, element count and JSON in both SDKs
- Nested tombstones cost O(depth^2) node visits during registration on `main`
  too (depth 32: 13,750 visits vs 14,905 here), so this adds a constant factor
  rather than the complexity. Folding the two passes into one is yorkie-js-sdk#1351
- `BenchmarkArray` allocations match `main` (`random_set_100000` 1,102,606;
  `mixed_workload_100000` 752,118) and bytes are lower, since the charge map no
  longer allocates a wrapper struct or a string key
- Server tests needing a running MongoDB were not run locally

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved document size tracking when undoing removals, restoring content, replacing values, and handling nested arrays or objects.
  - Improved garbage collection so removed content is reclaimed correctly without affecting live document size.
  - Improved consistency of document sizes across replicas after concurrent updates.
  - Ensured copied documents preserve accurate size information and remain fully editable after restoration.


hackerwins:
Thanks — same finding was raised on the JS side (yorkie-js-sdk#1350), and I measured it there before changing anything. It is **pre-existing on `main`, byte-identical**, so this PR neither introduces nor worsens it.

`k = [{a:'1', b:'2'}]` → `delete k[0].b` → `delete k[0]` → `Undo()` → `delete k[0]` → `GarbageCollect()` reaches the same state with and without this change. On an object container the same sequence is clean on both.

On the suggested fix — recording the copy in `gcElementPairMap` by element identity while keeping the original's charge — the obstacle is that `gcElementPairMap` and `sizeInGC` are both keyed by `createdAt`, one slot each. Keeping both the tombstone and the copy needs the maps to hold two entries for one key, which is a change to the identity model rather than to `adoptSize`. The real fix is upstream of the accounting: reissue ids for the whole restored subtree in the `Add` branch of undo, so two elements never answer to one `createdAt`. Then `adoptSize`'s guard becomes unreachable and can go.

What this PR does is refuse to make it worse: `adoptSize` stands aside exactly when a different element still holds a non-zero charge, and `restoring an array container holding an earlier tombstone test` pins the array path's current figures — that test passes on `main` too, so it records existing behaviour rather than claiming it correct.

I'll file the aliasing bug separately; bundling it here would mix a size-accounting fix with an identity-reissue change.

Also addressed the benchmark signal in the pushed update. The first version collected tombstones into a slice, and capturing it forced the `Descendants` closure onto the heap — one extra allocation per registration. `RegisterElement` now does the live walk and the tombstone walk as two passes with no captured slice:

```
BenchmarkArray/random_set_100000       main 1,102,606 allocs   before 1,202,606   now 1,102,606
BenchmarkArray/mixed_workload_100000   main   752,118 allocs   before   807,170   now   752,117
```

Times are back within noise of `main` as well. The CI benchmark comment also flagged `BenchmarkChannel*` moving 20-44%, which this PR does not touch — that part looks like runner noise.

hackerwins:
Followed up on this rather than deferring it — the negative `docSize` is now fixed in this PR, as a separate first commit.

The root cause turned out to be the charge map itself. Every read of `sizeInGC` already wanted "the charge for *this* element" and expressed it as a `createdAt` lookup plus an identity check, so the aliasing case had two elements competing for one slot. Keying the map by the element instead gives each a slot of its own, and the identity checks become what the map does on its own.

```
k = [{a,b}] -> delete b -> delete k[0] -> undo -> delete k[0] -> gc
before  live {-4, -48}  gc {4, 120}
after   live { 0,  72}  gc {0,   0}   == a freshly built {"k":[]}
```

That also made the `adoptSize` guard you were looking at unnecessary, so it is gone rather than hardened — there is no slot to lose any more. The array restore path now settles at the size it was built at, which resolves a JS/Go divergence I had noted as out of scope, and the test that pinned the old figures asserts the correct ones.

Net effect on the accounting code is smaller than before: `gcCharge`, the three identity checks and the guard all go away.

hackerwins:
Mirrored the two review points from yorkie-js-sdk#1350 here: the regression test's comment was still describing the model commit 1 replaced, and both SDKs now carry a note at the identity check explaining that the JS worklist holds keys while this one holds the pair — same rule, different resolver.

The traversal point raised there applies equally here. Measured: the quadratic growth in nested tombstones is on `main` too (the constructor already walks each tombstone's subtree once per tombstone), this PR adds a constant factor, and `BenchmarkArray` allocations are identical to `main`. Tracked in yorkie-js-sdk#1351, to be changed on both SDKs together.

hackerwins:
Mirrored a review point from yorkie-js-sdk#1350: the rebuild test now asserts the reconstructed root reports the tombstone as garbage and collects it, not just that the sizes match.

That is the assertion this PR most needed there: commit 2 removes `NewRoot`'s own `RegisterRemovedElementPair` scan and leaves entering the tombstone for collection to `RegisterElement`, so a rebuilt root that loaded the size correctly but never entered it would strand the charge with nothing reporting it as garbage — and the old test could not tell the two apart.

